PR #17, opened 2026-05-17
[codex] Add timeline generation hooks

michaeltrefry:
﻿## Summary
- Add timeline item version metadata and a frame extraction job endpoint for selected timeline clips.
- Wire editor actions for extracting frames, sending selected assets to studios, extending clips, generating bridge clips, and nondestructive replacement.
- Teach the worker to extract timeline frames and render extend/bridge jobs from source clip frames, with lineage metadata preserved.

## Validation
- `python -m pytest`
- `npm test` in `apps/web`
- Browser smoke pass on `http://127.0.0.1:5173` for timeline creation, adding/selecting an item, and the new generation hook controls.

### DIFF
--- a/apps/api/sceneworks_api/assets.py
+++ b/apps/api/sceneworks_api/assets.py
@@ -89,6 +89,7 @@ def list_assets(
 ) -> list[dict[str, Any]]:
     project_path = find_project_path(request.app.state.settings, project_id)
     assets = []
+    seen_asset_ids = set()
     ensure_project_db_ready(project_path)
     with sqlite3.connect(project_path / "project.db") as connection:
         total = connection.execute("select count(*) from assets").fetchone()[0]
@@ -119,6 +120,9 @@ def list_assets(
                 asset = normalize_asset(project_id, project_path, sidecar_path)
             except (OSError, json.JSONDecodeError):
                 continue
+            if asset.get("id") in seen_asset_ids:
+                break
+            seen_asset_ids.add(asset.get("id"))
             assets.append(asset)
             break
 
--- a/apps/api/sceneworks_api/jobs.py
+++ b/apps/api/sceneworks_api/jobs.py
@@ -20,6 +20,7 @@ JobType = Literal[
     "video_generate",
     "video_extend",
     "video_bridge",
+    "frame_extract",
     "timeline_export",
     "model_download",
     "lora_import",
--- a/apps/api/sceneworks_api/timelines.py
+++ b/apps/api/sceneworks_api/timelines.py
@@ -65,7 +65,7 @@ class TimelineItem(BaseModel):
     transitionOut: Transition | None = None
 
     @model_validator(mode="after")
-    def validate_ranges(self) -> "TimelineItem":
+    def validate_ranges_and_populate_version_metadata(self) -> "TimelineItem":
         if self.sourceOut <= self.sourceIn:
             raise ValueError("sourceOut must be greater than sourceIn.")
         if self.timelineEnd <= self.timelineStart:
--- a/apps/api/sceneworks_api/timelines.py
+++ b/apps/api/sceneworks_api/timelines.py
@@ -37,6 +37,14 @@ class Transition(BaseModel):
     duration: float = Field(default=0, ge=0, le=10)
 
 
+class TimelineItemVersion(BaseModel):
+    assetId: str = Field(min_length=1)
+    createdAt: str | None = None
+    source: Literal["original", "replacement", "extension", "bridge", "restore", "manual"] = "manual"
+    jobId: str | None = None
+    note: str | None = None
+
+
 class TimelineItem(BaseModel):
     id: str = Field(default_factory=lambda: f"item_{uuid4().hex}")
     trackId: str
@@ -51,6 +59,8 @@ class TimelineItem(BaseModel):
     fit: Literal["fit", "fill", "stretch"] = "fit"
     volume: float = Field(default=1, ge=0, le=2)
     versionAssetIds: list[str] = Field(default_factory=list)
+    currentVersionAssetId: str | None = None
+    versionHistory: list[TimelineItemVersion] = Field(default_factory=list)
     transitionIn: Transition | None = None
     transitionOut: Transition | None = None
 
@@ -60,6 +70,12 @@ class TimelineItem(BaseModel):
             raise ValueError("sourceOut must be greater than sourceIn.")
         if self.timelineEnd <= self.timelineStart:
             raise ValueError("timelineEnd must be greater than timelineStart.")
+        if not self.currentVersionAssetId:
+            self.currentVersionAssetId = self.assetId
+        if self.assetId not in self.versionAssetIds:
+            self.versionAssetIds.append(self.assetId)
+        if not self.versionHistory:
+            self.versionHistory.append(TimelineItemVersion(assetId=self.assetId, source="original"))
         return self
 
 
@@ -110,6 +126,12 @@ class TimelineExportRequest(BaseModel):
         return self
 
 
+class FrameExtractRequest(BaseModel):
+    playheadSeconds: float = Field(ge=0)
+    intendedUse: Literal["reuse", "first_frame", "last_frame", "video_studio", "image_studio", "bridge", "extension"] = "reuse"
+    requestedGpu: str = "auto"
+
+
 def timeline_file_path(project_path: Path, timeline_id: str, name: str) -> Path:
     return project_path / "timelines" / f"{slugify(name, fallback='timeline', max_length=48)}-{timeline_id[-8:]}.sceneworks.timeline.json"
 
@@ -129,6 +151,19 @@ def default_tracks() -> list[TimelineTrack]:
 def compute_duration(timeline: TimelineDocument) -> float:
     ends = [item.timelineEnd for track in timeline.tracks for item in track.items]
     return max(ends, default=0)
+
+
+def find_timeline_item(timeline: TimelineDocument, item_id: str) -> TimelineItem:
+    for track in timeline.tracks:
+        for item in track.items:
+            if item.id == item_id:
+                return item
+    raise HTTPException(status_code=404, detail="Timeline item not found")
+
+
+def source_timestamp_for_item(item: TimelineItem, playhead_seconds: float) -> float:
+    clamped = min(max(playhead_seconds, item.timelineStart), item.timelineEnd)
+    return item.sourceIn + ((clamped - item.timelineStart) * item.speed)
 
 
 def index_timeline(project_path: Path, timeline: TimelineDocument, rel_path: str) -> None:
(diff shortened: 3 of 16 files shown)


michaeltrefry (commented):
## Code review — PR #17 (Timeline AI Generation Hooks, Epic 1088)

Verified the slice end-to-end against [Phase 11 acceptance criteria](documents/IMPLEMENTATION_PLAN.md) and ran the full suite at PR HEAD (`d14c3c0`).

**Tests:** `pytest tests/` — 34/34 ✅. `npm test` in `apps/web` — 2/2 ✅.

**Acceptance criteria coverage:** all five Phase 11 criteria are satisfied (frame extract → reuse, extend, bridge across gap, bridge lands in the gap, prior versions preserved).

---

### Major

**M-1. Race in `applyCompletedTimelineGeneration` — concurrent generation completions can clobber each other** ([apps/web/src/App.jsx:585-606](apps/web/src/App.jsx#L585))

The handler does GET timeline → mutate → PUT, with no version/etag guard. If a user queues *Extend* and *Bridge* close together and both complete near-simultaneously, both fire `applyCompletedTimelineGeneration` against the same `selectedTimelineId`. Both fetch the pre-update timeline, compute their respective insertions, and the second PUT wins — the first job's new item silently vanishes from the timeline (the asset itself is still in the library, but the timeline never gets the version-history entry).

Two reasonable fixes:
- Serialize the apply step (queue completed jobs through an `async`/`await`-chained promise; only one at a time).
- Or add an `If-Match`-style version field to `TimelineDocument`, bump it on every save, and 409 on mismatch so the client can re-fetch + retry.

**M-2. Pydantic validator with side effects** ([apps/api/sceneworks_api/timelines.py:67-78](apps/api/sceneworks_api/timelines.py#L67))

`validate_ranges` now mutates `currentVersionAssetId`, `versionAssetIds`, and `versionHistory`. It's idempotent and the test confirms behavior, but a validator named "validate_ranges" doing field population is the kind of thing future contributors won't expect when they touch this model. Two options:
- Rename to `populate_version_metadata` (or split into two `@model_validator` methods).
- Move the population into `save_timeline` so the model stays a pure validator.

---

### Medium

**Md-1. `load_ffmpeg_frame` silently swallows ffmpeg stderr** ([apps/worker/scene_worker/video_adapters.py:355-378](apps/worker/scene_worker/video_adapters.py#L355))

`subprocess.run(..., stderr=subprocess.DEVNULL)` then `if result.returncode != 0: return None`. The user-facing failure surfaces only as the upstream "Could not decode a frame from the selected clip." with no clue whether ffmpeg is missing a codec, ffmpeg crashed, or the timestamp was past EOF. Capture the last lines of stderr the same way [F-031's fix did for `run_ffmpeg`](apps/worker/scene_worker/timeline_exporter.py:307-315) and include them in the eventual `RuntimeError`.

**Md-2. ImageStudio auto-forces `edit_image` mode on any image/frame selection** ([apps/web/src/screens/ImageStudio.jsx:34-42](apps/web/src/screens/ImageStudio.jsx#L34))

```js
if ((selectedAsset?.type === "image" || selectedAsset?.type === "frame") && mode !== "edit_image") {
  setMode("edit_image");
}
```
Before this PR, the user could compose a text-to-image prompt with an image asset selected for reference. Now selecting any image-type asset from the Library and visiting the Image Studio bounces them straight into Edit mode. May be the intent (since the new "Send Image" Editor button implies editing the frame), but it's a meaningful UX shift — consider gating this on a `forceMode` prop set only by the new "Send" buttons, not on every selection change.

**Md-3. VideoStudio auto-forces `extend_clip` mode on any video selection** ([apps/web/src/screens/VideoStudio.jsx:42-54](apps/web/src/screens/VideoStudio.jsx#L42))

Same pattern as Md-2 for video.

**Md-4. SSE EventSource now reconnects on every project/timeline switch** ([apps/web/src/App.jsx:215](apps/web/src/App.jsx#L215))

Adding `activeProject?.id` and `selectedTimelineId` to the effect's dep array tears down and reopens the EventSource whenever the user switches projects or timelines. Neither is an auth/identity change. This forces a `/jobs/events/ticket` round-trip + new EventSource per click, increases server load, and causes a brief window where job updates are missed. The handler only reads `activeProject?.id` inside `applyCompletedTimelineGeneration` (via `job.projectId ?? activeProject?.id` fallback), and reads `selectedTimelineId` to decide whether to set `activeTimeline` — neither requires a new SSE subscription. Drop both from the dep array, capture them via refs if you need a fresh value inside handlers.

---

### Low

**Lo-1. Hard-coded extend duration cap** ([EditorScreen.jsx:282](apps/web/src/screens/EditorScreen.jsx#L282))

`Math.min(6, Math.max(2, itemDuration(selectedItem)))` clamps extensions to 2–6 seconds regardless of model. LTX-2.3 supports up to 15s. Expose as a control in the generation panel.

**Lo-2. Duplicate lineage data** ([video_adapters.py:558-575](apps/worker/scene_worker/video_adapters.py#L558))

`recipe.normalizedSettings.timelineContext` and `lineage.timeline` carry the same dict. Pick one home and reference from the other.

**Lo-3. `setActiveView("Queue")` on every generation hook**

Extract / extend / bridge / replace all bounce the user to the Queue view. Disruptive when the user wants to keep editing. Consider a toast/notice instead (the `timelineNotice` mechanism is already in place), or only switch view if no other jobs are running.

**Lo-4. `refreshAssets` dedup with `Map`** ([App.jsx:243-245](apps/web/src/App.jsx#L243))

```js
const fetched = await apiFetch(...);
const items = Array.from(new Map(fetched.map((a) => [a.id, a])).values());
```
Defensive dedup like this usually points at a duplicate source. Worth a quick check of whether the V-4 auto-reindex (`apps/api/sceneworks_api/assets.py:88-94`) is somehow producing duplicate rows after the reindex — or if the new frame-extract write path is racing with reindex.

---

### Info

- **Acceptance:** all five Phase 11 criteria covered.
- **Tests:** `test_timelines.py` provides a solid contract test for the new validator behavior and the frame-extract endpoint's payload shape.
- **Worker capabilities:** `frame_extract` correctly added to `worker_capabilities`; build-in manifests updated to expose `extend_clip`/`video_bridge` on LTX-2.3 and Wan2.2.
- **Schema migration:** `timeline.schema.json` adds `currentVersionAssetId` + `versionHistory` cleanly; existing timelines load via the validator's auto-population path.

---

### Verdict

Ship-ready after addressing **M-1** (race) and **Md-4** (unnecessary EventSource churn) — both are real and easy to fix. The other items are quality-of-life and can land in follow-ups.

michaeltrefry (commented):
## Re-review — commit `756c558` ("Address timeline generation review feedback")

Walked through each item from my prior review. Re-ran the suite at the new HEAD: **34/34 pytest, 2/2 vitest** ✅.

### Summary

| ID | Title | Status |
|----|-------|--------|
| M-1 | Concurrent timeline-update race | ✅ Fixed — promise-chained apply queue |
| M-2 | Pydantic validator side effects | ⚠️ Renamed (acceptable, see note) |
| Md-1 | `load_ffmpeg_frame` swallowed stderr | ✅ Fixed — stderr captured + raised |
| Md-2 | ImageStudio forced `edit_image` mode | ✅ Fixed — explicit `launchRequest` |
| Md-3 | VideoStudio forced `extend_clip` mode | ✅ Fixed — explicit `launchRequest` |
| Md-4 | SSE reconnects on project/timeline switch | ✅ Fixed — deps trimmed, ref used |
| Lo-1 | Hard-coded extend duration cap | ✅ Fixed — UI control 1–15s |
| Lo-2 | Duplicate lineage data | ✅ Fixed — sentinel ref instead of dup |
| Lo-3 | `setActiveView("Queue")` on every hook | ✅ Fixed — opt-in via `navigateToQueue` |
| Lo-4 | `refreshAssets` defensive dedup | ✅ Moved to backend (note below) |

### Notes on the fixes

**M-1** — The promise-chained `timelineApplyQueueRef` in [App.jsx:585-589](apps/web/src/App.jsx#L585) is the right shape: each `applyCompletedTimelineGeneration` is appended to a serial chain, eliminating the GET→mutate→PUT race. The companion `selectedTimelineIdRef` keeps the post-save `setActiveTimeline` decision based on the *current* selection rather than a stale closure. Clean fix. The `.catch((err) => setError(err.message))` at the chain root is good — one failure won't break the chain.

**M-2** — Renamed `validate_ranges` → `validate_ranges_and_populate_version_metadata` ([timelines.py:67](apps/api/sceneworks_api/timelines.py#L67)). Doesn't split the responsibilities, but the new name flags the side effect honestly. Fine for now; if more mutating logic accumulates on this model, consider extracting a separate `@model_validator` per concern.

**Md-1** — Now captures stderr via `subprocess.PIPE`, takes the last 6 lines, and raises a `RuntimeError` with the message ([video_adapters.py:374-382](apps/worker/scene_worker/video_adapters.py#L374)). Also surfaces missing-ffmpeg explicitly. Note: this is a behavior change — `load_source_frame` callers that previously got `None` on ffmpeg failures will now see an exception bubble up. That's the right tradeoff (errors should be loud), and `run_frame_extract`'s try/except wraps it cleanly into a job failure.

**Md-2 / Md-3** — Replaced the on-selection mode coercion with an explicit `launchRequest` prop carrying `{id, view, assetId, mode}`. Each "Send" click generates a new UUID so the receiving studio's effect fires once. Selection alone no longer mutates the mode — exactly the UX shape I was hoping for.

**Md-4** — Effect deps reverted to `[access.authRequired, authenticated, token]`. `selectedTimelineIdRef` carries the current value into the handler without forcing reconnects. EventSource lifetime now correctly tracks auth, not navigation.

**Lo-1** — New "Extension duration" input bound to `extensionDuration` state (range 1-15, step 0.5), feeding into `extendSelectedClip` ([EditorScreen.jsx:276](apps/web/src/screens/EditorScreen.jsx#L276)). Matches LTX-2.3's `hardMaxDuration: 15`.

**Lo-2** — Replaced the duplicated `timelineContext` block in `recipe.normalizedSettings` with `timelineContextRef: "lineage.timeline"` — a single source of truth in lineage, pointer in the recipe ([video_adapters.py:560](apps/worker/scene_worker/video_adapters.py#L560)). Slightly novel convention (a string sentinel inside JSON) but unambiguous.

**Lo-3** — `createVideoJob` now takes `{ navigateToQueue = true }`. Editor's `queueTimelineVideoJob` passes `false`, and `extractTimelineFrame` no longer changes view either. Users stay in the editor; the existing `timelineNotice` strip already tells them the job is queued.

**Lo-4** — The frontend `Map` dedup was removed and the dedup logic moved to the backend `list_assets` ([assets.py:88-128](apps/api/sceneworks_api/assets.py#L88)) via a `seen_asset_ids` set. That's the right architectural choice — uniqueness is a server contract, not a frontend defense. One small thing this *doesn't* answer is **why** duplicates were appearing. My suspicion: the V-4 auto-reindex walks `ASSET_FOLDERS` which includes both the original asset folders *and* the `trash` folder ([project_db.py:12](packages/shared/sceneworks_shared/project_db.py#L12)) via the shared `_asset_sidecars` helper; a recently-trashed asset whose sidecar got copied into `trash/<id>/` would surface twice if the original folder still has the (deleted) media file's sidecar. Worth a follow-up to confirm the trash-flow leaves no orphan sidecar in the original folder, but the defensive dedup is fine.

### Verdict

**Approve.** All four major/medium items resolved cleanly, three of four low items resolved, the fourth is mitigated defensively at the right layer. The slice satisfies all five Phase 11 acceptance criteria and the new test coverage for the timeline validator and frame-extract endpoint is solid. Ready to merge.

PR #18, opened 2026-05-17
[codex] Build Character Studio

michaeltrefry:
## Summary

Implements Shortcut epic 1089, SceneWorks: Character Studio.

- Adds sidecar-backed Character APIs for create, list, detail, update, archive, references, approved references, looks, character LoRAs, and test jobs.
- Adds the Character Studio UI with list/detail editing, reference approval, look management, LoRA attachment/editing, and test output attachment.
- Wires Character and Look selections into Image Studio and Video Studio generation payloads.
- Adds a character schema and focused backend tests.

## Validation

- `python -m pytest`
- `npm test` in `apps/web`
- `npm run check`
- Browser verification against this branch's API for Character Studio render and Image/Video character selection hooks

## Notes

The working tree had an unrelated pre-existing edit in `config/manifests/builtin.models.jsonc`; it was intentionally left unstaged and is not part of this PR.


michaeltrefry:
Out-of-story addition: narrowed the LTX-2.3 model manifest download to only \ltx-2.3-22b-dev.safetensors\ instead of the full Hugging Face repo. This keeps the intended checkpoint download around 46 GB instead of pulling the entire ~146 GB repository.

michaeltrefry (commented):
## Code review — PR #18 (Character Studio, Epic 1089)

Reviewed the single commit `e5afe4e` (1,773 / 8) covering Phase 12. Tests at PR HEAD: **36/36 pytest, 2/2 vitest** ✅.

### Phase 12 acceptance coverage

| Criterion | Status |
|---|---|
| User creates a Character | ✅ `POST /characters` + `submitNewCharacter` |
| User adds reference images | ✅ `POST /characters/{id}/references` + reference grid |
| User marks approved references | ✅ `PATCH /references/{asset_id}` with `approved` toggle |
| User creates a Look | ✅ `POST /characters/{id}/looks` + look composer |
| Character can be selected in generation forms | ✅ Image/VideoStudio character/look selects, `characterId`/`characterLookId` in payloads |
| Test Character produces sample outputs (once adapter supports it) | ✅ UI works; **adapter does not condition on character data yet — see M-1** |

---

### Major

**M-1. Worker silently ignores all character payload fields** ([apps/worker/scene_worker/image_adapters.py](apps/worker/scene_worker/image_adapters.py))

`grep -n character apps/worker/scene_worker/*.py` returns nothing. The API stores `characterId`, `characterLookId`, `mode="character_image"`, `advanced.characterName/characterType/approvedReferenceIds/look`, and the character's `loras` in the job payload, and the recipe sidecar dutifully records them. But the worker's `ZImageDiffusersAdapter.generate` only branches on `mode == "edit_image"` (for img2img) — `character_image` falls through to the same text-to-image pipeline as `text_to_image`. No reference images are loaded, no LoRA weights are loaded, no character conditioning happens.

The Phase 12 acceptance explicitly defers this: *"Test Character produces/rates sample outputs once image adapter supports it"* — so this PR is on plan. **But the UI ("Test Character" button, character-image generation form) currently produces a generic text-to-image with the character's name in metadata only, with no visible affordance telling the user that.** Two reasonable mitigations before merge:

- Add a short "Character conditioning is recipe-only in this build; outputs are a text prompt with the character name attached" note in the Test Character panel and in the Image/Video studio's character row.
- Or stamp `recipe.normalizedSettings.characterConditioningActive: false` in the worker so post-hoc analysis can tell the two apart.

Either way, please confirm this gap is tracked as a follow-up (next epic = LoRA loading, presumably).

---

### Medium

**Md-1. `copy_lora_into_project` silently swallows missing source files** ([apps/api/sceneworks_api/characters.py:170-181](apps/api/sceneworks_api/characters.py#L170))

```python
if not source_path.exists() or not source_path.is_file():
    return source_path_text, False
```

The attach call then writes a `link` with `copiedIntoProject: False` and `projectPath: null`, and the user gets a successful 201 response with no error indication. The Character Studio LoRA row shows `link.scope` instead of the project-copy badge, so visually a missing-file attach is indistinguishable from a `global` scope attach. Prefer `raise HTTPException(400, detail="LoRA source path not found: ...")` so the frontend can surface the failure.

**Md-2. `attach_lora` accepts arbitrary `sourcePath` and copies the file into the project** ([apps/api/sceneworks_api/characters.py:411-432](apps/api/sceneworks_api/characters.py#L411))

The endpoint `shutil.copy2`s whatever path the client provides into `project_path/loras/characters/<id>/`. On a single-user local stack the realistic exploit surface is small, but the API performs no path validation — any readable file (`~/.ssh/id_rsa`, `/etc/passwd`) can be copied into a project folder by a token-holding client. Worth either constraining to a configured "lora import root" or at minimum checking the source's parent against `settings.data_dir` and a user-managed import directory.

**Md-3. `archive_character` uses `DELETE` but doesn't delete** ([characters.py:283-289](apps/api/sceneworks_api/characters.py#L283))

The endpoint flips `status.archived = true` and keeps the sidecar on disk. `DELETE` semantically means removal; a tool watching the API would conclude the character is gone. Either change to `POST /{id}/archive` (more accurate) or add a separate `DELETE /{id}/purge` like the assets flow now has, and let `DELETE` mean "soft archive + tombstone" with `purge` doing the actual removal.

**Md-4. `hydrate_character` triggers O(R) DB lookups per character read** ([characters.py:121-129](apps/api/sceneworks_api/characters.py#L121))

For a character with N references, each `GET /characters/{id}` and each row in `GET /characters` calls `asset_summary` → `find_asset_sidecar_path` → DB query, one per reference. A character with 30 references means 30 DB hits. Batch this with a single `select * from assets where id in (?,?,?,...)` in `find_asset_records(project_path, asset_ids)`. (This becomes much more visible if Phase 13's tracker hits the character endpoints in tight loops.)

---

### Low

**Lo-1. Asset metadata denormalizes character link state, with no reverse-sync guard** ([characters.py:142-167](apps/api/sceneworks_api/characters.py#L142))

`update_asset_character_link` writes `metadata.characterReferences[{characterId, approved, role, linkedAt, approvedAt}]` into the asset sidecar. The character sidecar is the source of truth for the same fields. If a user edits the character sidecar by hand (or a future feature does), the asset metadata becomes stale. Today the API is the only writer, so it's consistent — just worth flagging as a maintenance smell.

**Lo-2. CharacterStudio.jsx is 566 lines** ([apps/web/src/screens/CharacterStudio.jsx](apps/web/src/screens/CharacterStudio.jsx))

Inheriting from F-021's split-main-module work, this is close to the "needs splitting" threshold. Reasonable now (single feature surface), but consider extracting `CharacterReferences`, `CharacterLooks`, `CharacterLoras`, `CharacterTest` panels in a follow-up before another epic adds more sections.

**Lo-3. `submitLora` crashes if `loraId` is stale** ([CharacterStudio.jsx:141-153](apps/web/src/screens/CharacterStudio.jsx#L141))

```js
const lora = loras.find((item) => item.id === loraId);
await attachCharacterLora(selectedCharacter.id, {
  loraId: lora.id,  // throws if lora is undefined
  ...
});
```

The submit button is disabled while `!loraId`, so this is hard to hit in practice — but if the lora list refreshes between the user picking and submitting (e.g., a delete elsewhere), `lora` will be `undefined` and the destructure crashes. Add an early `if (!lora) return;`.

**Lo-4. New Look inherits the currently-selected Look's `recipeSettings`** ([CharacterStudio.jsx:120-133](apps/web/src/screens/CharacterStudio.jsx#L120))

```js
recipeSettings: selectedLook?.recipeSettings ?? {},
```

`selectedLook` is the *Test Character* `testLookId` selection. Creating a new look while the user has "Hero" selected gives the new look a copy of Hero's recipe — clever but unexpected. Either clarify in the UI ("Inherits from selected look") or start fresh from `{}`.

**Lo-5. Test Character form hard-codes count/width/height** ([CharacterStudio.jsx:159-167](apps/web/src/screens/CharacterStudio.jsx#L159))

`count: 4, width: 1024, height: 1024` — the API allows variation but the UI doesn't expose controls. Minor UX gap.

**Lo-6. Repeated try/catch/setError pattern in App.jsx** (12 new character methods at [App.jsx:440-633](apps/web/src/App.jsx#L440))

Cosmetic. A `withCharacterApi(fn)` wrapper would halve the boilerplate, but the explicit form is also readable.

---

### Info

- **Schema:** `character.schema.json` is well-formed and documentary (not validated at write time, consistent with the rest of the codebase).
- **Tests:** `test_characters.py` covers CRUD, reference approve/archive round-trip, and LoRA copy. Good shape, plus the asset-sidecar back-link round-trip is asserted.
- **App.jsx wiring:** the new `studioLaunch.characterId` branch is handled cleanly in both ImageStudio and VideoStudio without forcing mode changes outside the explicit "Send" flow (consistent with the PR #17 cleanup).
- **`update_reference` notes-clear** can't set notes to `None` because `model_dump(exclude_none=True)` drops it. To clear, send `""`. Probably fine.

---

### Verdict

**Approve** with one strong ask: **address M-1** by either adding UI copy that explains the test-character output is recipe-only today, or stamping the recipe to make it post-hoc clear. The Phase 12 acceptance allows the gap, but the user-facing flow doesn't communicate it. Md-1/Md-2 are worth a quick second commit before merge; Md-3/Md-4 and Lo-* can be follow-ups.

PR #20, opened 2026-05-17
[codex] SceneWorks Rust backend migration spine

michaeltrefry:
## Summary

This draft PR tracks the staged Rust backend migration epic on a long-lived branch.

Current slice:
- Adds Rust migration contract fixtures for the existing Python API surface, job/worker protocol strings, SSE events, and persisted sidecar shapes.
- Adds a pytest drift guard so future Python contract changes must update the migration baseline intentionally.

## Impact

The fixtures give the upcoming Rust implementation executable compatibility targets while preserving the current frontend and worker contracts.

## Validation

- `python -m pytest`

PR #22, opened 2026-05-18
[codex] Complete multi-GPU polish

michaeltrefry:
## Summary

Completes the SceneWorks Multi-GPU Polish epic by tightening worker-per-GPU behavior and improving queue visibility.

- Adds loaded-model-aware auto GPU claiming in both Python and Rust queue stores while preserving FIFO for explicit GPU selections and utility jobs.
- Reports warm model IDs from the Z-Image worker cache and improves loading progress around cached/downloaded model files.
- Improves image/video failure messages for GPU OOM and missing Hugging Face model-file cases.
- Makes the Queue screen show clearer worker warmth, busy state, and queued job waiting/blocked reasons.
- Includes the current branch's HF cache persistence, compose defaults, and Rust SSE keepalive safety change.

## Validation

- `python -m pytest tests/test_jobs_store.py tests/test_worker_runtime.py tests/test_worker_gpu.py -q`
- `npm --prefix apps/web test -- --run`
- `node scripts/check-compose-config.mjs`
- `cargo test -p sceneworks-core --test jobs_store`
- `cargo test -p sceneworks-rust-worker`
- `python -m pytest tests/test_python_rust_api_parity.py tests/test_rust_api_worker_smoke.py -q`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -55,6 +55,10 @@ SCENEWORKS_LEGACY_FFMPEG_JOBS=0
 # Optional Hugging Face token for gated model or LoRA repos.
 HF_TOKEN=
 
+# Persist Diffusers/Hugging Face model caches inside the mounted data volume.
+SCENEWORKS_HF_HOME=/sceneworks/data/cache/huggingface
+SCENEWORKS_HUGGINGFACE_HUB_CACHE=/sceneworks/data/cache/huggingface/hub
+
 # Worker adapter override: auto, procedural, z_image_diffusers, or qwen_image.
 SCENEWORKS_IMAGE_ADAPTER=auto
 
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ API volume contracts are shared across Python and Rust:
 - `${SCENEWORKS_DATA_BIND:-./data}:/sceneworks/data` read/write for projects, models, LoRAs, and cache-backed app data.
 - `${SCENEWORKS_CONFIG_BIND:-./config}:/sceneworks/config:ro` read-only for manifests and app configuration.
 - `./data/cache/jobs.db` is the shared queue database for both runtimes, preserving existing compose queue history across migration flips and rollback.
+- `./data/cache/huggingface` persists Diffusers/Hugging Face model downloads across worker container rebuilds and restarts.
 
 Both API runtimes expose `GET /api/v1/health`; Compose checks it with `curl`
 inside the container so dependent services wait for the selected implementation.
@@ -125,6 +126,10 @@ owns the model, LoRA, and FFmpeg utility families. Set
 The Python worker ID changed from `worker-gpu-auto-0` to
 `python-inference-worker-0`; existing queue databases may retain the old worker
 row until the stale-worker sweep marks it offline.
+When multiple GPU children are registered, auto GPU jobs may be claimed by a
+worker that already reports the requested model as warm before falling back to
+FIFO order. Explicit GPU selections and utility jobs keep their normal FIFO
+claim order.
 Both Docker worker images install Debian Bookworm `ffmpeg`; host-mode workers
 use the `ffmpeg` found on `PATH`. Set `HF_TOKEN` when downloading from gated
 Hugging Face repositories.
--- a/apps/api/sceneworks_api/jobs_store.py
+++ b/apps/api/sceneworks_api/jobs_store.py
@@ -49,6 +49,13 @@ def choose_claimable_job(rows: list[sqlite3.Row], capabilities: set[str], loaded
     if first["type"] in NON_GPU_JOB_TYPES or first["requested_gpu"] != "auto":
         return first
 
+    explicit_gpu_job = next(
+        (row for row in compatible if row["type"] not in NON_GPU_JOB_TYPES and row["requested_gpu"] != "auto"),
+        None,
+    )
+    if explicit_gpu_job is not None:
+        return explicit_gpu_job
+
     return next((row for row in compatible if job_matches_loaded_model(row, loaded_models)), first)
 
 
--- a/apps/api/sceneworks_api/jobs_store.py
+++ b/apps/api/sceneworks_api/jobs_store.py
@@ -17,6 +17,41 @@ NON_GPU_JOB_TYPES = ("model_download", "lora_import")
 MAX_JOB_ATTEMPTS = 5
 
 
+def desired_model_keys(payload: dict) -> set[str]:
+    keys: set[str] = set()
+    model = payload.get("model")
+    if isinstance(model, str) and model:
+        keys.add(model)
+    repo = payload.get("repo")
+    if isinstance(repo, str) and repo:
+        keys.add(repo)
+    advanced = payload.get("advanced")
+    if isinstance(advanced, dict):
+        for field in ("modelRepo", "repo"):
+            value = advanced.get(field)
+            if isinstance(value, str) and value:
+                keys.add(value)
+    return keys
+
+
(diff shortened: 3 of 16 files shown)


michaeltrefry (commented):
# Code Review — vs Epic [sc-1091 SceneWorks: Multi-GPU Polish](https://app.shortcut.com/trefry/epic/1091)

Scope: 15 files / +457 / -26. PR is branched off `48813ef`, which is 4 commits ahead of current `main` @ `9300bdd` — confirm merge plan with PRs #20/#21.

---

## Acceptance-Criteria Mapping

| Story | AC | Status | Notes |
|---|---|---|---|
| **sc-1163** Worker-per-GPU concurrency | One worker per GPU | ✅ preexisting | Registration model |
| | Two GPU workers run jobs concurrently | ✅ | `active_gpu_job` filter is per-`assigned_gpu` in both runtimes |
| | No two GPU jobs on same GPU | ✅ | `claim_next_job` SQL excludes when current GPU has an active GPU job; tested |
| **sc-1164** Loaded-model preference | Worker reports loaded-model hints | ✅ | `ZImageDiffusersAdapter.loaded_models()` now returns both repo & model id (`apps/worker/scene_worker/image_adapters.py:222`) |
| | Auto scheduler prefers warm GPU | ✅ | New `choose_claimable_job` in `apps/api/sceneworks_api/jobs_store.py:43` and `crates/sceneworks-core/src/jobs_store.rs:1119`; covered by tests in both runtimes |
| | User-selected GPU override still wins | ⚠️ partial | See **M2** — test only covers the case where the explicit job is *first* in queue |
| **sc-1165** Queue display | Per-GPU running/waiting state | ✅ | Worker cards now show `Busy with <jobId>` / `Ready` and warm-model list (`apps/web/src/screens/QueueScreen.jsx:115`) |
| | Top bar shows active GPU status | ✅ preexisting | `apps/web/src/App.jsx:1011–1013` — *not* in this PR; was added earlier |
| | Blocked jobs explain GPU/download/dependency/user-selected | ⚠️ partial | See **M3** — GPU/capability/explicit-device covered; \"download\" and \"dependency\" reasons not differentiated |
| **sc-1166** Failure messages | OOM gives model/job context + next actions | ✅ | `friendly_failure` (`apps/worker/scene_worker/runtime.py:145–172`) |
| | Missing-model points to Model Manager/download | ⚠️ minor | Message names HF_TOKEN and \"model is installed\" but doesn't name the Model Manager UI or the `model_download` job by name (**L4**) |
| | Failure visible in Queue + job detail | ✅ | `update_job` writes message/error; `JobRow` displays them for non-queued statuses |

---

## Findings

### Medium

**M1 — SSE heartbeat wire format diverges from Python**  `apps/rust-api/src/lib.rs:57`, `:1560–1564`

`HEARTBEAT_SSE_TEXT` went from `\"event: heartbeat\\ndata: {}\\n\\n\"` to `\"{}\"`. The fix is real — axum's `KeepAlive::text` requires no `\\r`/`\\n` — but the new behavior emits an SSE *comment line* (`: {}\\n\\n`) instead of a real `event: heartbeat` event. Python still emits the full event (`apps/api/sceneworks_api/jobs.py:139`), and the renamed test (`heartbeat_event_matches_python_wire_shape` → `heartbeat_keepalive_text_is_axum_safe`) is an explicit acknowledgement of the divergence. The parity snapshot tests don't cover SSE wire shape, so this silently regresses parity that prior epics deliberately maintained. Options: (a) merge a synthesized `heartbeat` event into the SSE stream instead of using `KeepAlive`, (b) switch Python to comment-style keepalives to re-converge, or (c) add a parity assertion for the heartbeat frame so the divergence is at least intentional and tested.

**M2 — Loaded-model preference can bypass an older explicit-GPU job**  `apps/api/sceneworks_api/jobs_store.py:43–52`, `crates/sceneworks-core/src/jobs_store.rs:1119–1138`

Trace for worker `gpu-0` with loaded `[X]` and queue `[A:auto/model=Y (oldest), B:explicit gpu-0/model=Y, C:auto/model=X]`: `first=A` is auto-GPU, so the loop scans for a loaded-match. `A` and `B` don't match (B is excluded because it's explicit), `C` matches → `C` is claimed, `B` is skipped despite being older and explicitly targeting this GPU. The AC (\"user-selected GPU override still wins\") most naturally reads as \"the user's GPU choice is honored\" rather than \"explicit jobs run before younger auto jobs,\" so this *may* be intentional — but the existing test only locks in the case where the explicit job is first. Recommendation: add a test for `[auto-other, explicit, auto-match]` ordering that captures the intended behavior either way, so future refactors don't silently flip it.

**M3 — `jobWaitingMessage` covers only a subset of the AC's \"blocked\" reasons**  `apps/web/src/screens/QueueScreen.jsx:24–44`

The AC enumerates GPU, download, dependency, and user-selected device. The implementation handles GPU/capability/explicit-device; there's no distinct messaging for \"waiting because a `model_download` job is in flight\" or \"waiting on a dependency.\" If those states don't exist in the current scheduler, fine — but worth confirming explicitly and noting on sc-1165 that the AC bullet for \"download/dependency\" was deemed N/A or deferred (rather than leaving it ambiguous). This matches the **acceptance-criteria drift** pattern flagged in memory for prior epics.

### Low

**L4 — Missing-model error message could name the recovery path more specifically**  `apps/worker/scene_worker/runtime.py:156–166`

AC says \"point to Model Manager or download job.\" Current text says *\"Check that the model is installed, the Hugging Face repo is reachable, and HF_TOKEN is set.\"* Substantively correct, but a one-line nudge like *\"Open Model Manager or queue a model_download job to install it.\"* would directly satisfy the AC wording.

**L5 — `missing_model_markers` matcher is loose**  `apps/worker/scene_worker/runtime.py:155–160`

`\"model files\"` and `\"model not found\"` are broad enough to false-positive on unrelated errors that mention those phrases. Result is a misleading-but-still-actionable message; low risk.

**L6 — `huggingface_repo_cache_exists` is a presence-only heuristic**  `apps/worker/scene_worker/image_adapters.py:38–43`

Checks that `snapshots/` or `blobs/` *exists*, not that the snapshot contains the needed files. A partially-pruned cache still reads as \"cached,\" producing a misleading `Loading cached …` log message before diffusers re-downloads. The progress message is informational, so impact is just user confusion. Fine to ship.

**L7 — `repo` is not sanitized in cache-path construction**  `apps/worker/scene_worker/image_adapters.py:42`

`repo.replace('/', '--')` handles forward slashes but not `..` or backslashes. Since `repo` originates from request payload, a hostile value could resolve outside the cache root. Worst case the path doesn't exist and returns False — no read/write — so low impact, but a `Path.resolve()` containment check would be cheap insurance.

### Info

- **I8** — PR base (`48813ef`) is 4 commits ahead of current `main` (`9300bdd`). Confirm the merge order with PRs #20/#21 before merging this.
- **I9** — `ZImageDiffusersAdapter.loaded_models()` returns a sorted set with both repo and model id; the new test locks in the exact order `[\"Tongyi-MAI/Z-Image-Turbo\", \"z_image_turbo\"]`. If either identifier's case changes, that test breaks even though behavior is correct. Consider asserting set membership instead of list equality.
- **I10** — CI **parity** check is green at the time of review.

---

## Recommendation

**Approve with follow-ups.** All four stories' core ACs are met or substantially met, code is well-tested in both runtimes, and the Python↔Rust scheduler parity for the new loaded-model logic is solid. The blocking items are **M1** (silent SSE parity regression — at minimum add a test that asserts the intentional divergence) and a decision on **M2** (lock the explicit-vs-loaded ordering with an explicit test). **M3/L4** are AC-tightness issues that should be addressed before closing the stories to avoid the kind of acceptance-criteria drift previously seen in epics 1081–1085.

PR #23, opened 2026-05-18
[codex] Implement scoped LoRA catalog and selection

michaeltrefry:
﻿## Summary

Implements the completed scoped LoRA slice for `sc-1167`, with the related import and generation-selection foundations needed by the dependent stories.

- Merge builtin, global, and project LoRA manifests through `/api/v1/loras?projectId=...`.
- Route project LoRA imports into `project/loras/imports/...` and `project/loras/manifest.jsonc`, while preserving global imports under `data/loras` and the user manifest.
- Teach both Python and Rust workers to honor `targetDir` for LoRA imports.
- Refresh the web LoRA catalog per active project and expose compatible Image Studio LoRA selection with the simple-mode cap of builtin LoRAs plus up to 2 user/project LoRAs.
- Preserve LoRA scope in character attachment and show scope in the Models LoRA list.

## Story Scope

Primary completed story: `sc-1167` Implement global and project LoRA scopes.

Related partial progress: `sc-1168` import flow foundation and `sc-1169` selection/payload foundation. Runtime adapter-level LoRA loading and metadata editing remain tracked in Shortcut tasks.

## Validation

- `python -m pytest tests/test_models.py tests/test_image_generation.py`
- `npm --prefix apps/web test`
- `rustfmt --check apps/rust-worker/src/lib.rs`
- `cargo test -p sceneworks-rust-worker lora_file_and_directory_import_preserve_copy_semantics`
- Browser sanity check at `http://127.0.0.1:5173` for Image Studio rendering and console errors

Note: workspace-wide `cargo fmt --all -- --check` is currently blocked by pre-existing newline-style issues in unrelated Rust files.

PR #24, opened 2026-05-18
[codex] Add recipe preset support

michaeltrefry:
## Summary

Adds manifest-backed recipe preset support owned by the Rust API, plus Image Studio wiring so built-in presets can apply defaults, prompt fragments, and preset-managed LoRAs without exposing internal LoRA details in the simple UI.

## Changes

- Added Rust `/api/v1/recipe-presets` with built-in, user, and project preset manifest merging.
- Added Rust server-side `recipePresetId` resolution for prompt fragments, preset-managed LoRAs, preset model/default fields, and preset metadata.
- Added built-in and user recipe preset manifests plus a shared schema.
- Added preset-managed built-in LoRA manifest entries for Cinematic, Photoreal, Anime, Fantasy, and Product Shot presets.
- Updated Image Studio to fetch presets, apply defaults in the form, show preset descriptions/fragments/managed LoRAs, and submit `recipePresetId` for backend enforcement.
- Kept Python as the rollback runtime; recipe preset endpoints and server-side expansion are intentionally Rust-only.

## Validation

- `npm --prefix apps\web test -- --run`
- `python -m pytest tests\test_rust_migration_contract_fixtures.py tests\test_python_rust_api_parity.py::test_system_manifest_and_http_contracts tests\test_python_rust_api_parity.py::test_model_lora_and_image_job_contracts tests\test_python_rust_api_parity.py::test_job_state_transition_contracts`
- `python -m pytest tests\test_python_rust_api_parity.py::test_model_lora_and_image_job_contracts tests\test_rust_migration_contract_fixtures.py::test_api_surface_fixture_matches_python_openapi`
- `git diff --check`


michaeltrefry (commented):
# Code Review — vs remaining Epic [sc-1092](https://app.shortcut.com/trefry/epic/1092) stories ([sc-1170](https://app.shortcut.com/trefry/story/1170), [sc-1171](https://app.shortcut.com/trefry/story/1171))

Scope: 15 files / +636 / -20. Stacked on PR #23 (base = `30148b1`). CI **pending** at review time. Both target stories are already marked **Done** in Shortcut.

Scope note: this PR explicitly does not address the still-open follow-ups on **sc-1168** (LoRA metadata editing) or **sc-1169** (adapter-level LoRA loading). Recipe-managed LoRAs route through the same payload as user LoRAs, so the adapter gap from sc-1169 applies to preset LoRAs too — they're sent in the payload, but no runtime adapter actually loads them yet.

---

## Acceptance-Criteria Mapping

| Story | AC | Status | Notes |
|---|---|---|---|
| **sc-1170** | Data model can store reusable saved recipes under `recipes/` | ⚠️ partial | PR adds `project/recipes/presets.jsonc` (preset manifest), not per-recipe files under `recipes/<id>.jsonc`. Reads as "presets at project scope" rather than "saved recipes". See **I1**. |
| | UI does not block later save-as-preset workflow | ✅ | Frontend reads via API; user manifest exists and can be appended to. |
| | Custom preset creation can be deferred without reworking built-in presets | ✅ | Built-in / user / project tiers cleanly separated; merge-by-id pattern matches LoRA scoping. |
| **sc-1171** | Presets include Cinematic, Anime, Fantasy, Photoreal, Product Shot | ✅ | All five present in [config/manifests/builtin.recipe-presets.jsonc](config/manifests/builtin.recipe-presets.jsonc). |
| | Presets can set model defaults, prompt fragments, negative prompt defaults, built-in LoRAs | ✅ | Schema supports `model`, `defaults.{count,resolution,negativePrompt}`, `prompt.{prefix,suffix}`, `builtInLoras`. |
| | Preset effects are visible enough for users to understand what changed | ❌ | See **M2** — prompt fragments are appended invisibly at submit, preset LoRAs are hidden, `ui.description` isn't surfaced. User picks "Cinematic" with no UI feedback on what changes. |
| (Epic) | Rust backend owns preset contracts | ⚠️ partial | Rust API serves the manifest with parity ✓, but preset application (prompt expansion, LoRA injection) is frontend-only — see **M4**. |

---

## Findings

### Medium

**M1 — Preset-default effect clobbers user edits on every background refresh**  `apps/web/src/screens/ImageStudio.jsx:215–228`

```jsx
useEffect(() => {
    if (!selectedRecipePreset) return;
    if (selectedRecipePreset.model && imageModels.some(...)) setModel(...);
    if (defaults.count) setCount(Number(defaults.count));
    if (defaults.resolution) setResolution(defaults.resolution);
    setNegativePrompt(defaults.negativePrompt ?? "");
}, [selectedRecipePreset, imageModels]);
```

`selectedRecipePreset` is derived from `availableRecipePresets.find(...)`. `availableRecipePresets` is memoed on `[mode, recipePresets]`. Every time `App.jsx` re-fetches recipe presets (project switch, full refresh), `recipePresets` is a new array of new object references, so `selectedRecipePreset` is a fresh reference, the effect fires, and the user's just-typed `count` / `resolution` / `negativePrompt` are stomped back to preset defaults. Same thing happens on every `imageModels` refresh. Suggested fix:

```jsx
}, [selectedRecipePreset?.id]);
```

…so the reset only fires when the user actually changes the preset.

**M2 — sc-1171 "Preset effects visible enough" AC not met**  `apps/web/src/screens/ImageStudio.jsx:200`, `:259–262`

Three layers of invisibility:
- `applyPresetPrompt(prompt, selectedRecipePreset)` runs at submit and silently prepends/appends the preset's prompt fragments. The textarea never reflects what will actually be sent; the user only sees the expanded prompt after the job lands.
- Preset LoRAs are filtered out of `compatibleLoras` via `lora.presetManaged` and stitched back in only at submit (`presetLoras`). User can't see *which* LoRAs the preset is going to apply.
- Each preset carries `ui.description` ("Balanced cinematic color, contrast, and detail.") in the manifest, but ImageStudio never renders it.

Cheapest fix that satisfies the AC: render the selected preset's `ui.description` under the Style select, plus a small "Adds: cinematic lighting, …" / "Uses LoRA: Cinematic Detail" disclosure. Anything that makes the picker non-magical.

**M3 — `fallbackRecipePresets` silently masks API failures and project-scoped presets**  `apps/web/src/screens/ImageStudio.jsx:4–10`, `:127`

If `/api/v1/recipe-presets` errors or returns `[]`, `availableRecipePresets` uses the hard-coded `fallbackRecipePresets` from the JS file — five generic Cinematic/Photoreal/Anime/Fantasy/Product entries with empty `builtInLoras` and no prompt fragments. Any project-scoped preset and any user override is invisible. App.jsx already sets an `error` state via `refreshRecipePresets`'s try/catch; the picker should surface that or fall back to "Presets unavailable" rather than pretending defaults exist. The fallback also duplicates manifest data inline — guaranteed drift as the JSONC manifest evolves.

**M4 — Preset contract is read by frontend but not enforced by backend**  `apps/web/src/screens/ImageStudio.jsx:210–222`

Per epic: *"Built-in presets, model defaults, prompt fragments, LoRA references, and preset APIs should be owned by **Rust backend contracts** and consumed by the UI/runtime adapters."* The PR delivers the *catalog* contract but not the *application* contract — the job payload arrives at the worker already-expanded (prompt has the suffix baked in, `loras[]` already has preset entries), and `advanced.recipePresetId` is a freely-chosen string with no server-side validation. A misbehaving client (or just a stale frontend) can send `recipePresetId: "cinematic"` with zero preset LoRAs and a completely different prompt, and the runtime will accept it.

Two options that would close this gap: (a) backend resolves the preset server-side from `recipePresetId` and rewrites `prompt` + `loras[]` before queueing, OR (b) backend validates that `recipePresetId`'s expected `builtInLoras` / `prompt` match what arrived, rejecting if not. Either is a follow-up — worth noting on sc-1171 rather than silently deferring.

### Low

**L1 — Missing preset LoRA falls back to a phantom payload entry**  `apps/web/src/screens/ImageStudio.jsx:151–157`

```js
return lora ? serializeLora(lora, presetLora) : { id: loraId, name: loraId, scope: "builtin", weight: presetLora.weight ?? 0.8 };
```

If a preset references a built-in LoRA that the API didn't return (uninstalled, manifest drift, fetch error), the frontend invents a placeholder LoRA reference with no `triggerWords`, `compatibility`, or real metadata, and submits it. The worker will fail to resolve it. Either skip with a console warning or surface a visible "preset incomplete" indicator.

**L2 — Preset switch always overwrites `negativePrompt` to `""`**  `apps/web/src/screens/ImageStudio.jsx:226`

`setNegativePrompt(defaults.negativePrompt ?? "")` — switching to a preset that doesn't declare `negativePrompt` wipes the user's existing one. Most builtins set one, so this is mostly latent, but a custom preset without `negativePrompt` will silently clear user input. Use a guard: only set when the preset actually provides a value.

**L3 — Recipe-preset schema is type-only**  `packages/schemas/recipe-preset.schema.json:1–14`

`items: { "type": "object" }` accepts anything that's an object. No constraint on `id`, `name`, `defaults` shape, `builtInLoras` shape. Fine as a v0 schema, but worth tightening before users start writing custom presets so editor validation actually helps.

**L4 — Mode-filter fallback returns *all* presets when none match**  `apps/web/src/screens/ImageStudio.jsx:127`

```js
const matchingMode = items.filter((preset) => !preset.modes?.length || preset.modes.includes(mode));
return matchingMode.length ? matchingMode : items;
```

If a user opens a mode that no preset opts into (e.g., a future mode), the Style dropdown shows presets declaring incompatible modes anyway. Mild — defaulting to "no presets" might confuse, but defaulting to "wrong presets" obscures the AC's mode-targeting intent.

### Info

- **I1** — sc-1170 AC wording reads "saved **recipes** under `recipes/`" — this PR adds `recipes/presets.jsonc` (preset manifest), not per-recipe files. Worth a one-line note on the story confirming the team accepts presets-as-recipes equivalence, or a follow-up to add per-recipe storage. Same acceptance-criteria-drift pattern as flagged for epics 1081–1085.
- **I2** — `fallbackRecipePresets` in [ImageStudio.jsx:4–10](apps/web/src/screens/ImageStudio.jsx:4) duplicates manifest data inline. Will drift the moment the JSONC manifest is edited without a sync to the JS file. Consider sourcing from a shared JSON or deleting the fallback in favor of the M3 fix.
- **I3** — Style dropdown is sorted by `(scope, name)` — that puts "Anime" first instead of "Cinematic". If "Cinematic" was the prior default, users will see a behavior change. Small UX detail worth a note on sc-1171.
- **I4** — `recipePresetName` is duplicated into `advanced` alongside `recipePresetId`. Names can change; if a job is retried after a rename, the name is stale. Either drop it (frontend can resolve by ID) or accept the staleness explicitly.
- **I5** — `load_manifest_entries` (Rust) returns empty on NotFound ✓, so a deployment missing `user.recipe-presets.jsonc` won't 500.

---

## Recommendation

**Approve with follow-ups.** sc-1171's "create the presets" mechanics are well-built and parity is preserved across Python/Rust APIs. The two material gaps are **M2** (the visibility AC is genuinely unmet — user-facing change of a preset-selecting UI shouldn't be invisible) and **M1** (the silent-clobber effect is a real UX bug that will surface as soon as someone leaves Image Studio open for any length of time). **M4** is more strategic — worth deciding whether preset enforcement belongs in this epic or a later one and noting that decision on sc-1171 so it doesn't quietly drift. **I1** is the kind of AC-mapping note that has bitten previous closed epics; one-line confirmation on sc-1170 keeps the audit trail clean.

michaeltrefry (commented):
# Follow-up Review — commit `d39838a`

Re-reviewed after "Address recipe preset review feedback." Two structural pivots since last review:

1. **Rust-only backend** — the Python `recipe_preset_catalog`, `recipe_presets_router`, parity snapshot entry, parity test case, and Python tests were all deleted. Net file count dropped 15 → 10. Per the team direction (stop adding parity for new development), this is intentional.
2. **Server-side preset resolution** — the frontend stopped expanding prompts and injecting preset LoRAs into job payloads. The frontend now just sends top-level `recipePresetId`; the Rust API resolves the preset and rewrites `prompt` + `loras[]` inside `create_image_job`.

CI **parity passes** (4m20s). ✅

---

## Previous findings — status

| # | Item | Status | Evidence |
|---|---|---|---|
| **M1** | `useEffect` clobbers user edits on every refresh | ✅ Resolved | Dep changed to `[selectedRecipePreset?.id]` ([ImageStudio.jsx:164](apps/web/src/screens/ImageStudio.jsx:164)) — fires only when the user changes preset. |
| **M2** | Preset effects invisible to user | ✅ Resolved | New guidance-strip below the Style select renders `ui.description`, `Adds: <prompt fragments>`, `Uses LoRA: <preset LoRA names>`, and `Preset incomplete` when LoRAs are missing ([ImageStudio.jsx:317–333](apps/web/src/screens/ImageStudio.jsx:317)); test asserts all three strings. |
| **M3** | `fallbackRecipePresets` masked API failures | ✅ Resolved | Hard-coded fallback deleted; `availableRecipePresets` is just the filtered list; empty case renders disabled select with "Presets unavailable" + matching guidance-strip copy. |
| **M4** | Backend doesn't enforce preset contract | ✅ Resolved (partial) | New `apply_recipe_preset_to_image_payload` in [apps/rust-api/src/lib.rs:1190–1276](apps/rust-api/src/lib.rs:1190) resolves preset server-side, expands the prompt (`preset_prompt`), injects preset LoRAs deduped against user selection, and writes `advanced.recipePresetId` plus `advanced.presetMissingLoras`. Test exercises end-to-end. See **NEW-L2** for the residual gap. |
| **L1** | Phantom preset LoRA payload | ✅ Resolved | Frontend no longer sends preset LoRAs; backend resolves them from the catalog and surfaces unresolved IDs via `advanced.presetMissingLoras`. |
| **L2** | `negativePrompt` always overwritten | ✅ Resolved | `Object.prototype.hasOwnProperty.call(defaults, "negativePrompt")` guard ([ImageStudio.jsx:161–163](apps/web/src/screens/ImageStudio.jsx:161)). |
| **L3** | Recipe-preset schema is type-only | ✅ Resolved | Schema now requires `id`/`name`, validates id pattern, mode array, resolution `^[0-9]+x[0-9]+$`, weight `[-2, 2]`, prompt prefix/suffix, builtInLoras `oneOf` string/object. |
| **L4** | Mode-fallback returned all presets | ✅ Resolved | `availableRecipePresets` is just the filtered list; no "return all if empty" branch. |
| **I3** | Sort put "Anime" before "Cinematic" | ✅ Resolved | New `order` field on each builtin (10/20/30/40/50); sort key is now `(scope, order, name)`. Cinematic leads as expected. |
| **I4** | `recipePresetName` staleness in `advanced` | ✅ Resolved | Backend writes only `recipePresetId`; explicitly `advanced.remove("recipePresetName")`. |
| **I2** | `fallbackRecipePresets` duplication | ✅ Resolved (with M3) |  |
| **I1** | sc-1170 "recipes/<id>.jsonc" wording | ❓ Still applies | This PR adds `project/recipes/presets.jsonc` (preset manifest), not per-recipe files. Worth a one-line confirmation on [sc-1170](https://app.shortcut.com/trefry/story/1170) that the team accepts presets-as-recipes as satisfying the AC. |

---

## New findings

### Low

**NEW-L1 — Initial `refreshData()` is brittle to any single fetch failing**  `apps/web/src/App.jsx:266–278`

`refreshData` now does `Promise.all([... apiFetch("/api/v1/recipe-presets", token) ...])`. If any one endpoint 404s or errors, the entire promise rejects and the catch sets `error` — meaning projects/jobs/workers/models/loras all fail to populate on initial load. For the Rust deployment this is fine (recipe-presets returns `[]` on no manifest), but if anyone ever launches against the Python API (which no longer has this endpoint after the pivot) they'll get an empty broken app shell. Cheap fix: wrap each fetch with `.catch(() => default)` so one missing endpoint can't blank the whole UI. Same pattern as `refreshRecipePresets` already uses for the project-change refresh. Skip if Python compat is explicitly out of scope going forward — worth a one-line decision either way.

**NEW-L2 — Server-side preset enforcement is partial**  `apps/rust-api/src/lib.rs:1190–1276`

`apply_recipe_preset_to_image_payload` rewrites `prompt`, `loras[]`, `stylePreset`, and `advanced.recipePresetId`. It does **not** override `model`, `count`, `width`, `height`, or `negativePrompt` from the preset's `defaults` or `model` field. Today the frontend `useEffect` applies all of those client-side and submits them, so the typical flow is correct. But a stale frontend (preset edited after the form was rendered) or any non-browser client posting `recipePresetId: "cinematic"` with a different `model` or `count` will run with the wrong settings while the job claims to have used the preset. If full enforcement is intended, mirror the prompt/LoRA pattern for `defaults.{count, resolution, negativePrompt}` and `model`. If partial enforcement is acceptable (the user-overridable knobs), worth a one-line code comment on `apply_recipe_preset_to_image_payload` documenting which fields are preset-owned vs. client-owned.

### Info

- **NEW-I1** — `category` is set on each new builtin LoRA (`enhance` / `style`) but is not in the LoRA schema or referenced by any code path I can see. Either wire it into the picker / family filter, or drop it from the manifest to avoid confusing future readers.
- **NEW-I2** — The Rust test fixture sets `"prompt": { "suffix": "cinematic lighting" }` and the integration test asserts `prompt` equals `"city at night, cinematic lighting"`, which locks in the comma-join + trim behavior of `preset_prompt`. Nice anchor.
- **NEW-I3** — `ImageJobRequest.project_id` is a required `String` (not `Option`), so the new `recipe_preset_catalog(state, Some(&payload.project_id))` call is safe under normal flow. Empty-string projectId would surface as a 500 from `project_path_for_id`; if that's worth a friendlier 400, it's a tiny additional `is_empty()` check.
- **NEW-I4** — Architecturally Python is now the older sibling on this endpoint. Recommend a brief CODEGRAPH.md or README note that recipe-presets are Rust-only going forward, so the next contributor doesn't try to add the Python side back out of reflex.

---

## Recommendation

**Approve.** All eight prior findings (M1–M4, L1–L4) are genuinely addressed, schema is tightened, the visibility AC for sc-1171 is now met with concrete UI affordances, and the M4 server-side resolution is a real architectural improvement over the previous frontend-only application. NEW-L1 and NEW-L2 are residual gaps that depend on team direction (Python deprecation stance; full vs. partial preset enforcement) — both are reasonable to defer to follow-up Shortcut stories rather than block on. The only thing I'd still ask for before closing the epic is a one-line note on **sc-1170** confirming presets-as-recipes satisfies the AC wording (same drift hazard previous closed epics hit).

PR #26, opened 2026-05-18
[codex] Complete character API parity coverage

michaeltrefry:
﻿## Summary

Adds a dedicated Python/Rust parity test for the characters subsystem and records the new snapshots. The flow now covers all 17 character routes: create/list/get/update, both soft-archive paths, purge, reference attach/update/delete with asset metadata back-link cleanup, look create/update/delete, LoRA attach/update/delete with project copy behavior, and character test-job creation.

Also fixes a parity drift in the Rust character test-job handler: Python returns `projectName: null` for these jobs, so Rust now passes `None` instead of the project stem.

## Impact

This gives sc-1282/sc-1284 a real parity gate for the character API work, reducing the risk of deleting the Python API while character routes still drift.

## Review follow-up

Addressed the PR review coverage gap and nits:

- Added PATCH/DELETE/purge coverage for the remaining character routes.
- Verified reference deletion cleans up asset-sidecar `metadata.characterReferences`.
- Verified `DELETE /characters/:id/purge` removes the character sidecar file.
- Verified `POST /archive` and `DELETE /:id` produce equivalent archive responses.
- Added a comment explaining the LoRA source path relationship to the parity harness root.
- Added character-path error parity for missing character/look/lora IDs and invalid LoRA source payloads.

## Validation

- `python -m pytest tests/test_python_rust_api_parity.py -q` -> 9 passed
- `cargo test -p sceneworks-rust-api` -> 21 passed
- `cargo test -p sceneworks-core --test contract_roundtrip` -> 6 passed
- `cargo clippy -- -D warnings` -> clean
- `git diff --check` -> clean

Manual web Character Studio exercise was not run in this pass.

### DIFF
--- a/apps/rust-api/src/lib.rs
+++ b/apps/rust-api/src/lib.rs
@@ -1368,11 +1368,6 @@ async fn create_character_test_job(
         move |store| store.get_character(&project_id, &character_id)
     })
     .await?;
-    let project_name = project_call(state.clone(), {
-        let project_id = project_id.clone();
-        move |store| store.project_stem(&project_id)
-    })
-    .await?;
     let look = payload.look_id.as_deref().and_then(|look_id| {
         character
             .get("looks")
@@ -1444,7 +1439,7 @@ async fn create_character_test_job(
         state,
         JobType::ImageGenerate,
         Some(project_id),
-        Some(project_name),
+        None,
         job_payload,
         requested_gpu_or_auto(payload.requested_gpu),
     )
--- a/tests/fixtures/python_rust_api_parity/snapshots.json
+++ b/tests/fixtures/python_rust_api_parity/snapshots.json
@@ -59,6 +59,61 @@
     "id": "asset_fixture",
     "status": "purged"
   },
+  "asset sidecar after character reference": {
+    "createdAt": "<timestamp>",
+    "displayName": "fixture image.png",
+    "file": {
+      "duration": null,
+      "fps": null,
+      "height": null,
+      "mimeType": "image/png",
+      "path": "assets/uploads/fixture-image-<upload-suffix>.png",
+      "width": null
+    },
+    "generationSetId": null,
+    "id": "asset_fixture",
+    "lineage": {
+      "jobId": null,
+      "parents": [],
+      "sourceAssetId": null,
+      "sourceTimestamp": null
+    },
+    "metadata": {
+      "characterReferences": [
+        {
+          "approved": true,
+          "approvedAt": "<timestamp>",
+          "characterId": "character_fixture",
+          "linkedAt": "<timestamp>",
+          "role": "hero",
+          "source": "character-sidecar"
+        }
+      ]
+    },
+    "projectId": "project_fixture",
+    "recipe": {
+      "adapter": "api-upload",
+      "loras": [],
+      "mode": "upload",
+      "model": "manual-import",
+      "negativePrompt": "",
+      "normalizedSettings": {},
+      "prompt": "fixture image.png",
+      "rawAdapterSettings": {
+        "contentType": "image/png"
+      },
+      "seed": 0,
+      "stylePreset": "none"
+    },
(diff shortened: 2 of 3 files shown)


michaeltrefry:
## Code Review

### Verdict

**Approve with a follow-up required.** The change is correct and CI-green, but the PR delivers ~50% of what sc-1284 needs. Don't close sc-1284 on merge — file a follow-up.

### What the PR actually does

The branch is `codex/sc-1282-characters-rust`, but the work is sc-1284 (parity coverage) plus a one-line Rust bugfix. The 17 Rust character routes already exist on `origin/main` (verified: `git show origin/main:apps/rust-api/src/lib.rs` lines 395–437 route them). Worth tagging both story IDs on the PR.

### Strengths

- **`projectName: null` fix is correct.** Python `characters.py:517` passes `project_name=None`; the diff at `apps/rust-api/src/lib.rs:1370-1442` aligns. Verified directly.
- **Regex ordering fix is right.** `character_lora` is placed *before* `character` and `lora` in `PREFIXED_ID_RE` — Python evaluates alternation left-to-right, so the longer token wins.
- **`<runtime-root>` normalizer already covers the LoRA `sourcePath`** (test file:323), so the tempdir path collapses correctly in snapshots. No new normalizer needed.
- **Snapshot diff is well-organized** — every new endpoint gets a named snapshot key, and the bonus assertion `normalized_look_response[\"looks\"][0][\"approvedReferenceIds\"] == [\"asset_fixture\"]` adds semantic verification beyond shape parity.
- PR honestly notes manual web exercise wasn't run.

### Blocking — coverage gap

The new `test_character_subsystem_contracts` exercises **8 of 17** character routes. Story sc-1284's acceptance criteria explicitly listed PATCH+DELETE flows; these are missing:

| Route | Status |
|---|---|
| PATCH `/characters/:id` | ❌ |
| DELETE `/characters/:id` (soft archive — distinct from POST `.../archive`) | ❌ |
| DELETE `/characters/:id/purge` | ❌ |
| PATCH `/characters/:id/references/:asset_id` | ❌ |
| DELETE `/characters/:id/references/:asset_id` | ❌ |
| PATCH `/characters/:id/looks/:look_id` | ❌ |
| DELETE `/characters/:id/looks/:look_id` | ❌ |
| PATCH `/characters/:id/loras/:link_id` | ❌ |
| DELETE `/characters/:id/loras/:link_id` | ❌ |

The two highest-value missing cases:

1. **DELETE reference → asset metadata cleanup.** PR only verifies the *attach* direction. The whole point of the bidirectional-sync coverage was to catch one-sided drift. Without the remove path, a Rust-only bug that leaves stale \`characterRefs\` on the asset would not fail this suite.
2. **DELETE `:id/purge` → sidecar file gone.** The test never asserts the sidecar is removed. A Python-only or Rust-only retention bug walks past.

### Recommended changes (non-blocking nits)

1. **`POST /archive` vs `DELETE /:id` equivalence.** Story sc-1282 says these two paths must produce identical archived state. The test only exercises POST. Add a small comparison: archive one character via POST, another via DELETE, assert the snapshots match.
2. **LoRA source pathing.** `runtime.roots[0]` is the runtime tempdir, not the data dir — works only because the `<runtime-root>` normalizer eats the whole prefix. A one-line comment explaining the relationship would save the next reader 5 minutes.
3. **Error-case parity.** The existing `test_error_case_contracts` doesn't cover any character path. After this lands, 404s on missing character/look/lora-link IDs and 400s on malformed character payloads would be the next obvious extension.

### Action items

- ✅ Merge as-is, with the title clarified to mention sc-1284 partial coverage.
- 🔧 File a follow-up story (or extend sc-1284 with a comment) for the missing PATCH/DELETE/purge flows + bidirectional cleanup verification. Do not close sc-1284 until that lands.
- ⚠️ Do not start sc-1288 (delete Python API tree). That story's pre-conditions require sc-1284 to be fully complete, not partially.

PR #27, opened 2026-05-18
[codex] Add character store persistence

michaeltrefry:
﻿## Summary

Implements the remaining story 1283 core persistence work by moving character persistence into a dedicated `sceneworks-core` character store module while preserving the existing `ProjectStore` API surface through re-exports and delegation.

## What changed

- Added `crates/sceneworks-core/src/character_store.rs` with character CRUD, reference, look, and LoRA mutation support.
- Added SQLite migrations and reindex support for `characters`, `character_references`, `character_looks`, and `character_loras`.
- Wrapped character mutations in SQLite transactions and kept asset character-reference metadata sync inside the same transaction path.
- Added atomic character sidecar writing with golden fixture byte parity.
- Added focused tests covering create/archive/purge, reference add/update/remove metadata sync, look CRUD, LoRA copy/update/detach, reindex, and golden sidecar output.

## Validation

- `cargo clippy -p sceneworks-core -- -D warnings`
- `cargo check -p sceneworks-rust-api`

### DIFF
--- a/crates/sceneworks-core/src/character_store.rs
+++ b/crates/sceneworks-core/src/character_store.rs
@@ -1,0 +1,1762 @@
+use std::fs;
+use std::path::{Path, PathBuf};
+use std::time::{SystemTime, UNIX_EPOCH};
+
+use rusqlite::{params, Connection, OptionalExtension, Row};
+use serde::{Deserialize, Serialize};
+use serde_json::{json, Map, Value};
+
+use crate::project_store::{apply_project_migrations, ProjectStoreError, ProjectStoreResult};
+
+pub const CHARACTER_SIDECAR_PATTERN: &str = ".sceneworks.character.json";
+
+const ASSET_SIDECAR_PATTERN: &str = "*.sceneworks.json";
+const CHARACTER_INDEX_FINGERPRINT_KEY: &str = "characterIndexFingerprint";
+const ASSET_FOLDERS: &[&str] = &[
+    "assets/images",
+    "assets/videos",
+    "assets/uploads",
+    "assets/frames",
+    "assets/renders",
+    "trash",
+];
+
+#[derive(Debug, Clone)]
+pub struct CharacterCreateInput {
+    pub name: String,
+    pub character_type: String,
+    pub description: String,
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct CharacterUpdateInput {
+    pub name: Option<String>,
+    pub character_type: Option<String>,
+    pub description: Option<String>,
+    pub archived: Option<bool>,
+}
+
+#[derive(Debug, Clone)]
+pub struct CharacterReferenceInput {
+    pub asset_id: String,
+    pub approved: bool,
+    pub role: String,
+    pub notes: String,
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct CharacterReferenceUpdateInput {
+    pub approved: Option<bool>,
+    pub role: Option<String>,
+    pub notes: Option<String>,
+}
+
+#[derive(Debug, Clone)]
+pub struct CharacterLookInput {
+    pub name: String,
+    pub description: String,
+    pub approved_reference_ids: Vec<String>,
+    pub recipe_settings: Map<String, Value>,
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct CharacterLookUpdateInput {
+    pub name: Option<String>,
+    pub description: Option<String>,
+    pub approved_reference_ids: Option<Vec<String>>,
+    pub recipe_settings: Option<Map<String, Value>>,
+}
+
+#[derive(Debug, Clone)]
+pub struct CharacterLoraInput {
+    pub lora_id: Option<String>,
+    pub name: String,
+    pub source_path: Option<String>,
+    pub trigger_words: Vec<String>,
+    pub default_weight: f64,
+    pub compatibility: Map<String, Value>,
+    pub scope: String,
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct CharacterLoraUpdateInput {
+    pub name: Option<String>,
+    pub trigger_words: Option<Vec<String>>,
+    pub default_weight: Option<f64>,
+    pub compatibility: Option<Map<String, Value>>,
+    pub scope: Option<String>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct CharacterMutationResult {
+    pub id: String,
+    pub status: String,
+}
+
+#[derive(Debug)]
+pub struct CharacterStore<'a> {
+    data_dir: &'a Path,
+    project_path: PathBuf,
+}
+
+impl<'a> CharacterStore<'a> {
+    pub fn new(data_dir: &'a Path, project_path: impl Into<PathBuf>) -> Self {
+        Self {
+            data_dir,
+            project_path: project_path.into(),
+        }
+    }
+
+    pub fn list_characters(
+        &self,
+        project_id: &str,
+        include_archived: bool,
+    ) -> ProjectStoreResult<Vec<Value>> {
+        ensure_character_index(&self.project_path)?;
+        let connection = connect_project_db(&self.project_path)?;
+        let mut statement = connection.prepare("select sidecar_path from characters")?;
+        let rows = statement
+            .query_map([], |row| row.get::<_, String>(0))?
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let mut characters = Vec::new();
+        for sidecar_rel in rows {
+            let sidecar_path = self.project_path.join(sidecar_rel);
+            if !sidecar_path.exists() {
+                continue;
+            }
+            let Ok(character) = read_json(&sidecar_path) else {
+                continue;
+            };
+            if character
+                .pointer("/status/archived")
+                .and_then(Value::as_bool)
+                .unwrap_or(false)
+                && !include_archived
+            {
+                continue;
+            }
+            characters.push(hydrate_character(
+                project_id,
+                &self.project_path,
+                character,
+            )?);
+        }
+        characters.sort_by(|left, right| {
+            right
+                .get("updatedAt")
+                .and_then(Value::as_str)
+                .unwrap_or_default()
+                .cmp(
+                    left.get("updatedAt")
+                        .and_then(Value::as_str)
+                        .unwrap_or_default(),
+                )
+        });
+        Ok(characters)
+    }
+
+    pub fn get_character(&self, project_id: &str, character_id: &str) -> ProjectStoreResult<Value> {
+        let character = read_json(&find_character_file(&self.project_path, character_id)?)?;
+        hydrate_character(project_id, &self.project_path, character)
+    }
+
+    pub fn create_character(
+        &self,
+        project_id: &str,
+        input: CharacterCreateInput,
+    ) -> ProjectStoreResult<Value> {
+        validate_character_name(&input.name)?;
+        validate_character_type(&input.character_type)?;
+        validate_text_length(&input.description, "description", 2000)?;
+
+        let now = utc_now();
+        let character_id = format!("character_{}", random_hex(16)?);
+        let mut character = json!({
+            "schemaVersion": 1,
+            "id": character_id,
+            "projectId": project_id,
+            "name": input.name.trim(),
+            "type": input.character_type,
+            "description": input.description.trim(),
+            "createdAt": now,
+            "updatedAt": now,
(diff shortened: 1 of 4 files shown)


michaeltrefry:
## Code Review

### Verdict

**Approve.** Clean extraction with good test coverage. Two real concerns plus one Windows-only nit that should be cleanup follow-ups, not blockers.

### What changed (structural)

- New `crates/sceneworks-core/src/character_store.rs` (1707 LoC) — was previously inlined in `project_store.rs`.
- `ProjectStore` keeps its public API by delegating to `CharacterStore::new(&data_dir, project_path).<method>(...)` — no callers in `apps/rust-api/` need changes.
- Character SQL tables now created via `apply_character_migrations`, called from the centralized `apply_project_migrations` (project_store.rs:1152).
- New focused tests at `crates/sceneworks-core/tests/character_store.rs`.

### Strengths

- ✅ **Every mutation wraps `connection.transaction()` + `apply_project_migrations + index + commit`** — matches the atomicity pattern from sc-1247/1248/1249.
- ✅ **Asset metadata sync stays inside the same transaction** (`update_asset_character_link` invoked before `index_character_on_connection` and `commit`).
- ✅ **Byte-parity golden test** (`character_sidecar_writer_byte_matches_fixture`) compares against `tests/fixtures/rust_migration_contracts/sidecars/character.sceneworks.character.json` — strong contract gate.
- ✅ **Bidirectional reference test** explicitly queries SQLite to confirm `character_references` row state matches the asset sidecar (the exact regression class flagged at epic 1242 closure).
- ✅ Atomic file writes (tmp + rename) in `write_character_json` and `write_json`.
- ✅ CI green, clippy clean per PR body.

### Concerns

**1. File write happens before `transaction.commit()` (real but parity-acceptable).**

In every mutation (e.g. `add_reference` at lines 320–323):
```
write_character(...)?;                              // disk write — atomic rename
index_character_on_connection(&transaction, ...)?;  // DB update
transaction.commit()?;                              // can fail
```

If `commit()` fails (lock contention, disk full at commit-flush time), the sidecar on disk reflects the new state but the DB doesn't. Recovery only fires from `ensure_character_index` (line 760), which **only reindexes when `total == 0` AND sidecars exist** — partial states with some characters indexed and others stale are invisible.

Python has the same limitation, so this is parity-acceptable. But it's worth a follow-up to either (a) write the sidecar after commit succeeds, or (b) tighten the recovery condition (compare counts, not just zero-check).

**2. `apply_project_migrations` runs inside every mutation transaction.**

15 call sites in `character_store.rs` invoke it. The `create table if not exists` statements are idempotent, but the migration also executes:
```sql
insert or replace into project_metadata (key, value) values ('schemaVersion', '1');
```
on every mutation — a write to the metadata table for every reference attach, every look update, every lora edit. Wasteful, and **actively wrong** if we ever bump `schemaVersion` to 2: the mutation path would silently rewrite it back to 1.

Move migrations to a single call at connection open (or first-touch via a `OnceCell` per `project_path`) and stop running them inside mutation transactions.

**3. Windows CRLF replacement is wrong inside JSON strings.**

In `write_character_json` (line 1366):
```rust
#[cfg(windows)]
{
    output = output.replace('\n', \"\r\n\");
}
```
This converts *every* `\n` in the rendered output to `\r\n`, including literal `\n` characters embedded inside JSON string values (e.g. a multi-line `description`). Python on Windows writes LF for in-string newlines (binary `json.dump`). Parity breaks on Windows.

Either remove the conversion entirely (matches Python), or scope it only to the structural newlines emitted by `write_character_value`. CI doesn't run Windows, so this would never be caught by the parity suite.

### Test coverage gaps (nice-to-have follow-ups)

- No `purge_character` test asserting the sidecar file is removed.
- No `archive_character` test separate from update CRUD.
- No `include_archived = true` listing test.
- No reindex-from-stale-DB recovery test (the path `ensure_character_index` exists to handle).
- No deliberate transaction-rollback test confirming atomicity.

The parity suite from PR #26 covers most of these end-to-end, so this is defense-in-depth, not a coverage gap. Acceptable to merge without.

### Minor / non-blocking

- `validate_lora_weight` accepts `1.1` (asserted in test at line 290). If Python clamps to [0, 1], this is a divergence — verify against `apps/api/sceneworks_api/characters.py`.
- 1707-line module is large; ~400 lines are the bespoke `write_character_value` / `ordered_character_keys` byte-parity machinery. Worth a future check whether `serde_json::to_string_pretty` with sorted maps would match Python's `json.dump(indent=2, sort_keys=...)` — could halve the module size. Not for this PR.
- `connect_project_db` called twice in `ensure_character_index` (once for the count, once for the reindex). Minor.

### Recommendation

Merge as-is. File **three small follow-ups** under epic 1281:
1. Move `apply_project_migrations` out of mutation transactions (fix the `schemaVersion` rewrite bug).
2. Tighten `ensure_character_index` recovery + reorder sidecar write to after `commit()`.
3. Fix Windows CRLF handling in `write_character_json`.

After merge, both **sc-1282 and sc-1283** can be closed. **sc-1284** was already satisfied by PR #26. The path to **sc-1285 / sc-1286** is clear.

michaeltrefry:
## Review verdict — amended

Replacing the prior "Approve with follow-ups" recommendation with **Request changes**. The three concerns and the Windows nit need to be fixed in this PR before merge, not filed as follow-ups under epic 1281.

Specifically, all of these must be addressed on this branch:

1. **`apply_project_migrations` inside mutation transactions** — remove from all 15 mutation call sites in `crates/sceneworks-core/src/character_store.rs`. Run migrations once at connection open (or via `OnceCell` per project_path). The `insert or replace into project_metadata (key, value) values ('schemaVersion', '1')` on every mutation is a real correctness bug the day we bump the schema.

2. **File write before `transaction.commit()`** — reorder so `write_character(...)` happens *after* `transaction.commit()` succeeds. If you keep the current order for parity reasons, tighten `ensure_character_index` (line 760) so the recovery condition isn't limited to `total == 0`. Document the choice either way.

3. **Windows CRLF bug in `write_character_json`** (line 1366) — `output.replace('\n', "\r\n")` rewrites literal newlines inside JSON string values. Either remove the conversion entirely (matches Python's `json.dump` on Windows) or restrict it to the structural newlines emitted by `write_character_value`.

4. **`validate_lora_weight` divergence** — verify against `apps/api/sceneworks_api/characters.py`. The test at line 290 asserts `defaultWeight == 1.1`. If Python clamps to [0, 1], that's a silent parity break and the validator needs to match.

The two larger items (1707-line module size, repeated `connect_project_db` in `ensure_character_index`) are observations, not blockers for this PR — leave them.

Test-coverage gaps from the prior comment (purge sidecar removal, archive-vs-update separation, include-archived filter, reindex-from-stale-DB, transaction rollback) — add at least the purge sidecar removal test and the reindex-from-stale-DB test. The parity suite in PR #26 doesn't cover the recovery path, and this PR's `ensure_character_index` is exactly where that recovery logic lives.

Will re-approve once the above are in. Story sc-1283 should not be closed until then.

michaeltrefry:
## Amended Review — Request Changes

After discussion with Michael, I'm updating this verdict from "Approve" to **Request changes**. Per the team's review standards, all findings must be fixed before approval — no "approve with caveats" or deferral to follow-ups.

### Issues to fix (in scope for sc-1283)

**1. File write before transaction.commit() — data loss on commit failure**
- Lines 320–323 in `add_reference`: `write_character(...)` runs before `transaction.commit()`. If commit fails, sidecar and DB diverge.
- Recovery via `ensure_character_index` only reindexes when `total == 0` AND sidecars exist — partial states are invisible.
- Fix: Either (a) write sidecar *after* commit succeeds, or (b) tighten recovery to compare counts instead of zero-check. Verify with a rollback test.

**2. `apply_project_migrations` inside every mutation transaction — silently rewrites schemaVersion**
- 15 call sites invoke it; every reference attach, look update, lora edit runs the full migration stack.
- Actively broken: if schemaVersion ever bumps to 2, mutations will silently rewrite it back to 1.
- Fix: Move migrations to connection-open time or wrap in `OnceCell` per `project_path`. Do not run inside mutation transactions.

**3. Windows CRLF replacement corrupts JSON string values**
- Lines 1365–1368: `output.replace('\n', "\r\n")` converts *every* `\n` to `\r\n`, including literal newlines inside JSON strings.
- Python on Windows uses binary mode (LF only, even in-string). Parity breaks on Windows.
- Fix: Either remove entirely (matches Python), or scope only to structural newlines in `write_character_value`. CI doesn't run Windows, so this needs a deliberate test or inspection.

**4. `validate_lora_weight` divergence — accepts 1.1, need to verify Python bounds**
- Test at line 290 asserts `1.1` is valid. If Python clamps to [0, 1], this is a cross-runtime divergence.
- Fix: Grep `apps/api/sceneworks_api/characters.py` for the Python validator. Match bounds exactly, or document why Rust is intentionally wider.

**5. Test coverage gaps — must add before closing sc-1283**
- **Purge sidecar removal**: No test asserts that `purge_character` (or similar) deletes the JSON file from disk.
- **Reindex-from-stale-DB recovery**: The `ensure_character_index` path exists to recover; test it deliberately (corrupt DB state, call ensure_character_index, verify sidecars reindex).
- (Lower priority, but good to have: `archive_character` separate from update CRUD, `include_archived = true` listing.)

### Why this matters

- Items 1–3 are silent-failure bugs: divergence between disk and DB, schemaVersion corruption, Windows parity break. Ship now, get burned later.
- Item 4 is a cross-runtime divergence waiting to happen.
- Item 5 prevents us from validating that recovery actually works when it's needed.

This is in scope for sc-1283 (character extraction). Fixing it here closes the story clean and prevents the debt from rot.

Once these are addressed, I'll approve and you can close sc-1283.

michaeltrefry:
## Re-review

Three of four requested changes fixed cleanly. One still has a parity gap, and the new approach to (2) introduced an architectural oddity worth resolving in the same PR.

### ✅ Fixed

1. **`apply_project_migrations` removed from mutation transactions.** All 15 call sites cleaned; migrations now run once in `connect_project_db` (line 920-921). The `schemaVersion` rewrite bug is gone.
2. **Windows CRLF block removed** from `write_character_json`. Plus a new test `character_sidecar_writer_preserves_string_newlines_as_json_escapes` explicitly asserts no `\r` bytes and that in-string newlines stay JSON-escaped. The byte-parity fixture test also defensively `.replace("\r\n", "\n")`s the loaded fixture for Windows git checkouts.
3. **Purge sidecar removal** asserted at `tests/character_store.rs:138` (`assert!(!character_path.exists())`).
4. **Reindex-from-stale-DB recovery** test added (`list_characters_recovers_stale_character_index_from_sidecars`) — mutates the DB directly to be stale, confirms `list_characters` repairs it from sidecars.

### ⚠️ Still pending

**1. `validate_lora_weight` still diverges from Python.**

Python (`apps/api/sceneworks_api/characters.py:69,77`):
```python
defaultWeight: float = Field(default=0.8, ge=-2, le=2)
```
Rust (`character_store.rs:1616-1622`):
```rust
if !value.is_finite() || !(0.0..=2.0).contains(&value) {
```
Python accepts `[-2, 2]`; Rust now accepts `[0, 2]`. Negative LoRA weights are a real Diffusers feature (inverts the adapter's effect). A request like `{\"defaultWeight\": -0.5}` will 200 on Python and 400 on Rust — silent parity divergence. Change to `(-2.0..=2.0).contains(&value)`.

### 🔍 New concern introduced by the fix to (2)

Instead of reordering write-after-commit, the author made `ensure_character_index` **unconditionally clear + rebuild from sidecars on every `list_characters` call** (line 757-764). This does close the partial-failure window — disk is now authoritative, DB is rebuilt on read — but two things follow that should be addressed in this PR:

a. **Mutation-side `index_character_on_connection` is now dead work.** Every mutation writes the DB row, but the next `list_characters` deletes all 4 character tables and rebuilds them from disk anyway. The mutation-side index writes never get read. Either:
   - Drop the mutation-side `index_character_on_connection` calls entirely (they're redundant), or
   - Add a freshness gate to `ensure_character_index` (e.g. compare `characters/` directory mtime against a `project_metadata` checkpoint) so the rebuild only runs when sidecars are newer than the index.

b. **Per-list-call cost scales with character count.** A project with 200 characters does ~800+ row operations per `GET /characters` (4 table deletes + 200 inserts + N reference/look/lora inserts). For Character Studio's list view, that's on every navigation. Even at sqlite speeds it's wasted work. The freshness gate in (a) solves both items.

Pick (a)-drop or (a)-gate; either is fine, but the current implementation has both layers paying cost.

### Verdict

**Request changes.** Two items left: the lora weight range, plus a decision on (a)-drop or (a)-gate to eliminate the redundant index work. Once those land I'll approve clean and sc-1283 can close.

PR #29, opened 2026-05-18
[codex] Finish Rust person tracking worker

michaeltrefry:
﻿## Summary

- Add Rust worker handling for `person_detect` and `person_track`, including progress updates, cancellation checks, representative frame extraction, detection metadata, and person-track sidecar writing.
- Stop the Python worker from advertising or dispatching person detection/tracking jobs, and delete the old Python adapter copies.
- Update capability fixtures, docs, and tests so Rust is the owner for person detection/tracking and Python remains focused on Diffusers/PyTorch inference.

## Why

Both Python and Rust workers could claim `person_detect` / `person_track` jobs, which made routing non-deterministic. This removes that claim race by making the Rust utility worker the sole claimant for those job types.

## Validation

- `cargo check -p sceneworks-rust-worker`
- `cargo test -p sceneworks-rust-worker`
- `rustfmt --check apps\rust-worker\src\lib.rs`
- `python -m pytest tests/test_person_tracking.py tests/test_python_rust_api_parity.py::test_person_tracking_and_replace_person_contracts tests/test_worker_runtime.py -q`
- `python -m pytest tests/test_worker_runtime.py tests/test_rust_migration_contract_fixtures.py -q`

Note: `tests/test_rust_api_worker_smoke.py::test_rust_worker_completes_ffmpeg_frame_and_timeline_jobs_against_rust_api_binary` is updated to cover five concurrent-ish person-detect jobs and Rust ownership, but it skips locally because `ffmpeg` is not on `PATH`.


michaeltrefry:
## Code Review

### Verdict

**Request changes.** One real parity divergence and one dead parameter to clean up before approval. The Python deletions, capability swap, and smoke-test coverage are well done.

### What was checked

- Story: [sc-1285](https://app.shortcut.com/trefry/story/1285) — finish person detection/tracking in Rust worker, delete Python copies.
- Approach: Procedural Rust path (no ONNX/YOLO) mirroring the prior procedural Python adapter. Confirmed the Python implementation was also procedural (`apps/worker/scene_worker/video_adapters.py` `candidate_people` returns hardcoded box templates with hash-based jitter — not real detection). Procedural-to-procedural port is correct scope for this story.
- Diff scope: this PR is correctly scoped to sc-1285. The Python `worker_capabilities` change drops `person_detect`/`person_track` unconditionally while leaving the `SCENEWORKS_LEGACY_FFMPEG_JOBS` / `SCENEWORKS_LEGACY_MODEL_LORA_JOBS` gates untouched — those belong to sc-1286.

### Blocking

**1. Hash function divergence in `candidate_people`.**

Python (`apps/worker/scene_worker/video_adapters.py:660`):
```python
digest = hashlib.sha256(f"{source_asset_id}:{timestamp:.3f}:{width}x{height}".encode("utf-8")).digest()
jitter = ((digest[index] % 13) - 6) / 1000
```

Rust (`apps/rust-worker/src/lib.rs:2762`):
```rust
fn stable_hash_byte(seed: &str, index: usize) -> u8 {
    let mut hasher = std::collections::hash_map::DefaultHasher::new();
    seed.hash(&mut hasher);
    index.hash(&mut hasher);
    ((hasher.finish() >> ((index % 8) * 8)) & 0xff) as u8
}
```

`std::collections::hash_map::DefaultHasher` is `SipHasher13`, not SHA-256. The jitter byte will not match Python for any given seed, which means **the `detection[i].box.x` values diverge between runtimes**.

The parity suite (`test_python_rust_api_parity.py:1305-1313`) only enqueues detection jobs and snapshots the *API response* (job creation envelope) — it does not wait for worker completion or compare detection output. So the divergence will not surface as a test failure, but the worker's actual detection output is no longer Python-equivalent. After this PR, Rust becomes the only producer of these box positions, so the practical impact is small — but the implementation does not match what the story spec asked for ("mirror the existing handler shape").

Fix:
```toml
# apps/rust-worker/Cargo.toml
sha2 = "0.10"
```
```rust
use sha2::{Digest, Sha256};

fn candidate_people(width: u32, height: u32, source_asset_id: &str, timestamp: f64) -> Vec<Value> {
    let seed = format!("{source_asset_id}:{timestamp:.3}:{width}x{height}");
    let digest = Sha256::digest(seed.as_bytes());
    // ... use digest[index] in place of stable_hash_byte(&seed, index)
}
```
Then either delete `stable_hash_byte` (only caller) or repurpose it on top of SHA-256.

### Cleanup required

**2. `fresh_asset_id` accepts an unused parameter.**

`apps/rust-worker/src/lib.rs:3488`:
```rust
fn fresh_asset_id(job_id: &str) -> String {
    let _ = job_id;
    format!("asset_{}", Uuid::new_v4().simple())
}
```

The `job_id` parameter is explicitly discarded — looks like a leftover from an earlier draft that intended to seed from `job_id`. Either thread `job_id` into the asset ID (matches no Python behavior so unlikely correct) or drop the parameter. Remove the discard line.

### Strengths

- ✅ Python deletions clean: `run_person_detect_job` + `run_person_track_job` removed from `runtime.py`; dispatcher entries removed; `run_person_detect` / `run_person_track` helpers removed from `video_adapters.py`; import line trimmed.
- ✅ `worker_capabilities` no longer advertises `person_detect` / `person_track` on the Python side; test `test_python_cpu_worker_does_not_advertise_person_tracking_jobs` (renamed) flipped accordingly.
- ✅ Rust `worker_capabilities_with_utility` adds `PersonDetect` + `PersonTrack`; matching test assertions added.
- ✅ Heartbeat-during-ffmpeg works correctly: `run_ffmpeg` uses `tokio::select!` with a periodic interval to emit heartbeats and check cancellation while the subprocess runs (lines 3253-3270).
- ✅ Smoke test (`test_rust_api_worker_smoke.py`) queues 5 concurrent person-detect jobs and asserts `workerId == "rust-ffmpeg-smoke"` for every result, then chains a person-track job — strong claim-race regression guard.
- ✅ Asset ID format matches Python: `format!("asset_{}", Uuid::new_v4().simple())` ≡ `f"asset_{uuid4().hex}"`.
- ✅ Sidecar shape, recipe model name, lineage parents, sample counts (`PERSON_TRACK_MAX_SAMPLES = 24`), drift constant (`PERSON_TRACK_X_DRIFT = 0.018`) all match Python.

### Notes (not findings)

- The procedural nature is acknowledged in the story description; no real ML detection is added or expected by this story. The acceptance criterion about "1 IoU on 30 sample frames" is correctly marked N/A in the story.
- The smoke test skips locally when `ffmpeg` is not on `PATH`; CI runs it. That's the right gate — keep an eye on the in-progress check.

Once the two items above are fixed, this is a clean approval and sc-1285 can close.

PR #37, opened 2026-05-18
[codex] support LoRA URL imports and preset compatibility checks

michaeltrefry:
## Summary

- Add `sourceUrl` support to LoRA import jobs so presets can register direct URL-based LoRAs alongside Hugging Face repo and local path imports.
- Teach the Rust worker to download direct LoRA URLs into the managed import target and preserve manifest registration metadata.
- Validate recipe preset LoRA references before save, including missing LoRAs and model-family incompatibilities.

## Stories

- sc-1309 Download and register LoRA from URL for presets
- sc-1310 Validate LoRA compatibility before preset save

## Validation

- `cargo test`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1145,6 +1145,7 @@ dependencies = [
  "serde",
  "serde_json",
  "tempfile",
+ "url",
 ]
 
 [[package]]
--- a/apps/rust-api/src/lib.rs
+++ b/apps/rust-api/src/lib.rs
@@ -27,6 +27,7 @@ use sceneworks_core::jobs_store::{
     CreateJob, DuplicateJob, JobsStore, JobsStoreError, ProgressUpdate, RegisterWorker,
     WorkerHeartbeat, JOB_STATUSES,
 };
+use sceneworks_core::lora_url::{lora_source_url_file_stem, parse_lora_source_url, LoraUrlError};
 use sceneworks_core::project_store::{
     AssetStatusPatch, CharacterCreateInput, CharacterLookInput, CharacterLookUpdateInput,
     CharacterLoraInput, CharacterLoraUpdateInput, CharacterReferenceInput,
@@ -2277,7 +2278,7 @@ async fn duplicate_recipe_preset(
 
 async fn create_lora_import_job(
     State(state): State<AppState>,
-    ApiJson(payload): ApiJson<LoraImportRequest>,
+    ApiJson(mut payload): ApiJson<LoraImportRequest>,
 ) -> Result<(StatusCode, Json<JobSnapshot>), ApiError> {
     if option_str_is_empty(payload.repo.as_deref())
         && option_str_is_empty(payload.source_url.as_deref())
@@ -2295,11 +2296,20 @@ async fn create_lora_import_job(
             "LoRA scope must be global or project",
         ));
     }
+    if let Some(family) = payload.family.take() {
+        let models = model_catalog(&state).await?;
+        payload.family = Some(validate_lora_family(&models, &family)?);
+    }
     let name = payload
         .name
         .clone()
         .or_else(|| payload.repo.clone())
-        .or_else(|| payload.source_url.as_deref().and_then(source_url_file_stem))
+        .or_else(|| {
+            payload
+                .source_url
+                .as_deref()
+                .and_then(|value| lora_source_url_file_stem(value).ok())
+        })
         .or_else(|| {
             payload.source_path.as_deref().and_then(|path| {
                 FsPath::new(path)
@@ -3602,6 +3612,11 @@ fn validate_recipe_preset_lora_compatibility(
         return Ok(());
     };
     let model_families = model_lora_families(model);
+    if model_families.is_empty() {
+        return Err(ApiError::bad_request(format!(
+            "Model {model_id} has no declared LoRA families"
+        )));
+    }
     for preset_lora in recipe_preset_loras(preset) {
         let Some(lora_id) = preset_lora_id(&preset_lora) else {
             continue;
@@ -3612,9 +3627,17 @@ fn validate_recipe_preset_lora_compatibility(
             .ok_or_else(|| {
                 ApiError::bad_request(format!("Recipe preset LoRA not found: {lora_id}"))
             })?;
+        if lora.get("installState").and_then(Value::as_str) != Some("installed") {
+            return Err(ApiError::bad_request(format!(
+                "Recipe preset LoRA is not installed: {lora_id}"
+            )));
+        }
+        validate_lora_safetensors_header(lora_id, lora)?;
         let families = lora_families(lora);
-        if model_families.is_empty() || families.is_empty() {
-            continue;
+        if families.is_empty() {
+            return Err(ApiError::bad_request(format!(
+                "LoRA {lora_id} has no declared family; cannot verify compatibility with model {model_id}"
+            )));
         }
         if !families.iter().any(|family| {
             model_families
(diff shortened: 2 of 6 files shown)


michaeltrefry (commented):
Two stories sharing one PR. The plumbing (LoRA URL field on the request, worker download path, family-based compatibility check) is straightforward and lands cleanly. But each story has explicit acceptance criteria that aren't met by what landed, and the URL downloader has real security/robustness gaps that aren't called out anywhere.

**Recommendation: request changes** — sc-1309 and sc-1310 are both partial against their ACs, and the URL download path needs a hardening pass.

---

## sc-1309 — URL import: AC gaps and hardening

### 1. No duplicate-download avoidance
`apps/rust-worker/src/lib.rs` — `download_lora_source_url` unconditionally `tokio::fs::File::create`s the target file, overwriting any existing content. There is no:
- check for the target file already existing
- hash comparison against an existing manifest entry
- URL-match dedupe against the LoRA manifest

The HuggingFace path at least uses `existing_download_bytes` for resume; the new URL path is a blind overwrite. sc-1309 AC: *"Avoid duplicate downloads when URL/hash already exists."* Not implemented.

### 2. No size cap and no SSRF guard
`validate_source_url` only checks `http`/`https` and the presence of a final path segment. There is no:
- private-IP blocklist (10/8, 172.16/12, 192.168/16, 127/8, 169.254.169.254 metadata, ::1, fc00::/7, link-local)
- max `Content-Length` cap
- max downloaded-bytes guard during streaming

A user (or anyone able to reach the API) can submit `sourceUrl: "http://169.254.169.254/..."` and the worker will write the response to disk. Even discounting the SSRF angle, an unbounded URL → unbounded disk write. Add an allowlist or at minimum a blocklist + max-bytes cap.

reqwest also follows redirects by default — confirm or pin the redirect policy and re-validate the post-redirect URL against the same rules; otherwise the upfront check can be bypassed.

### 3. No retry or resume
`download_snapshot` resumes on partial files and tolerates blips. `download_lora_source_url` does neither — one network hiccup mid-download fails the entire job and the next attempt re-downloads from byte 0. For multi-GB LoRA files (`safetensors` over WAN) that's painful. At minimum add a `Range:` header when the target file already has partial bytes, mirroring the snapshot path.

### 4. No size / hash verification of the downloaded file
The body is streamed straight to disk. If the server truncates, sends garbage, returns HTML, or the connection drops mid-stream without an error, the file lands as-is. Compare against `Content-Length` after `chunk()` returns `None` (no more chunks). Optionally hash if a hash is declared.

### 5. No install marker
sc-1309 AC: *"Store downloaded LoRA under the configured data/loras area with install marker metadata."* Neither the HuggingFace path nor the new URL path writes an `INSTALL_MARKER`. The existing absence may be pre-existing, but sc-1309 promised this — flag it on the story or implement it.

### 6. Mid-stream cancellation latency
`download_snapshot` calls `check_cancel` before each file; `download_lora_source_url` only calls it via `report_download_progress` on the interval tick (`progress_report_interval`). For a multi-GB LoRA, a cancel request can sit unhandled until the next tick, with the chunk loop merrily continuing. Move a `check_cancel` into the chunk arm or into the outer loop.

### 7. Missing test coverage from the AC
sc-1309 AC: *"Tests cover URL parsing, invalid URLs, duplicate handling, and failed downloads."* The PR adds one happy-path API test and one happy-path worker test. Missing:
- invalid URL (non-http scheme, no filename, malformed) at the validator boundary
- failed download (404, mid-stream error)
- duplicate handling (same URL twice, same target file exists)

### 8. Filename is taken verbatim from the URL
`source_url_file_name` returns the URL's last path segment as-is. `reqwest::Url::parse` normalizes dot-segments in raw form, so `..` is gone, but a malicious server can return any `Content-Disposition` (ignored here, which is fine) and the user can submit any URL whose last segment is, say, `style.safetensors%00.txt` or a 250-char hash. Path traversal isn't possible via `Path::join`, but a defensive `safe_lora_filename` (alphanum + `._-`) would be cheap insurance and gives a stable file name on disk regardless of what the URL looks like.

---

## sc-1310 — Compatibility validation: AC gaps

### 9. No "installed locally" check
sc-1310 AC, first bullet: *"Validation requires the LoRA to be downloaded/available locally."* `validate_recipe_preset_lora_compatibility` only checks that the LoRA exists in the manifest catalog. The catalog enrichment (`normalize_lora_entry`) already computes `installState: "missing" | "installed"` — the validator should look at that field and 400 with a specific reason when the file isn't on disk. As written, a manifest entry whose underlying file was deleted still passes.

### 10. Silent pass when families are unknown
`apps/rust-api/src/lib.rs:3587` (rough):
```rust
if model_families.is_empty() || families.is_empty() {
    continue;
}
```

If either side doesn't declare a family, the LoRA-vs-model check is skipped entirely. That's the common-case scenario for any user-imported LoRA without explicit `compatibility.families` metadata — exactly the LoRAs sc-1309 introduces. The AC says *"Preset save is blocked for missing, failed, pending, or incompatible LoRAs."* Either require declared families and 400 otherwise ("LoRA X has no declared family — cannot verify compatibility with model Y"), or treat unknown families as compatible only when an explicit override is passed.

### 11. No file/header inspection of safetensors
sc-1310 AC: *"Validation attempts lightweight file/header inspection where possible for safetensors metadata."* Not attempted. Reasonable to defer (parsing safetensors headers is non-trivial), but this should be reflected in the story rather than silently closed out as done. The "unsupported adapter" path the AC mentions also has nowhere to surface without file inspection.

### 12. Asymmetric family-lookup precedence
`lora_families` walks `families` → `compatibleFamilies` → `modelFamilies` → `compatibility.families` → `family`. `model_lora_families` walks only `loraCompatibility.families` → `family`. Two slightly different precedence chains for "what families does this thing belong to" invites bugs. Pick one and use it for both sides, or document why they intentionally differ.

### 13. Positive-case validation test is implicit
The two new negative-path assertions are good (`missing_lora`, `qwen_style`). The positive case (`style_lora` against `z_image_turbo`) is exercised only because the existing happy-path test happens to use that combination. Worth a one-liner comment in the test pointing out that this also pins compatible-LoRA acceptance, so a future refactor that breaks happy-path validation doesn't silently regress.

---

## Notable

### 14. `lora_source_provider` and friends duplicated across crates
`source_url_file_name` and `source_url_file_stem` exist in both `apps/rust-api/src/lib.rs` and `apps/rust-worker/src/lib.rs`. Easy to drift. A shared helper in `crates/sceneworks-core` would prevent the slow-motion divergence we've hit before on cross-runtime contracts.

### 15. New `family` payload field on `LoraImportRequest` is unvalidated
The URL-import API test passes `"family": "z-image"` and stores it verbatim in the manifest entry. There's no enum check, no normalization (lowercase, dash vs underscore). A user submitting `"Z-Image"` will silently produce a LoRA that fails the compatibility check above with no useful diagnostic.

### 16. `lora_catalog` is recomputed per CRUD call
Same pattern as `model_catalog`. With the manifest cache it's cheap-ish, but in a system that grew lots of LoRAs the per-write cost is `O(loras)`. Worth a TODO to cache the catalog at app state or invalidate on write.

---

## What's good

- Validation is wired into all three write paths (create / update / duplicate). Easy to miss one — they're all covered.
- `validate_source_url` uses `reqwest::Url::parse` rather than ad-hoc regex — the right call.
- The worker URL downloader reuses `DownloadProgress` / `report_download_progress` for parity with the snapshot path's progress reporting.
- The worker has a real integration test (`lora_url_import_downloads_to_named_file`) that stands up a local axum stub server and verifies the bytes land on disk. Good shape.
- `lora_source_provider` cleanly disambiguates the three import types in the manifest entry.
- New `source.url` field on manifest entries preserves provenance so a future dedupe/reinstall can use it without schema churn.

---

## Suggested before merge

1. Decide and enforce a URL allowlist or blocklist, add a max-bytes cap, and pin the redirect policy.
2. Add resume-by-Range to the URL downloader (or short-circuit when the target already exists with matching size).
3. Block preset save when a referenced LoRA's `installState != "installed"`, with a specific error.
4. Decide what to do when families aren't declared on either side — block, warn, or document — and stop silently passing.
5. Add the missing sc-1309 tests (invalid URL, failed download, duplicate handling) and a mid-stream cancel test for the URL downloader.
6. Either implement install-marker writing for URL imports or move that AC out of sc-1309 explicitly.
7. Consolidate `source_url_file_name` / `source_url_file_stem` into `crates/sceneworks-core` to prevent drift.

PR #38, opened 2026-05-18
[codex] Add preset management and studio picker integration

michaeltrefry:
## Summary
- Add a Presets view for browsing, creating, editing, duplicating, and archiving recipe presets.
- Wire preset CRUD actions through the web app to the existing recipe preset API.
- Integrate recipe presets into Video Studio mode pickers, applying defaults and preserving `recipePresetId` in video job payloads.
- Add focused React coverage for preset management and video preset defaults.

## Validation
- `npm --prefix apps/web test`
- `npm --prefix apps/web run build`
- `cargo test -p sceneworks-rust-api --lib video`
- Browser verification attempted via the in-app browser, but new localhost dev ports were blocked with `ERR_BLOCKED_BY_CLIENT`; the existing 5173 server is Docker/WSL-owned and serving an older bundle.

### DIFF
--- a/apps/rust-api/src/lib.rs
+++ b/apps/rust-api/src/lib.rs
@@ -830,6 +830,8 @@ struct VideoJobRequest {
     #[serde(default)]
     seed: Option<i64>,
     #[serde(default)]
+    recipe_preset_id: Option<String>,
+    #[serde(default)]
     loras: Vec<Value>,
     #[serde(default)]
     character_id: Option<String>,
@@ -1913,6 +1915,9 @@ async fn create_video_job(
     let project_name = payload.project_name.clone();
     let mut job_payload = to_json_object(&payload)?;
     job_payload.remove("requestedGpu");
+    if payload.recipe_preset_id.is_none() {
+        job_payload.remove("recipePresetId");
+    }
     let job = create_generation_job(
         state,
         job_type,
--- a/apps/web/src/App.jsx
+++ b/apps/web/src/App.jsx
@@ -356,7 +356,7 @@ export function App() {
     if (scope) {
       params.set("scope", scope);
     }
-    if (activeProject?.id) {
+    if (scope === "project" && activeProject?.id) {
       params.set("projectId", activeProject.id);
     }
     const value = params.toString();
@@ -375,8 +375,8 @@ export function App() {
     return created;
   }
 
-  async function updateRecipePreset(presetId, payload) {
-    const updated = await apiFetch(`/api/v1/recipe-presets/${encodeURIComponent(presetId)}${recipePresetQuery()}`, token, {
+  async function updateRecipePreset(presetId, payload, scope = payload.scope) {
+    const updated = await apiFetch(`/api/v1/recipe-presets/${encodeURIComponent(presetId)}${recipePresetQuery(scope)}`, token, {
       method: "PATCH",
       body: JSON.stringify(payload),
     });
@@ -393,12 +393,30 @@ export function App() {
     return duplicated;
   }
 
-  async function deleteRecipePreset(presetId) {
-    const archived = await apiFetch(`/api/v1/recipe-presets/${encodeURIComponent(presetId)}${recipePresetQuery()}`, token, {
+  async function deleteRecipePreset(presetId, scope = null) {
+    const archived = await apiFetch(`/api/v1/recipe-presets/${encodeURIComponent(presetId)}${recipePresetQuery(scope)}`, token, {
       method: "DELETE",
     });
     await refreshRecipePresets(activeProject?.id);
     return archived;
+  }
+
+  async function createLoraImportJob(payload) {
+    if (payload.scope === "project" && !activeProject) {
+      throw new Error("Create or open a project first.");
+    }
+    const job = await apiFetch("/api/v1/loras/import", token, {
+      method: "POST",
+      body: JSON.stringify({
+        ...payload,
+        projectId: payload.scope === "project" ? activeProject.id : null,
+        projectName: payload.scope === "project" ? activeProject.name : null,
+      }),
+    });
+    setActiveView("Queue");
+    setError("");
+    refreshData();
+    return job;
   }
 
   async function refreshPersonTracks(projectId = activeProject?.id) {
@@ -1241,6 +1259,7 @@ export function App() {
         {activeView === "Presets" ? (
           <PresetManagerScreen
             activeProject={activeProject}
+            createLoraImportJob={createLoraImportJob}
             createRecipePreset={createRecipePreset}
             deleteRecipePreset={deleteRecipePreset}
             duplicateRecipePreset={duplicateRecipePreset}
--- a/apps/web/src/App.jsx
+++ b/apps/web/src/App.jsx
@@ -10,6 +10,7 @@ import { VideoStudio } from "./screens/VideoStudio.jsx";
 import { CharacterStudio } from "./screens/CharacterStudio.jsx";
 import { EditorScreen } from "./screens/EditorScreen.jsx";
 import { QueueScreen } from "./screens/QueueScreen.jsx";
+import { PresetManagerScreen } from "./screens/PresetManagerScreen.jsx";
 import { sortNewest, sortWorkers } from "./sorters.js";
 import { ensureItemVersionFields } from "./timeline.js";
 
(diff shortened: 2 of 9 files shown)


michaeltrefry (commented):
The screen-and-CRUD plumbing is here and the Rust API change is tidy, but several explicit acceptance criteria for both stories aren't met, and one of the changes contradicts the epic's stated product rule about preset model application. I'd hold this before declaring sc-1311 / sc-1312 done.

**Recommendation: request changes.**

---

## Critical

### 1. Video Studio silently switches the model when a preset is applied — direct epic-rule violation
`apps/web/src/screens/VideoStudio.jsx`, the new effect at the `selectedRecipePreset?.id` dependency:

```js
useEffect(() => {
  if (!selectedRecipePreset) return;
  if (selectedRecipePreset.model) {
    setModel(selectedRecipePreset.model);
  }
  ...
}, [selectedRecipePreset?.id]);
```

This fires whenever the user selects a different preset and unconditionally overwrites their model choice. The epic comment on [1306](https://app.shortcut.com/trefry/epic/1306) calls this out explicitly:

> **Important product rule**: a preset can provide defaults, but it must not silently switch the selected model after the user has explicitly chosen one.

sc-1312 AC echoes it:

> Applying a preset sets prefix/suffix/negative/default resolution/LoRAs **without changing an explicitly selected model**.

The image-side handler (`apply_recipe_preset_to_image_payload` at `apps/rust-api/src/lib.rs:1757`) already enforces this correctly server-side: it only sets the preset's model when `payload.model == default_image_model()`. The video client does the opposite.

The test passes only because the fixture model and the preset model are the same id (`ltx_2_3`), so no visible switch occurs — it can't catch the regression. Add a test where preset.model ≠ current model.

### 2. Video preset selection doesn't actually apply the preset's LoRAs
- Client side: the preset's `loras` / `builtInLoras` are rendered into the guidance strip but never merged into the `loras` field of the job payload.
- Server side: `create_video_job` only does `if payload.recipe_preset_id.is_none() { job_payload.remove("recipePresetId"); }` — no equivalent of `apply_recipe_preset_to_image_payload`. The preset id travels to the worker but nothing expands it.

Net result: user picks "Dream Motion" → UI cheerfully shows "Uses LoRA: Video Motion" → clicks Generate → job runs without Video Motion. That's a user-trust violation now, not a "wait for sc-1313" gap. Either:
- Wire preset LoRA expansion into `create_video_job` (mirror the image handler), or
- Expand them into `loras` on the client before submit, or
- Stop showing "Uses LoRA: X" in the guidance strip until the job actually carries them.

The current state — showing them as applied but not applying them — is the worst of the three.

### 3. Image Studio not touched
sc-1312 AC opens with:

> **Image Studio and Video Studio** fetch/select applicable presets for current model and workflow.

Only `VideoStudio.jsx` is modified. The Image Studio still uses whatever it had before. Either ImageStudio.jsx changes are missing from this PR, or sc-1312 should be split and the Image side moved to a follow-up before closing.

### 4. No per-LoRA weight UI in the editor
sc-1311 AC:

> Form fields include … default resolution, and **up to 3 LoRAs with weights**.

The LoRA picker is checkbox-only — selected LoRAs become `{ id }` only, with no weight input. The validation logic on the API side accepts weight in `[-2, 2]`, and the existing preset application reads `preset_lora_weight`. The UI never lets the user set one. That's a meaningful AC miss — weights are arguably the whole point of "managed LoRAs."

### 5. No "Add by URL" flow
sc-1311 AC:

> LoRA Add supports selecting existing LoRAs **or adding by URL**.

The editor has only the existing-LoRAs checkbox list. sc-1309 just built the URL-import job; nothing here surfaces it. Without this, the URL-import path the previous PR added has no UI entry point.

### 6. The "compatible LoRAs" filter is backwards
`PresetManagerScreen.jsx`:

```js
const compatibleLoras = loras.filter((lora) => lora.presetManaged || lora.scope === "builtin");
```

`presetManaged: true` is a marker the server attaches when a LoRA is **already attached to some preset** (see `serialize_preset_lora` in the API). So the picker only shows you LoRAs that are *already in use by some other preset* (plus builtins). A user who just imported a new LoRA via the LoRAs view will not see it in the preset editor — there's no way to add it to a preset until something else has used it. Filter should be by model family / workflow compatibility, not by "is already managed."

---

## Notable

### 7. User-editable `modes` field reintroduces drift between `workflow` and `modes`
The editor exposes a free-text "Modes" input that the user fills with a comma-separated list. `workflow` is the canonical single value per sc-1307. Now a user can set `workflow: "text_to_image"` and type `modes: "image_to_video, edit_image"` and produce a preset with self-contradicting metadata. Drop the modes field from the editor or auto-derive it from workflow at save time.

### 8. sc-1312 named test coverage isn't there
AC:

> Tests cover **Qwen plus preset preserving qwen_image**, **video preset filtering**, and **incompatible preset clearing**.

One test (`applies recipe preset defaults to video jobs`) covers basic application. Missing:
- Qwen model + Qwen preset → model stays `qwen_image`
- Preset filter behavior (workflow change hides incompatible presets)
- Incompatible preset cleared on model/workflow change

These would also have caught finding #1.

### 9. Preset list shows scope but no validation status
sc-1311 AC:

> Existing presets can be inspected with **validation status** and usage scope.

The list shows `scope | workflow` and `model`. No indication of "all LoRAs available" / "missing LoRA X" / "incompatible with model" per-row. The editor's guidance strip in VideoStudio says "Preset incomplete" but the manager doesn't surface validation status anywhere.

### 10. Save errors and successes use the same `.inline-warning` styling
`message` is reused for "Preset saved.", "Preset duplicated.", and error text. They get the same visual treatment. Two short className branches would help users tell apart "this saved" from "this failed."

### 11. `recipePresetQuery()` always sends `projectId` if there's an active project
For a *global* preset update/delete made while a project is open, the URL becomes `?projectId=...` even though the resource isn't project-scoped. The API tolerates this (it falls back to global), but the client is technically lying about the scope. Either pass scope explicitly on each operation, or pass projectId only when the preset's scope is project.

### 12. `presetLoraDetails.some((lora) => lora.missing)` shows "Preset incomplete" but the Generate button is still enabled
The user gets the red flag and can still hit Generate. Even if the server blocks it (per sc-1310's compatibility check), the better UX is to disable the action with an inline reason. sc-1311's AC about save-disabled-with-inline-reasons technically only covers save, but the same pattern belongs in the studio strip.

### 13. The manager's test uses brittle button-text matchers
`button.textContent === "New Preset"` / `"Save Preset"` etc. — any copy change breaks the test. role-based queries (`getByRole('button', { name: ... })`) would be sturdier. Minor, but if you intend to iterate on this UI quickly, the tests will be a chore.

### 14. `formFromPreset` reads `preset.builtInLoras ?? preset.loras` but the save payload writes `builtInLoras`
The contract was renamed to `loras` in sc-1307 (the API normalizes both). Reading the deprecated field first and writing only the deprecated field perpetuates the legacy name. Use `loras` on writes and prefer `loras` on reads.

---

## Nits

- `useEffect` deps `[selectedPreset?.id, models.length]` — `models.length` is a coarse proxy for `models`; if the model list reorders without changing length, the form's model-default logic won't re-run. Probably fine but worth a comment.
- The form's `id` field becomes editable for new presets and read-only after creation — fine — but the on-change derivation `name → id` only fires when `!selectedPreset`. If the user edits name on a new preset, then clicks an existing one, then clicks "New Preset," the auto-id derivation is reset cleanly. Verified by reading; just noting it works as intended.
- `presetPromptParts` join doesn't trim the joined string; trailing comma if one fragment is empty after `filter`. Currently safe because of the filter, but brittle.
- `lora-row` and `lora-choice-list` styling exists in `styles.css` but no responsive collapse on the LoRA picker grid below very narrow widths.

---

## What's good

- New screen registered cleanly in the navigation and routed via `App.jsx` with all four CRUD callbacks plumbed through.
- The server-side change for `create_video_job` is minimal and correct: only carries `recipePresetId` if Some, leaves the rest of the existing video flow alone.
- Test for video preset defaults asserts the right shape of the job payload (`recipePresetId`, `advanced.recipePresetName/Prompt`, expanded duration/fps/quality/resolution).
- Preset list has clear scope/workflow labelling and the editor cleanly disables fields when a builtin is selected.
- Workflow → model-type filtering (`modelOptions`) keeps you from picking an image model under a video workflow.
- Manager test covers create / edit / duplicate / archive in one flow, which is appropriately thorough for a screen test.

---

## Suggested before merge

1. Make the video preset effect respect explicit model selection (mirror the image-side `payload.model == default_image_model()` rule), and add a regression test where preset.model ≠ current model.
2. Either expand preset LoRAs into `loras` on the client OR add `apply_recipe_preset_to_video_payload` server-side — pick one and stop showing "Uses LoRA: X" until the job actually carries them.
3. Add Image Studio preset picker, or split sc-1312 and move Image to a follow-up.
4. Add weight controls to the LoRA picker; without them the editor cannot produce the LoRA shape the API and adapters expect.
5. Add an "Add by URL" entry point in the LoRA picker that triggers the sc-1309 import job.
6. Fix the `compatibleLoras` filter — model-family/workflow match, not `presetManaged`.
7. Drop or auto-derive the `modes` input in the editor.
8. Add the three sc-1312 tests the AC names explicitly.

PR #39, opened 2026-05-18
[codex] Complete next preset management stories

michaeltrefry:
## Summary

- Complete sc-1317 by documenting `modes` as the preset picker discoverability surface and adding Image Studio coverage for text/edit filtering semantics.
- Complete sc-1313 by adding worker-side LoRA loading for Diffusers image/video adapters, cache-safe LoRA clearing between jobs, pre-inference validation for missing/excess LoRAs, and installed path propagation from preset payloads.

## Validation

- `npm test` in `apps/web`
- `python -m pytest tests\test_worker_runtime.py`

## Notes

- `cargo fmt --all -- --check` is currently blocked by existing newline-style issues in Rust files, including files outside this PR scope.

### DIFF
--- a/apps/rust-api/src/lib.rs
+++ b/apps/rust-api/src/lib.rs
@@ -3874,6 +3874,8 @@ fn serialize_preset_lora(lora: &Value, preset_lora: &Value, lora_id: &str) -> Va
         "weight": preset_lora_weight(lora, preset_lora),
         "triggerWords": lora.get("triggerWords").cloned().unwrap_or_else(|| Value::Array(Vec::new())),
         "compatibility": lora.get("compatibility").cloned().unwrap_or_else(|| Value::Object(JsonObject::new())),
+        "installedPath": lora.get("installedPath").cloned().unwrap_or(Value::Null),
+        "source": lora.get("source").cloned().unwrap_or(Value::Null),
         "presetManaged": true
     })
 }
@@ -5619,6 +5621,12 @@ mod tests {
             "city at night, cinematic lighting"
         );
         assert_eq!(image_job["payload"]["loras"][0]["id"], "style-lora");
+        assert!(image_job["payload"]["loras"][0]["installedPath"]
+            .as_str()
+            .is_some_and(|value| value.ends_with("data/loras/style.safetensors")
+                || value.ends_with("data\\loras\\style.safetensors")
+                || value.ends_with("loras/style.safetensors")
+                || value.ends_with("loras\\style.safetensors")));
         assert_eq!(image_job["payload"]["model"], "client-model");
         assert_eq!(image_job["payload"]["count"], 2);
         assert_eq!(image_job["payload"]["seeds"].as_array().unwrap().len(), 2);
--- a/apps/rust-api/src/lib.rs
+++ b/apps/rust-api/src/lib.rs
@@ -5627,6 +5627,10 @@ mod tests {
                 || value.ends_with("data\\loras\\style.safetensors")
                 || value.ends_with("loras/style.safetensors")
                 || value.ends_with("loras\\style.safetensors")));
+        assert_eq!(
+            image_job["payload"]["loras"][0]["source"]["path"],
+            "loras/style.safetensors"
+        );
         assert_eq!(image_job["payload"]["model"], "client-model");
         assert_eq!(image_job["payload"]["count"], 2);
         assert_eq!(image_job["payload"]["seeds"].as_array().unwrap().len(), 2);
--- a/apps/web/src/main.test.jsx
+++ b/apps/web/src/main.test.jsx
@@ -724,6 +724,74 @@ describe("SceneWorks app shell", () => {
     expect(container.textContent).not.toContain("LTX Story");
   });
 
+  it("uses preset modes as the Video Studio picker surface", async () => {
+    root = createRoot(container);
+    await act(async () => {
+      root.render(
+        <VideoStudio
+          activeProject={{ id: "project-1", name: "Noir" }}
+          assets={[
+            { id: "image-1", type: "image", displayName: "Frame One" },
+            { id: "image-2", type: "image", displayName: "Frame Two" },
+          ]}
+          characters={[]}
+          createPersonDetectionJob={() => {}}
+          createPersonTrackJob={() => {}}
+          createVideoJob={() => {}}
+          deleteAsset={() => {}}
+          gpuOptions={["auto"]}
+          latestAssets={[]}
+          loras={[]}
+          onPreview={() => {}}
+          personTracks={[]}
+          purgeAsset={() => {}}
+          recipePresets={[
+            {
+              id: "camera_bridge",
+              name: "Camera Bridge",
+              workflow: "image_to_video",
+              modes: ["image_to_video", "first_last_frame"],
+              model: "ltx_2_3",
+            },
+            {
+              id: "start_frame",
+              name: "Start Frame",
+              workflow: "image_to_video",
+              modes: ["image_to_video"],
+              model: "ltx_2_3",
+            },
+          ]}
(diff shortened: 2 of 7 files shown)


michaeltrefry (commented):
Solid PR — the LoRA loader is well-shaped, the cache-by-key design is right, and adapter coverage is reasonably complete. A few notable items and one ergonomic naming issue that will trip future readers. Most findings are easy fixes.

**Recommendation: comment** — nothing here blocks merge, but a couple of items are worth addressing before sc-1313/1317 close.

---

## Notable

### 1. `assert_loras_supported` is named for the opposite of what it does
`apps/worker/scene_worker/lora_adapters.py`:

```python
def assert_loras_supported(loras: list[dict[str, Any]], adapter_id: str) -> None:
    if loras:
        raise RuntimeError(f"{adapter_id} does not support LoRA application for generation jobs.")
```

The function raises when LoRAs ARE present — it's a "this adapter doesn't support LoRAs, so fail" guard, called from `ProceduralImageAdapter.generate`, `ProceduralVideoAdapter.ensure_models`, etc. A future reader skimming `assert_loras_supported(request.loras, self.id)` at the top of `ZImageDiffusersAdapter.generate` would assume the adapter is asserting that LoRA support exists (which is the inverse). Rename to something like `reject_loras_if_unsupported` or `assert_no_loras` so the call site reads correctly.

### 2. `lora_cache_key` is order-sensitive
```python
def lora_cache_key(loras: list[dict[str, Any]]) -> str:
    specs = normalize_lora_specs(loras)
    return "|".join(f"{spec['id']}@{spec['path']}#{spec['weight']:g}" for spec in specs)
```

`[A, B]` and `[B, A]` produce different keys → cache miss → unload + reload everything. A common case is reordering LoRAs in the preset editor (or a UI that doesn't preserve order). Sort by adapter name (or id) before joining so the cache hits as long as the set + weights match. Performance only, no correctness issue.

### 3. Partial-overlap LoRA transitions reload unchanged adapters
`[A, B] → [A, C]` unloads everything (via `unload_lora_weights` which drops ALL adapters) then loads A and C. A is reloaded from disk unnecessarily. Acceptable for first cut, but worth noting on sc-1313 if it ever shows up in latency. The Diffusers `delete_adapters([name])` path can target individual adapters; reusing the existing-overlap diff would skip the round-trip.

### 4. sc-1317 AC asks for "at least one video workflow" filtering test
> Add tests covering preset filtering for **text-to-image, edit-image, and at least one video workflow**.

The new test covers text-to-image and edit-image in `ImageStudio`. The existing video preset test (from PR #38) uses `workflow: "image_to_video"` with no `modes` set — exercises the fallback, not the modes-based discoverability path that this story is specifically deciding. A short video test where `modes` ≠ `workflow` (e.g., a preset with `workflow: "image_to_video"` but `modes: ["image_to_video", "first_last_frame"]`, asserting it shows up under both modes) would close the AC properly.

### 5. `MAX_JOB_LORAS = 3` is a third copy of the same constant
The Rust API enforces ≤3 in `normalize_recipe_preset_loras`, the schema says `maxItems: 3`, and now the Python worker re-declares it. Three sources of truth for the same magic number. Promote to a shared contract constant (`crates/sceneworks-core`) and import on both sides, or at minimum cross-reference in a comment so a future bump touches all three.

### 6. `lora_cache_key`'s `|` separator can collide with paths
Linux filesystems allow `|` in filenames. `id_a@/path/with|pipe.safetensors#0.5` will produce ambiguous keys. Constrained in practice (the safe_download_dir / slug rules sanitize), but a tuple-based hash or JSON encoding would be drift-proof.

### 7. UI still shows "Preset incomplete" but lets the user click Generate
Carrying over from PR #38: now that the worker fails-before-inference for missing LoRAs (per `normalize_lora_specs`), the user clicks Generate, the job hits "failed" with "LoRA X is not installed". Functionally correct but the UX would be better if the client disabled the Generate button when `presetLoraDetails.some((lora) => lora.missing)`. Could be a sc-1315 polish item.

### 8. `lora_path` fallback chain includes a `source` string case that no live schema uses
```python
or (lora.get("source") if isinstance(lora.get("source"), str) else None)
or (lora.get("source") or {}).get("path")
```

`source` is always an object per the manifest schema (`{provider, repo/url, path}`). The bare-string fallback is dead. Drop it or document why it's kept.

### 9. Cache invalidation in `_load_pipeline` works but spreads the same `pop()` pattern in three places
`ZImageDiffusersAdapter._load_pipeline` has three branches each calling `self._loaded_lora_keys.pop(key, None); self._loaded_lora_names.pop(key, None)`. Easy to forget one in a future edit. A tiny helper `self._forget_loaded_loras(key)` would keep them in lockstep.

### 10. Tests don't cover the `set_adapters` fallback or the missing-`load_lora_weights` path
The three new Python tests cover happy-path cache, eviction, and validation failures. Not covered:
- adapter without `set_adapters` but with single LoRA at weight 1.0 (the implicit-supported path)
- adapter with `load_lora_weights` missing → expect "does not support loading LoRA weights"
- `assert_loras_supported` raises when LoRAs present (currently inferred from integration but never directly tested)
- weight-out-of-bounds (covered upstream in the API, but `lora_weight` parse-failure → default 0.8 has no test either)

---

## Nits

- `pipe.set_adapters(names, adapter_weights=...)` / fallback to `weights=...` — works across diffusers versions but a one-line comment naming which version uses which kwarg would save future archaeology.
- `safe_adapter_name` truncates to 48 chars; two long ids with the same 48-char prefix would collide at the same `index`. Fine in practice, but the comment-free magic number invites future bugs.
- `apply_loras_to_pipeline` returns `(key, names)` and callers store them. The pair is always read/written together — a tiny dataclass or a single dict would make the call sites less error-prone (it's easy to update one and forget the other).
- The Rust API serializer adds `installedPath` and `source` — there's no test asserting `source` actually round-trips into the job payload. The added test only checks `installedPath`.
- `tests/test_worker_runtime.py` uses `try/except: else: raise AssertionError(...)` instead of `pytest.raises(RuntimeError, match=...)`. Minor stylistic drift from pytest conventions in the rest of the file.

---

## What's good

- Clean separation: `lora_adapters.py` is its own module, both image and video adapters import it. No copy-paste of the load/unload logic across adapter files.
- `apply_loras_to_pipeline` is self-contained and pure-ish (mutates the pipe but doesn't touch global state) — testable with a fake pipe.
- Cache short-circuit `if key == previous_key` correctly returns early without touching the pipeline, which is the desired idempotent behavior.
- `if not specs: clear and return ""` cleanly handles the "job has no LoRAs but the pipe had some" case.
- `_evict_pipelines` clears both `_loaded_lora_keys` and `_loaded_lora_names` — easy to miss, and getting this wrong would silently stick stale LoRA state on the next pipeline. Glad it's caught.
- `assert_loras_supported` is wired into both procedural image and procedural video adapters, satisfying "unsupported adapter/model combinations fail before inference."
- `installedPath` is propagated through the Rust API serializer, the JS `serializePresetLora`, and consumed by the worker — full chain works without the worker having to re-resolve manifest paths.
- The sc-1317 decision lives as a code comment in `presetUtils.js`, exactly where future readers will look. Small but the right place.
- Validation messages from `normalize_lora_specs` are specific ("LoRA X is not installed", "LoRA X file is missing: path", "Generation supports at most 3 LoRAs per job") — matches sc-1310's "user-actionable" bar.

---

## Suggested before close

1. Rename `assert_loras_supported` to something whose name matches its inverted behavior.
2. Sort `lora_cache_key` inputs so reordering doesn't invalidate the cache.
3. Add a video-side filtering test where `modes` ≠ `workflow` to close sc-1317's AC.
4. Promote `MAX_JOB_LORAS` to a single shared source.
5. (Optional) consider a UI disable on missing-LoRA presets so the failure doesn't have to happen at the worker.

PR #40, opened 2026-05-18
[codex] clean up preset migration and validation UX

michaeltrefry:
## Summary
- Remove shipped builtin presets and placeholder builtin LoRAs so fresh installs start with user-managed presets only.
- Fix LoRA install-state normalization so entries without a real file path are reported as missing instead of installed.
- Change legacy preset read defaults to use the current model catalog instead of hardcoded model IDs, and expose read-time defaulting under documented `appliedDefaults` metadata.
- Consolidate preset validation copy so missing and incompatible LoRAs are reported by one shared path, and scope empty-state pending-import copy to compatible LoRAs.
- Add a regression guard that empty shipped preset/LoRA manifests return empty API catalogs.

## Validation
- `cargo fmt --package sceneworks-rust-api -- --check`
- `cargo test -p sceneworks-rust-api empty_builtin_preset_and_lora_manifests_ship_empty_catalogs --lib`
- `cargo test -p sceneworks-rust-api model_and_lora_routes_match_manifest_behavior --lib`
- `cargo test -p sceneworks-rust-api recipe_preset --lib`
- `cargo test -p sceneworks-core --test contract_roundtrip`
- `python -m pytest tests/test_rust_api_contract_snapshots.py -q`
- `npm test` in `apps/web`
- Browser check against `http://localhost:5174` for empty Preset Manager state

## Shortcut
- Completes sc-1314
- Completes sc-1315


michaeltrefry (commented):
The migration plumbing and the UX copy refresh are both solid. sc-1315's empty-state and validation-message work in particular nails the AC list. There's one significant finding that touches sc-1314's central AC — the builtin LoRAs the presets reference aren't actually installable — and a handful of smaller items.

**Recommendation: request changes** — the builtin LoRA install-path issue means every shipped preset will fail with a misleading "LoRA not installed" at generation time, which is the exact thing sc-1314 set out to prevent. The rest is cleanup.

---

## Critical

### 1. Builtin presets advertise LoRAs that have no installable file
sc-1314 AC, second bullet:

> Builtins are model/workflow scoped and **no longer misleadingly advertise LoRA application unless actual LoRA application is wired.**

The shipped builtins (`cinematic`, `photoreal`, the new `cinematic_edit`, etc.) all reference `builtin_*` LoRAs. Those LoRA manifest entries exist in `config/manifests/builtin.loras.jsonc`:

```jsonc
{
  "id": "builtin_cinematic_detail",
  ...
  "source": { "provider": "builtin" },   // ← no path
  ...
}
```

`source.path` is missing for all five builtin LoRAs. In `normalize_lora_entry` (`apps/rust-api/src/lib.rs`):

```rust
let source_path = object.get("source")...get("path")...;  // → None for builtins
let installed_path = source_path.map(...);                // → None
let install_state = if installed_path.as_ref().is_some_and(|path| !path.exists()) {
    "missing"
} else {
    "installed"   // ← falls here when path is None
};
```

So the API reports these LoRAs as `installState: "installed"` (with `installedPath: null`). The UI sees them as ready, save validation passes, the user clicks Generate, and the worker hits `lora_path()` in `lora_adapters.py`:

```python
def lora_path(lora):
    value = (lora.get("installedPath") or lora.get("sourcePath") or ...)
    if not value:
        return None
```

Returns `None`, so `normalize_lora_specs` raises:

> `LoRA builtin_cinematic_detail is not installed. Import or download it before generation.`

Two layers to fix:
- **The data**: either ship the LoRA files and give them a real `source.path`, or drop the LoRA references from the builtin presets and let users add their own LoRAs.
- **The bug**: `install_state = "installed"` when `installed_path` is `None` is wrong. A LoRA with no path should report `installState: "missing"` (or a new `"unknown"` state), so validation catches it at save / select time rather than letting the worker fail mid-job.

If the intent is "these are placeholder ids for future ship-with-app LoRAs," document that on sc-1314 and drop them from the builtin preset definitions until they exist. Without one of these fixes, the entire builtin preset set is broken on a fresh install.

---

## Notable

### 2. Hardcoded model IDs in `default_recipe_preset_model_for_workflow`
```rust
fn default_recipe_preset_model_for_workflow(workflow: &str) -> Option<&'static str> {
    match workflow {
        "text_to_image" | "edit_image" => Some("z_image_turbo"),
        "image_to_video" | "text_to_video" | "first_last_frame" => Some("ltx_2_3"),
        _ => None,
    }
}
```

Two model IDs baked into Rust. If `z_image_turbo` or `ltx_2_3` is renamed, retired, or split, the migration silently points legacy presets at a model that may not exist in the user's catalog. If the user doesn't have either model installed, the migrated preset is unusable. Worth deferring to the existing model catalog (first installed model matching the workflow), or surface as a config knob.

### 3. Migration runs on every read; notes are transient
`finalize_recipe_preset_entry` rebuilds the `migration` block on every `recipe_preset_catalog` call. If the user manifest is never re-saved, the same migration steps fire forever:
- Reading legacy preset → infer workflow, default model, expand modes, copy `builtInLoras → loras` → add migration notes.
- Re-reading 10 seconds later → all of the above again, fresh `migration` block.

Two side-effects:
- Wasted work on every list/get call (small but it grows with preset count).
- The `migration` block is a misleading API field: it doesn't mean "this preset was migrated at some point in the past" — it means "the read just synthesized defaults right now." A client seeing `migration.notes: ["model defaulted to z_image_turbo"]` might assume the user's manifest now has that model; it does not.

Either persist the migration on the next write (one-shot migration semantics) or rename the block to something that reflects what it actually represents (`appliedDefaults` rather than `migration`).

### 4. Two parallel save-blocker messaging systems with overlapping strings
The editor now produces two blocker messages from two different code paths:
- `presetValidationMessage(validation)` returns `"Save blocked: X has not finished importing..."` based on `presetValidation`.
- `saveDisabledReason` returns `"Save blocked: X is still importing..."` based on local `pendingSelectedLoras`.

Both describe the same condition (a referenced LoRA has `installState: "missing"`) with slightly different copy. The render order prefers `validationMessage` and only shows `saveDisabledReason` as a fallback. The new test asserts the validationMessage variant.

This is the kind of duplication that drifts: when someone tweaks copy on one path, the other goes stale, and depending on render order the change may or may not be visible. Consolidate to one source of truth (probably `presetValidation` since it's already shared across the studios).

### 5. `presetValidationMessage` only reports one category at a time
```js
if (validation.missing.length) {
  return `Save blocked: ${validation.missing.join(", ")} has not finished importing...`;
}
return `Save blocked: ${validation.incompatible.join(", ")} ...`;
```

If a preset has both a missing LoRA *and* an incompatible LoRA, the user fixes the missing one, clicks Save, gets a different error, and fixes the incompatible one. Two cycles. Either chain them ("Save blocked: X is still importing; Y is not compatible with the selected model") or be explicit about the order ("Resolve missing imports first, then re-check compatibility").

### 6. New `migration` field is undocumented in the schema
`packages/schemas/recipe-preset.schema.json` only adds a description on `builtInLoras`. The schema lets extra properties through (`additionalProperties: true`), so `migration` is accepted by accident. Worth a schema entry so external integrators know what to expect.

### 7. Edit variants reuse text-variant LoRAs
`cinematic_edit` / `photoreal_edit` / `product_edit` reference the same `builtin_cinematic_detail` / `builtin_photoreal_texture` / `builtin_product_finish` LoRAs as their text counterparts. Acceptable as a first cut, but in practice an "edit-tuned" LoRA usually wants different strength or different training. Worth a note on sc-1314 that edit variants share the text LoRA until separate tuning ships, so this doesn't get forgotten.

### 8. `loraEmptyMessage` "pending import" check is too broad
```js
loras.some((lora) => lora.installState === "missing")
  ? "No installed compatible LoRAs. Imports appear after the Queue completes."
  : ...
```

`some` looks at any LoRA in the catalog, not just LoRAs compatible with the currently selected model. A user on Z-Image with a pending Qwen import will see the "Imports appear after the Queue completes" message — but that pending Qwen LoRA will never appear under their Z-Image picker even when the import finishes. Scope the `some()` to compatible LoRAs.

---

## Nits

- `compactModeList(workflow)` translation table is a nice ergonomics win — text_to_image → "Text", edit_image → "Edit", etc. Worth lifting to a shared module so the Studios can use the same labels.
- `builtInLoras migrated to loras` migration note fires for empty arrays too (when both `loras` is absent and `builtInLoras` is `[]`). The note then technically reads "migrated an empty array." Minor noise.
- Tests still rely on `button.textContent === "Save Preset"` matching exact copy. Same pattern from PR #38 — sturdier role-based queries would prevent every copy tweak from breaking tests.
- `validationMessage` is computed even when `selectedPreset.scope === "builtin"` (where save is already blocked by `editable`). Cheap to compute, but it would be clearer to short-circuit.
- The new test passes both `imageModels=[]` AND `videoModels=[]` plus an `activeProject`. The empty model dropdown shows "No models". Fine, but adding a follow-up test with `videoModels=[{}]` and `imageModels=[]` to exercise the workflow-driven filtering of model dropdown would be useful.

---

## What's good

- Splitting the four builtins into `text` + `edit` pairs is exactly the right answer to the modes-vs-workflow drift sc-1317 surfaced. The diff is mechanical and contained.
- Read-time migration is non-destructive: the user's manifest stays as-is, the API returns enriched values. Safe for hand-edited manifests.
- Migration notes give debugging visibility ("model defaulted to z_image_turbo for legacy preset") — even though I'd rename the surface, the information itself is valuable.
- `presetLoraDetails` now correctly reports `missing: true` when `lora.installState === "missing"`. Previously it was hard-coded to `false` for any LoRA in the catalog regardless of install state — that was a real bug, now fixed.
- "Preset LoRA applied at generation: X" is a meaningful copy win over "Uses LoRA: X" — explicitly tells the user the LoRA WILL be applied (which is true now that sc-1313 actually applies them).
- Empty-state copy is specific and actionable ("No installed LoRAs match LTX. Imports appear after the Queue completes.").
- Save button disabling with helper copy ("Choose a model before saving.", "Choose a model before importing so compatibility can be recorded.") — the URL-import inputs being disabled when no model is selected closes the loop with sc-1309.
- The new test directly exercises the no-model + pending-import paths and verifies the Save button is `disabled` and `updateRecipePreset` is never called. Strong behavioral assertion.
- Schema gains a deprecation note on `builtInLoras` — small but the right place.

---

## Suggested before merge

1. Resolve the builtin LoRA situation — either ship the files with real `source.path` entries, or drop the LoRA references from builtin presets. Either way, fix `install_state` so a LoRA with no path doesn't report `"installed"`.
2. Consolidate `presetValidationMessage` and `saveDisabledReason` so only one source emits "Save blocked: ..." for each condition.
3. Defer hardcoded model IDs in the migration to either the model catalog or a config constant.
4. Either persist the migration on next write or rename the `migration` block to reflect that it's a per-read computed view.
5. Scope `loraEmptyMessage`'s pending-import check to LoRAs compatible with the selected model.

michaeltrefry (commented):
**Re-review — direction change.** The product call is now: no shipped/pre-packaged presets, no shipped LoRAs. Users create and manage their own. That flips the framing on this PR: my earlier "critical" finding about builtin LoRA install state becomes moot (those LoRAs shouldn't ship at all), but the PR is now actively moving in the wrong direction by adding three more shipped variants. The subtractive version is much smaller than what's currently here.

---

## What this PR should do instead

### Remove, don't add
1. **Drop the new builtin variants from `config/manifests/builtin.recipe-presets.jsonc`** — `cinematic_edit`, `photoreal_edit`, `product_edit` were added by sc-1314 here. Revert that change.
2. **Empty (or delete) `config/manifests/builtin.recipe-presets.jsonc`** — replace the five existing builtins (Cinematic, Photoreal, Anime, Fantasy, Product Shot) with `{ "schemaVersion": 1, "presets": [] }`, or delete the file. `load_manifest_entries` already returns an empty `Vec` on `NotFound` (`apps/rust-api/src/lib.rs:2913-2917`), so either form works.
3. **Empty (or delete) `config/manifests/builtin.loras.jsonc`** — remove the five `builtin_*` LoRA entries. Same NotFound-tolerant load path applies.
4. **Drop the `modes`-narrowing edits** on the existing five builtins (Cinematic etc. losing `edit_image` from their `modes`) since the presets themselves are going away.

### Verify nothing assumes builtins exist
- The `scope: "builtin"` enum value (`crates/sceneworks-core/src/contracts.rs:343`) and the read-only enforcement (`recipe_preset_readonly_or_not_found` at `apps/rust-api/src/lib.rs:3323-3338`) — keep these. They're forward-compat for a hypothetical future "vendor presets" feature and they're cheap.
- Tests that use `builtin_readonly` / `builtin_cinematic_detail` are test-fixture constructs that write the manifest inline — those stay, they're not shipped content.
- Confirm the Studio preset dropdowns render cleanly with zero presets (the picker uses `<option value="">No presets</option>` per VideoStudio, ImageStudio has its own — sanity-check both).
- Confirm `PresetManagerScreen`'s "No presets yet" empty state is reachable on a fresh install (already wired correctly in PR #38).
- Check whether `recipe_preset_scope_order` (`apps/rust-api/src/lib.rs:2996-3003`) still wants `Some("builtin") => 0` as the first tier when nothing ever lands in that tier. Cosmetic but worth thinking about.

---

## What from this PR is still worth keeping

- **All of the sc-1315 UX copy/validation work** is independent of the builtin question and stays valid:
  - "Preset LoRA applied at generation: X" (replacing the misleading "Uses LoRA: X")
  - "Save blocked: ..." messaging with explicit fix instructions
  - Empty-state copy for no-model, no-compatible-LoRAs, pending-import, etc.
  - `presetLoraDetails.missing = lora.installState === "missing"` fix
  - Save button disabling with helper copy
  - URL-import inputs disabled when no model is selected
- **Read-time migration in `finalize_recipe_preset_entry`** — still useful for users with hand-edited or legacy user/project preset manifests (`workflow` inference, `modes` defaulting, `builtInLoras → loras` rename). They might never have used a shipped builtin but they may have written their manifests against the older field names.
- **Schema deprecation note on `builtInLoras`** — fine, the field is still tolerated for backwards-compat on user-authored manifests.

---

## What needs reconsidering under the new direction

### 1. `default_recipe_preset_model_for_workflow`
Hardcoded `z_image_turbo` for image and `ltx_2_3` for video. If those models also aren't shipped/pre-installed (which is consistent with "no heavy models bundled"), the migration silently points legacy presets at models the user may not have. Two cleaner options:
- **Drop the default entirely** — leave `model` unset on migrated presets and let `validate_recipe_preset_model_workflow` produce a clean error message ("Recipe preset model not set; choose a model in the editor"). Forces an explicit user choice, which matches the "users own their presets" philosophy.
- **Default to the first installed model that supports the workflow** — read from `model_catalog`. More work, no hardcoded ids that go stale.

### 2. The "critical" builtin LoRA install-state bug from my first pass
`normalize_lora_entry` reporting `install_state = "installed"` when `installed_path` is `None` is still wrong — but it stops mattering for shipped LoRAs once they're gone. It's still a latent bug for *user-imported* LoRAs whose manifest entries somehow lack a path. Probably worth fixing defensively, but it's no longer blocking this PR.

### 3. The migration applies on every read
My earlier complaint still stands but the impact shifts. Without shipped builtins, the migration mostly runs on user-authored legacy presets, and the per-read regeneration of `migration.notes: [...]` is still noisy. One-shot persistence (or renaming the block to `appliedDefaults` to reflect that it's a per-read view, not history) is still the right answer.

### 4. The two-parallel-blocker-message duplication (`presetValidationMessage` vs `saveDisabledReason`)
Independent of the direction change; still worth consolidating.

---

## Suggested diff for the new direction

A small PR (or revised version of this one) along the lines of:

- `config/manifests/builtin.recipe-presets.jsonc` — empty `presets: []` (or `git rm`)
- `config/manifests/builtin.loras.jsonc` — empty `loras: []` (or `git rm`)
- Revert the three new `*_edit` builtin entries
- Drop or rework `default_recipe_preset_model_for_workflow` per #1 above
- Keep everything in `apps/web/src/screens/*.jsx`, `apps/web/src/presetUtils.js`, `apps/web/src/main.test.jsx`
- Keep the schema deprecation note
- Add a snapshot/contract test that exercises an empty catalog → empty `/api/v1/recipe-presets` and `/api/v1/loras` responses, so future regressions don't sneak shipped content back in

Once the manifests are empty, the existing recipe-preset CRUD tests still cover the happy paths (they write their own user-scope manifests inline). The empty-catalog test would be a quick guard.

michaeltrefry (commented):
**Re-review (third pass) — sorry, I was looking at a stale local fetch on my second pass.** With `cae2be0` and `d66b73a`, this PR now does what the new direction calls for. Walking through what changed:

---

## Items previously flagged, now addressed

- **`config/manifests/builtin.recipe-presets.jsonc` and `builtin.loras.jsonc` emptied** to `presets: []` / `loras: []` ✓
- **No more `*_edit` shipped variants** — the three additions from sc-1314's initial commit were reverted ✓
- **`install_state` no-path bug fixed** — `match installed_path.as_ref() { Some(path) if path.exists() => "installed", _ => "missing" }` is the right shape. A new test case (`no_path_style`) pins this end-to-end: a manifest entry with `source: { "provider": "local" }` and no path now correctly fails preset save with "Recipe preset LoRA is not installed: no_path_style." ✓
- **Hardcoded `z_image_turbo`/`ltx_2_3` model defaults removed** — `default_recipe_preset_model_for_workflow(models, workflow)` now reads from the live `model_catalog`, prefers installed models, falls back to first match. Much better. ✓
- **`migration` field renamed to `appliedDefaults`** — exactly the right framing for a per-read computed view rather than persisted history. Schema now documents it as "Read-only API metadata describing defaults synthesized while reading a legacy preset." ✓
- **Save-blocker messaging consolidated** — `validationMessage` now feeds directly into `saveDisabledReason` as the final fallback. One source of truth, no parallel string drift. ✓
- **`presetValidationMessage` compounds both missing AND incompatible** — `Save blocked: X has not finished importing; Y is not compatible with the selected model. Wait for imports to finish, remove incompatible LoRAs, or choose a matching model.` Single round-trip for the user. ✓
- **`loraEmptyMessage` pending-import check scoped to compatible LoRAs** — `hasPendingCompatibleLoras` only fires when a *compatible* LoRA is pending. ✓
- **`builtInLoras → loras` migration note suppressed for empty arrays** — `if migrated_count > 0` guard added. ✓
- **`compactModeList` lifted to `presetUtils.js`** — both Studios and the Manager can share. ✓
- **New `empty_builtin_preset_and_lora_manifests_ship_empty_catalogs` test** — exactly the snapshot guard I asked for. Future PRs that reintroduce shipped content will fail this test. ✓
- **`appliedDefaults` stripped from write payloads** — `strip_recipe_preset_write_context` and `strip_recipe_preset_runtime_fields` both remove it so the field never round-trips into a persisted manifest entry. ✓
- **Contract fixture renamed away from `cinematic`/`builtin_cinematic_detail`** — now `soft_glow` / `lora_fixture` to avoid implying shipped content. ✓

---

## Two small things still worth noting

### 1. `finalized_recipe_preset` calls `finalize_recipe_preset_entry(&mut preset, &[])`
`apps/rust-api/src/lib.rs:3257-3260` — single-preset finalization (used by Create/Update/Duplicate handlers) passes an empty `models` slice, so the model-defaulter never fires for those code paths. That's actually correct — by the time someone POSTs/PATCHes, `normalize_recipe_preset_for_write` has enforced that `model` is set (on create) or already present (on update). But the asymmetry between the two `finalize_recipe_preset_entry` callers is subtle. A two-line comment on `finalized_recipe_preset` saying "single-preset finalization runs without a model catalog because writes already require an explicit model" would save a future reader the trace.

### 2. `default_recipe_preset_model_for_workflow` falls back to the first uninstalled match
```rust
.find(|model| model_supports_recipe_workflow(model, workflow)
    && model.get("installState").and_then(Value::as_str) == Some("installed"))
.or_else(|| models.iter().find(|model| model_supports_recipe_workflow(model, workflow)))
```

If no installed model supports the workflow, the read still advertises a model that the user can't actually use until they install it. The downstream `validate_recipe_preset_model_workflow` will reject the preset on save, so this only bites on the read-side display, not on actual functionality. Two options if you want to tighten: drop the fallback (leave `model` unset and let the UI surface "Choose a model"), or include a hint in `appliedDefaults.notes` like "model defaulted to base-model (not installed); install the model to use this preset." Optional.

---

## Net assessment

The PR is now subtractive at the data layer (manifests emptied, fixture renamed) and additive at the code layer (defensive migration for user-authored legacy presets, schema docs, snapshot guard). That's the right shape for the new direction. Both sc-1314 and sc-1315 are honestly satisfied as written.

The `appliedDefaults`-as-per-read-view design is actually quite nice: the user's manifest stays clean, the API surfaces what defaults would apply if the user opened the editor, and the rename makes the semantics obvious. Good iteration.

Approving in spirit (GitHub won't let me actually approve my own PR). Suggest landing.

PR #41, opened 2026-05-18
[codex] Add local LoRA upload import

michaeltrefry:
## Summary
- Adds multipart LoRA uploads to `/api/v1/loras/import`, streaming the uploaded file into a staged server path and queueing the normal `lora_import` job with `sourcePath`.
- Keeps URL imports on the existing JSON path and preserves URL validation, family validation, scope handling, manifest payloads, and project/global targets.
- Updates the Preset Manager import panel with URL vs Local File modes, upload progress/disabled duplicate submit state, and file-selection UI.
- Cleans staged local upload files after the Rust worker finishes, fails, or cancels the LoRA import job.

## Validation
- `cargo test -p sceneworks-rust-api --lib`
- `cargo test -p sceneworks-rust-worker --lib`
- `npm test` in `apps/web`
- Browser sanity check of the Preset Manager Local File import mode at `http://localhost:5177`

### DIFF
--- a/apps/rust-api/src/lib.rs
+++ b/apps/rust-api/src/lib.rs
@@ -65,6 +65,7 @@ const HEARTBEAT_SSE_DATA: &str = "{}";
 #[cfg(test)]
 const HEARTBEAT_SSE_WIRE: &str = "event: heartbeat\ndata: {}\n\n";
 const MAX_UPLOAD_BYTES: usize = 2 * 1024 * 1024 * 1024;
+const MAX_MULTIPART_BODY_BYTES: usize = MAX_UPLOAD_BYTES + 16 * 1024 * 1024;
 const MANIFEST_CACHE_LIMIT: usize = 16;
 const MODEL_SIZE_CACHE_LIMIT: usize = 64;
 const API_MANAGED_MANIFEST_HEADER: &str = "// This file is rewritten by the SceneWorks API. Inline JSONC comments are not preserved across writes.";
@@ -512,7 +513,7 @@ pub fn create_app(settings: Settings) -> Result<Router, JobsStoreError> {
         )
         .fallback(route_not_found)
         .with_state(state.clone())
-        .layer(DefaultBodyLimit::max(MAX_UPLOAD_BYTES))
+        .layer(DefaultBodyLimit::max(MAX_MULTIPART_BODY_BYTES))
         .layer(middleware::from_fn_with_state(state, access_control))
         .layer(cors))
 }
@@ -883,6 +884,8 @@ struct LoraImportRequest {
     scope: String,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     project_id: Option<String>,
+    #[serde(default, skip_deserializing, skip_serializing_if = "bool_is_false")]
+    uploaded_source_path: bool,
 }
 
 async fn health(State(state): State<AppState>) -> Json<HealthResponse> {
@@ -2283,7 +2286,42 @@ async fn duplicate_recipe_preset(
 
 async fn create_lora_import_job(
     State(state): State<AppState>,
-    ApiJson(mut payload): ApiJson<LoraImportRequest>,
+    request: AxumRequest,
+) -> Result<(StatusCode, Json<JobSnapshot>), Response> {
+    let is_multipart = request
+        .headers()
+        .get(header::CONTENT_TYPE)
+        .and_then(|value| value.to_str().ok())
+        .is_some_and(|value| value.starts_with("multipart/form-data"));
+    if is_multipart {
+        let multipart = Multipart::from_request(request, &state)
+            .await
+            .map_err(|error| ApiError::bad_request(error.to_string()).into_response())?;
+        let (payload, staged_path) = lora_import_request_from_multipart(&state, multipart)
+            .await
+            .map_err(IntoResponse::into_response)?;
+        let result = queue_lora_import_job(state, payload).await;
+        if result.is_err() {
+            let _ = tokio::fs::remove_file(&staged_path).await;
+            if let Some(parent) = staged_path.parent() {
+                let _ = tokio::fs::remove_dir(parent).await;
+            }
+        }
+        return result.map_err(IntoResponse::into_response);
+    }
+
+    let payload = Json::<LoraImportRequest>::from_request(request, &state)
+        .await
+        .map(|Json(payload)| payload)
+        .map_err(json_rejection_response)?;
+    queue_lora_import_job(state, payload)
+        .await
+        .map_err(IntoResponse::into_response)
+}
+
+async fn queue_lora_import_job(
+    state: AppState,
+    mut payload: LoraImportRequest,
 ) -> Result<(StatusCode, Json<JobSnapshot>), ApiError> {
     if option_str_is_empty(payload.repo.as_deref())
         && option_str_is_empty(payload.source_url.as_deref())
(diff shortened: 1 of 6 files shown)


michaeltrefry (commented):
The bones of this are right — bounded-memory streaming, a server-generated staged path, an internal `uploaded_source_path` flag that can't be faked by clients, and worker cleanup wired into the existing `run_utility_job` exit. A few cleanup-on-error gaps, one pre-existing security concern this PR could fold in, and some smaller items. Otherwise this lands the sc-1322 ACs cleanly.

**Recommendation: comment** — nothing blocks merge, but #1 and #4 are worth resolving before this goes out.

---

## Notable

### 1. Streaming-write errors don't clean up the staged file/dir
`write_lora_upload_field_to_staged_file` (`apps/rust-api/src/lib.rs:2521-2562`) cleans up only on the size-exceeded branch:

```rust
if uploaded_bytes > MAX_UPLOAD_BYTES {
    let _ = tokio::fs::remove_file(&temp_path).await;
    let _ = tokio::fs::remove_dir(&upload_dir).await;
    return Err(...);
}
```

But `field.chunk().await?`, `file.write_all(&chunk).await?`, and `file.flush().await?` all early-return with the temp file/dir still on disk. Causes:
- a flaky client connection mid-upload → orphan
- disk full mid-write → orphan
- any axum/multipart parse error after the file was created → orphan

A scope guard (`scopeguard::defer!` or a manual `Drop` impl on a small `StagedUpload` wrapper) that removes the file/dir unless explicitly committed (`forget()`) would cover this without duplicating cleanup code at each `?`. Pattern: create staged dir, register defer, write chunks, on success call `commit()` which suppresses the defer.

### 2. No orphan sweep on startup
`run_utility_job`'s cleanup covers the worker-side success/failure path. The API-side cleanup in `create_lora_import_job` covers job-creation failure. But staged files survive in three cases:
- Client disconnect mid-upload (per #1 above)
- API process crash between staging and job enqueue
- Worker never picks up the job (worker offline, queue stuck)

`data_dir/cache/lora-uploads/upload-*` accumulates. The AC says "where practical," so this isn't strictly required, but a small sweep on app startup that removes `upload-*` directories older than, say, 24 hours would close the loop. Cheap to add and matches the spirit of "clean up temporary upload files on failure/cancel where practical."

### 3. Double disk I/O on the upload path
Staged file → `copy_lora_source` copies to final location → `cleanup_uploaded_lora_source` deletes the original. For a 2GB upload that's ~4GB of writes (staging + copy) plus a delete. A `tokio::fs::rename` would be O(metadata) if the staging dir and the LoRA dir are on the same filesystem. Worth attempting rename first, falling back to copy on `EXDEV`:

```rust
match tokio::fs::rename(&staged, &target).await {
    Ok(()) => return Ok(()),
    Err(error) if error.raw_os_error() == Some(libc::EXDEV) => {} // fall through to copy
    Err(other) => return Err(other.into()),
}
copy_lora_source(&staged, &target).await
```

Halves the I/O on the happy path. Pure optimization — not a correctness issue.

### 4. Pre-existing: arbitrary-file-read via JSON `sourcePath`
Not introduced here, but worth flagging since this PR touches the same handler. `queue_lora_import_job` accepts `sourcePath` from the JSON body without any path containment check (`apps/rust-api/src/lib.rs:2294-2403`). A client with API access can POST `{ "sourcePath": "/etc/passwd", "scope": "global" }` and the worker will copy that into `data/loras/<lora_id>/`. The new multipart path is safe (server-generated staged path under `data_dir/cache/lora-uploads/`), but the JSON path widens the trust boundary.

Two options:
- Restrict JSON `sourcePath` to a configured allowlist of roots (e.g., `data_dir/loras/imports` for local dev) and reject anything outside.
- Drop JSON `sourcePath` entirely now that we have multipart upload — the dev-loop use case it served can use multipart with `curl -F file=@...`.

I'd lean toward dropping it, but either way it's worth a follow-up story rather than letting this PR ship with the same surface area.

### 5. `MAX_MULTIPART_BODY_BYTES` applies to all routes
`apps/rust-api/src/lib.rs:516`:
```rust
.layer(DefaultBodyLimit::max(MAX_MULTIPART_BODY_BYTES))
```

Bumps the body limit for every endpoint from 2GB to ~2.016GB. The 16MB of headroom is reasonable for the upload route specifically, but applying it globally means every other endpoint also tolerates that much. Probably harmless — there's no other endpoint with a body that large — but a per-route `DefaultBodyLimit::max(MAX_MULTIPART_BODY_BYTES)` on the upload route, with the global limit kept at `MAX_UPLOAD_BYTES`, would be tidier.

---

## Nits

- **No upload progress** beyond "Uploading" button copy and a static status message. `fetch` doesn't expose upload progress without a `ReadableStream` body; XHR with `xhr.upload.onprogress` does. Acceptable per AC ("enough status text to make it clear the app is uploading") but a real progress bar on 2GB uploads would be a clear UX win.
- **No test for the size-limit rejection path.** `MAX_UPLOAD_BYTES + 1` bytes → 413 isn't covered. A small test (e.g., a 16-byte limit override for testing or a constructed multipart body > limit) would pin this.
- **`uploaded_source_path` reads a bit clunky** as a field name — `staged_upload: bool` reads more naturally and matches the worker's intent ("clean this up because we staged it"). Doesn't really matter; works fine as-is.
- **File extension `accept=".safetensors,.ckpt,.pt,.bin"`** is browser-side only. A malformed file passes through and fails later at adapter load. Fine — extension lies are cheap and content validation is the right authority.
- **Filename collision risk** when two uploads share filenames and don't specify `loraId`. Both slugify to the same id. The existing manifest dedup either rejects or replaces; worth a one-line code comment about which.
- **No resume support** on partial uploads. A failed 1.9GB upload restarts from byte 0. Out of scope for first cut, but a follow-up for large-file ergonomics.
- **Frontend cleanup `setFileInputKey((current) => current + 1)`** to reset the file input via re-mount is a common React idiom but reads as magic. A comment ("force re-mount of the file input so the same file can be selected again") would help.

---

## What's good

- **Streaming, not buffering** — chunks go straight to disk. Memory stays bounded. AC satisfied.
- **`uploaded_source_path` is `skip_deserializing`** — the trust boundary is right. A client can't post `uploadedSourcePath: true` to make the worker clean up a path the client points at. Internal-only flag, set server-side after the actual upload.
- **`sanitized_upload_filename` rsplit-then-`safe_download_dir`** — handles `../../evil.safetensors` and Windows-style paths cleanly. The fallback to `lora.safetensors` for empty / `download`-collision is a thoughtful guardrail.
- **Per-upload uuid directory** — concurrent uploads don't collide; the directory naming makes it obvious where to look in `data/cache/lora-uploads/upload-<uuid>/`.
- **Worker cleanup hook in `run_utility_job`** rather than only on success — covers worker-side failures too. The `if job.job_type == JobType::LoraImport` gate is the right level.
- **API cleanup on job-creation failure** — `if result.is_err() { remove_file; remove_dir }` block in `create_lora_import_job`. Half of the cleanup story (the other half being #1 above).
- **`busy = saving || importingLora`** — clean separation in the UI. Prevents the user from saving the preset while an upload is in flight that would change the LoRA library. Good ergonomics.
- **Segmented control for URL vs Local File** is a nice mode toggle, matches the SceneWorks pattern.
- **Worker test for cleanup** asserts both file removal AND parent directory removal. Good behavioral assertion.
- **Multipart test in the API** exercises the full request lifecycle and verifies the staged file lands with the right contents and filename.

---

## Suggested before merge

1. Wrap the staging-dir creation + file write in a defer-style guard so streaming-write errors clean up the temp file/dir too.
2. Decide on the JSON `sourcePath` arbitrary-file-read issue: either restrict to an allowlist or drop the field now that multipart exists. Worth at least filing a follow-up story.
3. Scope the body-limit bump to the upload route rather than globally.
4. (Optional) Try `tokio::fs::rename` first in the worker copy path, fall back to copy on cross-fs.

PR #42, opened 2026-05-18
[codex] Keep queued actions in context

michaeltrefry:
﻿## Summary
- Stop image generation, video generation, and model downloads from forcing users into the global Queue view.
- Add a shared local job progress card and render it in the Image Review, Video Review, and Models surfaces.
- Refresh model state when model-download jobs complete while keeping Queue available for explicit inspection.

## Shortcut
- https://app.shortcut.com/trefry/story/1324/keep-users-in-context-for-queued-actions

## Validation
- `npm test -- --run`
- `npm run build`
- Local browser smoke check at `http://127.0.0.1:5173` with no console errors.

### DIFF
--- a/apps/web/src/App.jsx
+++ b/apps/web/src/App.jsx
@@ -202,6 +202,9 @@ export function App() {
       if (job.status === "completed" && job.projectId && job.type === "person_detect") {
         refreshAssets(job.projectId);
       }
+      if (job.status === "completed" && job.type === "model_download") {
+        refreshData();
+      }
       if (job.status === "failed") {
         setError(failedJobNotice(job));
       }
@@ -592,10 +595,10 @@ export function App() {
   async function createImageJob(payload) {
     if (!activeProject) {
       setError("Create or open a project first.");
-      return;
+      return null;
     }
     try {
-      await apiFetch("/api/v1/image/jobs", token, {
+      const job = await apiFetch("/api/v1/image/jobs", token, {
         method: "POST",
         body: JSON.stringify({
           ...payload,
@@ -604,11 +607,13 @@ export function App() {
           requestedGpu,
         }),
       });
-      setActiveView("Queue");
+      setJobs((items) => [job, ...items.filter((item) => item.id !== job.id)].sort(sortNewest));
       setError("");
       refreshData();
+      return job;
     } catch (err) {
       setError(err.message);
+      return null;
     }
   }
 
@@ -766,7 +771,7 @@ export function App() {
   }
 
   async function createVideoJob(payload, options = {}) {
-    const { navigateToQueue = true } = options;
+    const { navigateToQueue = false } = options;
     if (!activeProject) {
       setError("Create or open a project first.");
       return null;
@@ -784,6 +789,7 @@ export function App() {
       if (navigateToQueue) {
         setActiveView("Queue");
       }
+      setJobs((items) => [job, ...items.filter((item) => item.id !== job.id)].sort(sortNewest));
       setError("");
       refreshData();
       return job;
@@ -1076,15 +1082,17 @@ export function App() {
 
   async function createModelDownloadJob(model) {
     try {
-      await apiFetch(`/api/v1/models/${model.id}/download`, token, {
+      const job = await apiFetch(`/api/v1/models/${model.id}/download`, token, {
         method: "POST",
         body: JSON.stringify({ requestedGpu: "auto" }),
       });
-      setActiveView("Queue");
+      setJobs((items) => [job, ...items.filter((item) => item.id !== job.id)].sort(sortNewest));
       setError("");
       refreshData();
+      return job;
     } catch (err) {
       setError(err.message);
+      return null;
     }
   }
 
@@ -1233,9 +1241,11 @@ export function App() {
             createImageJob={createImageJob}
             gpuOptions={gpuOptions}
             imageModels={imageModels}
+            jobs={jobs}
             latestAssets={latestImageAssets}
             launchRequest={studioLaunch}
             loras={loras}
+            onOpenQueue={() => setActiveView("Queue")}
             onPreview={setPreviewAsset}
             recipePresets={recipePresets}
             requestedGpu={requestedGpu}
@@ -1262,6 +1272,7 @@ export function App() {
             launchRequest={studioLaunch}
             loras={loras}
             jobs={jobs}
+            onOpenQueue={() => setActiveView("Queue")}
             onPreview={setPreviewAsset}
             personTracks={personTracks}
             recipePresets={recipePresets}
@@ -1308,7 +1319,13 @@ export function App() {
         ) : null}
 
         {activeView === "Models" ? (
-          <ModelManagerScreen jobs={jobs} loras={loras} models={models} onDownloadModel={createModelDownloadJob} />
+          <ModelManagerScreen
+            jobs={jobs}
+            loras={loras}
+            models={models}
+            onDownloadModel={createModelDownloadJob}
+            onOpenQueue={() => setActiveView("Queue")}
+          />
         ) : null}
 
         {activeView === "Editor" ? (
--- a/apps/web/src/App.jsx
+++ b/apps/web/src/App.jsx
@@ -1316,6 +1316,7 @@ export function App() {
             latestAssets={latestVideoAssets}
             launchRequest={studioLaunch}
             loras={loras}
+            jobs={jobs}
             localJobs={videoLocalJobs}
             onLocalJobCreated={(job) => rememberLocalGenerationJob("video", job)}
             onOpenQueue={() => setActiveView("Queue")}
--- a/apps/web/src/App.jsx
+++ b/apps/web/src/App.jsx
@@ -40,6 +40,8 @@ function failedJobNotice(job) {
   return `${label}: ${detail}`;
 }
 
+const localJobStackLimit = 4;
+
 export function App() {
   const [health, setHealth] = useState(null);
   const [access, setAccess] = useState({ authRequired: false });
@@ -49,6 +51,7 @@ export function App() {
   const [activeView, setActiveView] = useState("Library");
   const [projectName, setProjectName] = useState("");
   const [jobs, setJobs] = useState([]);
+  const [localGenerationJobIds, setLocalGenerationJobIds] = useState({ image: [], video: [] });
   const [workers, setWorkers] = useState([]);
   const [queueSummary, setQueueSummary] = useState(null);
   const [models, setModels] = useState([]);
@@ -68,6 +71,8 @@ export function App() {
   const [previewAsset, setPreviewAsset] = useState(null);
   const [studioLaunch, setStudioLaunch] = useState(null);
   const [error, setError] = useState("");
+  const activeViewRef = useRef(activeView);
+  const localGenerationJobIdsRef = useRef(localGenerationJobIds);
   const selectedTimelineIdRef = useRef(null);
   const timelineApplyQueueRef = useRef(Promise.resolve());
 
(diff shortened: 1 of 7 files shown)


michaeltrefry (commented):
Clean and focused. Lifts the forced-navigation friction without losing the global Queue view, and the JobProgressCard is a single reusable component dropped into three surfaces. Tests pin the negative case ("does not contain 'Jobs and GPUs'") which is exactly what you want for a "stop force-navigating" change. A few notable items around state lifetime and double-submission, plus a handful of UX nits.

**Recommendation: comment** — sc-1324's ACs are honestly satisfied, but #1 and #2 are worth addressing before this lands.

---

## Notable

### 1. `localJobIds` is per-Studio component state, lost on unmount
Both `ImageStudio.jsx` and `VideoStudio.jsx` keep `localJobIds` in component state. `apps/web/src/App.jsx` conditionally renders them (`{activeView === "Image" ? <ImageStudio ... /> : null}`), so navigating to another view unmounts the Studio and the local job tracking is gone. Coming back shows zero local cards even if the job is still running.

The AC doesn't directly cover the "navigate-away-and-back" case, but the spirit ("keep users in context") gets weaker if you have to babysit the page. Two options:
- Lift `localJobIds` to App so it survives view changes.
- Drop the explicit tracking entirely — derive the local stack from the global `jobs` state, filtered by `(type, projectId, status !== "completed", createdAt within last N minutes)`. The deduper in `setJobs` already handles the optimistic-insert case, so this reduces to a single derivation.

The second feels cleaner: the component doesn't need its own job ledger, and "show me jobs I recently spawned" falls out of the global state naturally.

### 2. Submit button isn't disabled during an in-flight job creation
`submit` is now async (`apps/web/src/screens/ImageStudio.jsx`, `VideoStudio.jsx`):

```js
async function submit(event) {
  event.preventDefault();
  const job = await createImageJob({...});
  if (job?.id) {
    setLocalJobIds((ids) => [job.id, ...ids.filter(...)].slice(0, 4));
  }
}
```

Double-clicking Generate sends two requests. The first finishes, the optimistic insert lands; the second finishes too and lands as a second job. There's no `disabled` on the submit button while the request is in flight, and no `submitting` state guarding the body of `submit`. Add a `submitting` state (or guard via `localJobs.length` if you want to limit to one in-flight per Studio). The disabled prop should reflect either.

### 3. Brief empty-state flash between "completed" status and asset refresh
The local stack filters out completed jobs:
```js
.filter((job) => job && job.status !== "completed");
```

When the job hits "completed", the card disappears immediately. `latestAssets` then refreshes async via `onJobUpdate → refreshAssets`. In the gap between those two events, `localJobs.length === 0` AND `latestAssets.length === 0`, so the review section falls back to `<div className="empty-panel">No fresh image batch</div>`. Flickers "No fresh image batch" for a moment before the new asset appears.

Two fixes:
- Show a "Generating finished — fetching asset..." card for the brief window between completion and asset load.
- Keep the completed card visible until `latestAssets` updates, then swap.

### 4. Failure shows up twice — once globally, once locally
`apps/web/src/App.jsx` line 207 still sets `setError(failedJobNotice(job))` on any failed job. The new `JobProgressCard` *also* shows the error inline. User sees the same failure in the top banner and in the Review card. Either:
- Suppress the global notice for jobs that have a local card.
- Keep both intentionally, but reword the global one ("Image job failed — see Review for details").

### 5. Long prompts have no truncation in `JobProgressCard`
`jobTitle(job)` returns `job.payload?.prompt ?? ...`. A 300-char prompt becomes an `<h3>` that fills the card. The CSS adds `overflow-wrap: anywhere` so it doesn't overflow horizontally, but the heading still dominates the layout. A `-webkit-line-clamp: 2` (or truncate at 80-120 chars with ellipsis) would keep the card a stable height.

### 6. `localDownloadFor` and `activeDownloadFor` overlap in ModelManagerScreen
Two filter helpers, both walking `jobs` for model_download entries matching the model id, with slightly different status filters. The existing `activeDownloadFor` and the new `localDownloadFor` could share an underlying filter and then branch on status. Minor cleanup; risk of drift if either filter logic changes.

---

## Nits

- **`localJobIds.slice(0, 4)` magic number** — 4 cards is a reasonable visual cap, but no comment explains why. One-liner: `// cap visual stack at 4 cards to keep the review pane usable in burst submissions`.
- **`onOpenQueue` only renders the button on error states.** For non-error states, the user can still navigate via the nav bar, which is fine — but a small "View in Queue" link on the running card would help users who started 3+ jobs and want the full list. Optional.
- **`status-badge` rendered with no variant class** in `JobProgressCard` — the existing `.status-badge.installed` / `.status-badge.failed` styles exist elsewhere; this one renders the base style only. Visual consistency check.
- **`refreshData()` on every model_download completion** fetches everything (models, loras, projects, jobs, ...). A focused `refreshModels()` would be cheaper, but `refreshData` is the existing helper and consistent — fine to defer.
- **`formatJobType(type).replaceAll("_", " ")`** produces lowercase "image generate". A capitalize wouldn't hurt ("Image generate" or "Image generation").
- **Tests use button-text matchers ("Image", "Video", "Models", "Download 12 GB").** Same brittle pattern as PRs 38/40 — copy changes break tests. Not a regression, just noted again.
- **Empty `?` fallback in `localJobs.length ? null : <empty-panel>`** is a bit dense; either an extracted helper variable (`const hasContent = localJobs.length || latestAssets.length`) or a short comment would help. Minor.

---

## What's good

- **`JobProgressCard` is a single, well-shaped component** with sensible default copy, an error affordance, and a clean prop surface. Reusing it across three screens with a `label` override is exactly right.
- **Optimistic job insert** (`setJobs((items) => [job, ...items.filter(...)].sort(sortNewest))`) means the progress card appears immediately, before the next SSE round-trip. The dedup-by-id filter handles the subsequent refresh cleanly.
- **`refreshData()` on `model_download` completion** is the right side-effect — users on the Models screen see "Ready" appear without having to switch back and forth.
- **`createVideoJob` flipping default `navigateToQueue` from `true` to `false`** preserves the option for callers that explicitly want the old behavior. Cleaner than ripping the option out.
- **Tests assert the negative case** (`expect(container.textContent).not.toContain("Jobs and GPUs")`) — that's the right shape for a "we no longer navigate to X" change. Future regressions that accidentally bring the navigation back will fail loudly.
- **The "Retry Download" button label** when a model_download is in a terminal-failed state (`failedDownload && terminalStatuses.has(localDownloadJob.status)`) is a thoughtful affordance — users who hit a download failure can retry directly without leaving the Models page.
- **`localJobs.length ? null : <empty-panel>`** correctly hides the "no fresh batch" message while a generation is in progress, so the user doesn't see contradictory copy.
- **Worker tests intact** — only frontend changed, the backend job lifecycle is untouched.

---

## Suggested before merge

1. Lift `localJobIds` to App, or derive `localJobs` from the global `jobs` state by type + recency — survives view changes.
2. Add a `submitting` state (or `disabled={localJobs.length}` if you want at-most-one) so double-clicking Generate doesn't spawn two jobs.
3. Bridge the "completed → asset appears" window so the empty panel doesn't flash.
4. Decide whether the global failure banner should still fire when the local card is visible.
5. Add a line-clamp to the JobProgressCard heading.

michaeltrefry (commented):
**Re-review.** `4810c21 Address queued progress review feedback` cleans up essentially everything from my first pass. Going through the prior items:

- `localGenerationJobIds: { image: [], video: [] }` lifted to App state, derived via `useMemo`, passed down as `localJobs`. Survives view changes. ✓
- `submitting` state in both Studios, button disabled and labeled "Queueing..." in flight; early-return guard inside the handler as belt-and-suspenders. ✓
- `resultVisible(job)` keeps the card visible until the result asset is in `latestAssets` (or `assets`), plus a "Finished. Fetching result..." message during the transition. Both halves of the flash addressed. ✓
- `hasVisibleLocalFailure(job)` suppresses the global error banner when the user is on the surface that already shows the local failure. Uses refs to dodge stale closures from the SSE callback. ✓
- `.local-job-main h3` gets `-webkit-line-clamp: 2` so long prompts stay bounded. ✓
- `downloadJobsFor(model)` consolidates the prior `activeDownloadFor` + `localDownloadFor` overlap into a single filter; render code branches on status. ✓
- `localJobStackLimit = 4` constant with a comment explaining the cap. ✓
- `onOpenQueue` now renders on all states with "Open Queue" / "View in Queue" copy. ✓
- `status-badge` gets variant classes (`.completed`, `.failed`, `.canceled`, `.interrupted`) with appropriate colors. ✓
- `formatJobType` capitalizes. ✓

That's all the notable items from the first review. Clean follow-up.

---

## One regression introduced by the follow-up

### `VideoStudio`'s `jobs` prop was dropped but the component still needs it
`apps/web/src/App.jsx` removed `jobs={jobs}` from the `<VideoStudio>` invocation as part of the rename to `localJobs={videoLocalJobs}`. But `apps/web/src/screens/VideoStudio.jsx:194-205` still uses `jobs` to derive `latestDetectionJob`:

```js
const latestDetectionJob = jobs
  .filter((job) =>
    job.type === "person_detect" &&
    job.status === "completed" &&
    job.projectId === activeProject?.id &&
    job.payload?.sourceAssetId === sourceClipAssetId,
  )
  .sort((a, b) => b.createdAt.localeCompare(a.createdAt))[0];
const detectionResult = latestDetectionJob?.result ?? null;
const representativeFrame = assets.find((asset) => asset.id === detectionResult?.frameAssetId);
const selectedDetection = detectionResult?.detections?.find((item) => item.id === selectedDetectionId) ?? detectionResult?.detections?.[0];
```

With the prop gone, `jobs` defaults to `[]` per the destructure default (`apps/web/src/screens/VideoStudio.jsx:26`). `latestDetectionJob` is `undefined`, `detectionResult` is `null`, and `representativeFrame` / `selectedDetection` are both undefined. The Replace Person mode then can't show the detection frame or the detection list at all — the panel that normally lets the user click a detected person to lock the track silently goes empty.

The existing tests don't cover the Replace Person flow with a completed detection job, so this slips through. Fix is one line: re-add `jobs={jobs}` to the VideoStudio invocation. (Or rename the prop to make the dual purpose clearer — `localJobs` for the review stack, `jobs` for the global filter — and leave both.)

---

## Smaller observations

### 1. `resultVisible` depends on a server-side contract I can't verify from the diff
```js
function resultVisible(job) {
  if (job.result?.generationSetId) {
    return latestAssets.some((asset) => asset.generationSetId === job.result.generationSetId);
  }
  const assetIds = job.result?.assetIds ?? [];
  return assetIds.length > 0 && assetIds.every((id) => assets.some((asset) => asset.id === id));
}
```

If the completion event arrives with `job.status === "completed"` but `job.result.generationSetId` is unset AND `job.result.assetIds` is empty (or missing), this returns `false` forever, and the card stays in `localJobs` until it gets pushed out by 4 newer submissions. Worse UX than the original brief flash.

Worth confirming: does the existing image/video job completion path actually populate `job.result.generationSetId` (or `assetIds`) when `onJobUpdate` fires? If the server populates them at a later moment than the status flip to "completed," there's a race where the card persists longer than expected. A defensive timeout ("if the job has been completed for >N seconds and `resultVisible` is still false, hide it anyway") would close the loop without depending on the server contract.

### 2. No new tests for the new behaviors
The three existing test cases still pass (they're submit-then-assert-status flows). Worth adding:
- Asserts that after a "completed" status arrives, the card stays visible until `latestAssets` updates, then disappears.
- Asserts that submitting=true disables the button (prevents the double-submit regression).
- Asserts that `hasVisibleLocalFailure` suppresses the global error banner.

These pin the new logic; the next refactor will benefit.

### 3. `localJobs.length || latestAssets.length` is recomputed every render
Inline in both Studios — fine for now, just noting that `useMemo` would be cheap insurance if either becomes large. Not worth optimizing yet.

---

## Net

The follow-up commit lands all the substantive improvements I asked for and goes further on the polish (`status-badge` variants, "View in Queue" copy, capitalized job type). The one regression on VideoStudio's `jobs` prop is a quick fix — re-add the prop and the Replace Person flow is back. Once that's resolved I'd merge.

PR #43, opened 2026-05-18
[codex] Add LoRA imports to Models page

michaeltrefry:
﻿## Summary

Adds LoRA import controls directly to the Models page so users can queue LoRA imports from either a source URL or a local upload while choosing global/project scope and model-family compatibility.

## What changed

- Added a Models-page LoRA import panel with URL and upload modes.
- Kept imports initiated from Models on the Models page while showing active LoRA import progress there.
- Refreshed LoRA catalog data when `lora_import` jobs complete.
- Added focused UI coverage for URL imports and local file uploads from the Models page.

## Validation

- `npm test`
- `npm run build`
- `cargo test -p sceneworks-core lora_source_urls_validate_scheme_host_and_filename`
- Browser smoke check on `http://localhost:5175` for the Models LoRA import panel and source toggle behavior.

### DIFF
--- a/apps/web/src/App.jsx
+++ b/apps/web/src/App.jsx
@@ -41,6 +41,7 @@ function failedJobNotice(job) {
 }
 
 const localJobStackLimit = 4;
+const maxLoraUploadBytes = 2 * 1024 * 1024 * 1024;
 
 export function App() {
   const [health, setHealth] = useState(null);
@@ -228,6 +229,7 @@ export function App() {
       }
       if (job.status === "completed" && job.type === "lora_import") {
         refreshData();
+        // refreshData loads the global catalog; project imports need the project overlay too.
         if (job.projectId) {
           refreshLoras(job.projectId);
         }
@@ -436,6 +438,9 @@ export function App() {
       throw new Error("Create or open a project first.");
     }
     const { file, ...metadata } = payload;
+    if (file?.size > maxLoraUploadBytes) {
+      throw new Error("Uploaded LoRA file exceeds the 2GB limit");
+    }
     let body;
     if (file) {
       body = new FormData();
@@ -460,7 +465,8 @@ export function App() {
       method: "POST",
       body,
     });
-    if (options.navigateToQueue ?? true) {
+    setJobs((items) => [job, ...items.filter((item) => item.id !== job.id)].sort(sortNewest));
+    if (options.navigateToQueue ?? false) {
       setActiveView("Queue");
     }
     setError("");
@@ -1380,7 +1386,7 @@ export function App() {
             loras={loras}
             models={models}
             onDownloadModel={createModelDownloadJob}
-            onImportLora={(payload) => createLoraImportJob(payload, { navigateToQueue: false })}
+            onImportLora={createLoraImportJob}
             onOpenQueue={() => setActiveView("Queue")}
           />
         ) : null}
--- a/apps/web/src/App.jsx
+++ b/apps/web/src/App.jsx
@@ -226,6 +226,12 @@ export function App() {
       if (job.status === "completed" && job.type === "model_download") {
         refreshData();
       }
+      if (job.status === "completed" && job.type === "lora_import") {
+        refreshData();
+        if (job.projectId) {
+          refreshLoras(job.projectId);
+        }
+      }
       if (job.status === "failed" && !hasVisibleLocalFailure(job)) {
         setError(failedJobNotice(job));
       }
@@ -425,7 +431,7 @@ export function App() {
     return archived;
   }
 
-  async function createLoraImportJob(payload) {
+  async function createLoraImportJob(payload, options = {}) {
     if (payload.scope === "project" && !activeProject) {
       throw new Error("Create or open a project first.");
     }
@@ -454,7 +460,9 @@ export function App() {
       method: "POST",
       body,
     });
-    setActiveView("Queue");
+    if (options.navigateToQueue ?? true) {
+      setActiveView("Queue");
+    }
     setError("");
     refreshData();
     return job;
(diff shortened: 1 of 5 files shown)


michaeltrefry (commented):
Small, focused, mostly reuses pieces from prior PRs (`createLoraImportJob`, `JobProgressCard`). Two real items: the new opt-out leaves the Preset Manager still force-navigating to Queue (regression-by-omission), and `lora_import` doesn't get the optimistic-job-insert treatment that the other "stay in context" flows from PR #42 received. Otherwise it's polish.

**Recommendation: comment** — neither item blocks, but #1 should be addressed since it's adjacent to the change and contradicts the sc-1324 direction.

---

## Notable

### 1. Preset Manager LoRA imports still force-navigate to Queue
`apps/web/src/App.jsx` now reads:

```js
async function createLoraImportJob(payload, options = {}) {
  ...
  if (options.navigateToQueue ?? true) {
    setActiveView("Queue");
  }
  ...
}
```

The Models invocation correctly passes `{ navigateToQueue: false }`. But the Preset Manager still calls `createLoraImportJob({...})` with no options (`apps/web/src/screens/PresetManagerScreen.jsx:317`), so it defaults to `true` and force-navigates to Queue — exactly the friction sc-1324 was set up to remove. The user who's mid-edit on a preset, queues a URL import, gets yanked over to Queue and loses the editor context.

Two ways to fix:
- Flip the default to `false` and update any caller that wants the old behavior to opt in.
- Update the Preset Manager call site to pass `{ navigateToQueue: false }`.

The first is cleaner and matches the sc-1324 direction across the app — "default is local progress, force-navigation is opt-in."

### 2. No optimistic job insert for `lora_import`
PR #42 added optimistic inserts to `createImageJob`, `createVideoJob`, and `createModelDownloadJob`:

```js
setJobs((items) => [job, ...items.filter((item) => item.id !== job.id)].sort(sortNewest));
```

`createLoraImportJob` doesn't do this. Effect: the user clicks Queue Import, the form clears, the success message shows ("LoRA import queued for X."), but the `JobProgressCard` on the Models page doesn't appear until `refreshData()` round-trips. Inconsistent with the pattern the other staying-in-context flows established. Add the same optimistic insert in `createLoraImportJob` and the progress card will render immediately, matching the other surfaces.

### 3. `refreshLoras(job.projectId)` may be redundant with `refreshData()`
```js
if (job.status === "completed" && job.type === "lora_import") {
  refreshData();
  if (job.projectId) {
    refreshLoras(job.projectId);
  }
}
```

`refreshData()` is the "fetch everything" helper that's used elsewhere as the catchall (per PR #42 it's called for model_download completion too). If it already covers loras for the active project, the explicit `refreshLoras(job.projectId)` is redundant. If it doesn't, then global LoRA imports (`!job.projectId`) won't trigger any LoRA-specific refresh and would rely on `refreshData()` happening to include the global LoRA catalog.

Worth either dropping the redundant call or making the side-effect explicit so future readers know which path is which.

---

## Smaller observations

### 4. "Queue Import" button label doesn't change for URL mode in-flight
```jsx
<button disabled={importDisabled} type="submit">
  {importingLora && isFileImport ? "Uploading" : "Queue Import"}
</button>
```

URL-mode imports also have an `importingLora === true` window — the request goes out, the server creates the job, the response comes back. Currently the button stays "Queue Import" and just gets disabled. A small consistency: "Queueing..." for URL mode while `importingLora` is true, mirroring the studios' "Queueing..." pattern from PR #42.

### 5. `activeLoraImportJobs` shows ALL in-flight LoRA imports, not just Models-initiated ones
```js
const activeLoraImportJobs = jobs.filter((job) => job.type === "lora_import" && !terminalStatuses.has(job.status));
```

If the user queues a LoRA import from the Preset Manager and then navigates to Models, they see it on Models too. Probably the right behavior — it's a global "imports in flight" indicator — but worth noting it's a Models-page surface for what's effectively a global state. Could rename the visual to "LoRA imports in progress" to match.

### 6. No client-side file size guard
The server enforces a 2GB limit (PR #41). The Models import form doesn't echo that. A 3GB upload attempts and then fails server-side after the bytes are wasted. Cheap to add a `file.size > 2 * 1024 * 1024 * 1024` check at submit time.

### 7. No test for the failure path
Both new tests are happy-path. `onImportLora` throwing → `importMessage.tone === "error"` path isn't exercised. One small test ("import fails with bad URL, error message appears, importingLora resets") would pin the error UX.

### 8. `setFileInputKey((current) => current + 1)` magic re-mount
Same idiom from PR #41. Works. A one-line comment ("force-remount the file input so the same file can be selected again") would save future readers.

### 9. The `family` dropdown defaults to `families[0]` if the form's family isn't in the list
That's the right UX (sticky if valid, sensible fallback if not), but the effect dep `[families.join("|"), familyFilter]` is a stringified array as a dep — common React workaround. A one-liner comment or `useMemo`-derived `familiesKey` would be a touch cleaner.

---

## What's good

- **Reuses `createLoraImportJob` with an options bag** rather than duplicating the upload logic in `ModelManagerScreen`. Clean composition.
- **Reuses `JobProgressCard`** for the in-flight imports — visual consistency with the other staying-in-context surfaces.
- **Family defaulting from the family filter** is thoughtful UX — if the user is looking at Qwen LoRAs, the import form defaults to Qwen family.
- **`importDisabled` is well-shaped** — covers all four blockers (in-flight, no callback, project-without-activeProject, missing input). Submit button stays correctly off.
- **"Open a project before importing a project LoRA"** helper copy when the user picks project scope without an open project.
- **`<option disabled={!activeProject} value="project">`** properly disables the project option in the dropdown, complementing the helper copy and submit guard. Belt-and-suspenders.
- **Two integration tests** exercise URL import (global) and file upload (project), including the "Upload" mode toggle and the active-jobs progress card display.
- **`importMessage` distinguishes tones** (`success` vs `error`) and renders with the appropriate `.inline-success` / `.inline-warning` classes — matches the consolidated messaging pattern from PR #40.
- **`refreshData() + refreshLoras(job.projectId)` on `lora_import` completion** means the newly imported LoRA actually appears in the list without a manual refresh.

---

## Suggested before merge

1. Address #1 — either flip the `navigateToQueue` default to `false` or update the PresetManagerScreen call site. Otherwise the Preset Manager flow regresses against the sc-1324 direction by omission.
2. Add the optimistic `setJobs` insert in `createLoraImportJob` so the progress card appears immediately on Models, matching image/video/model_download behavior.
3. Decide on #3 — either drop the redundant `refreshLoras` call or document why both run.

PR #44, opened 2026-05-18
[codex] Show live GPU utilization in Queue

michaeltrefry:
﻿## Summary

Implements Shortcut story sc-1328 by wiring live GPU utilization through the worker heartbeat contract, Queue UI, and dispatch ranking.

## What changed

- Added optional worker utilization snapshots with GPU memory total/used/free and GPU load.
- Updated Python GPU workers to report `nvidia-smi` utilization on registration and heartbeat.
- Updated Queue GPU cards to hide CPU utility workers and show available memory, used/total memory, memory percentage, GPU load, warm models, and assigned GPU details for active jobs.
- Updated GPU job claiming so auto GPU work defers from a heavily loaded or memory-constrained worker when a compatible idle GPU has better live headroom.
- Kept Rust utility worker and Rust API contract builders compatible with the new optional field.

## Validation

- `cargo test`
- `npm test`
- `npm run build`
- `python -m pytest tests/test_worker_gpu.py tests/test_worker_runtime.py`
- Browser smoke check against a mocked Queue response confirmed GPU stats render and CPU utility workers are hidden from the Queue resource cards.

### DIFF
--- a/apps/rust-api/src/lib.rs
+++ b/apps/rust-api/src/lib.rs
@@ -2777,6 +2777,7 @@ async fn register_worker(
             gpu_name: payload.gpu_name,
             capabilities: payload.capabilities,
             loaded_models: payload.loaded_models,
+            utilization: payload.utilization,
         })
     })
     .await?;
@@ -2796,6 +2797,7 @@ async fn heartbeat_worker(
             status: payload.status,
             current_job_id: payload.current_job_id,
             loaded_models: payload.loaded_models,
+            utilization: payload.utilization,
         })
     })
     .await?;
--- a/apps/rust-worker/src/lib.rs
+++ b/apps/rust-worker/src/lib.rs
@@ -724,6 +724,7 @@ async fn register_worker(
             gpu_name: Some(gpu.name.clone()),
             capabilities: worker_capabilities(gpu),
             loaded_models: Vec::new(),
+            utilization: None,
             extra: BTreeMap::new(),
         },
     )
@@ -742,6 +743,7 @@ async fn heartbeat(
             status,
             current_job_id: current_job_id.map(str::to_owned),
             loaded_models: Vec::new(),
+            utilization: None,
             extra: BTreeMap::new(),
         },
     )
--- a/apps/rust-worker/src/lib.rs
+++ b/apps/rust-worker/src/lib.rs
@@ -10,7 +10,7 @@ use reqwest::StatusCode;
 use sceneworks_core::contracts::{
     ClaimRequest, ClaimResponse, ContractNumber, JobSnapshot, JobStatus, JobType, JsonObject,
     ProgressRequest, ProgressStage, WorkerCapability, WorkerHeartbeatRequest,
-    WorkerRegisterRequest, WorkerSnapshot, WorkerStatus,
+    WorkerRegisterRequest, WorkerSnapshot, WorkerStatus, WorkerUtilizationSnapshot,
 };
 use sceneworks_core::lora_url::{
     lora_source_url_file_name, lora_source_url_file_stem, parse_lora_source_url_with_private,
@@ -228,11 +228,12 @@ where
     Ok(response.json::<T>().await?)
 }
 
-#[derive(Debug, Clone, Eq, PartialEq)]
+#[derive(Debug, Clone, PartialEq)]
 struct DiscoveredGpu {
     id: String,
     name: String,
     capabilities: Vec<WorkerCapability>,
+    utilization: Option<WorkerUtilizationSnapshot>,
 }
 
 #[derive(Debug, Clone, Eq, PartialEq)]
@@ -467,7 +468,7 @@ async fn query_nvidia_gpus() -> Vec<DiscoveredGpu> {
         Duration::from_secs(3),
         Command::new("nvidia-smi")
             .args([
-                "--query-gpu=index,name,memory.total",
+                "--query-gpu=index,name,memory.total,memory.used,memory.free,utilization.gpu",
                 "--format=csv,noheader,nounits",
             ])
             .output(),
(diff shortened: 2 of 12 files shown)


michaeltrefry (commented):
Solid end-to-end implementation — contract extension, Python worker collection via nvidia-smi, store migration, dispatch logic, UI cards, and tests on both sides. The dispatch heuristic is the most interesting piece and it's reasonable but worth a closer read. A few items worth resolving before merge.

**Recommendation: comment** — sc-1328's ACs land in the cases this PR exercises, but #1 makes the dispatch improvement partial in mixed setups.

---

## Notable

### 1. Rust workers don't report utilization, so the smart dispatch is Python-only
`apps/rust-worker/src/lib.rs:724-744` sends `utilization: None` on both register and heartbeat. `should_defer_auto_gpu_claim` in `jobs_store.rs` short-circuits when the current worker has no utilization:

```rust
let current_score = dispatch_score(job, worker);
if !current_score.has_utilization {
    return Ok(false);
}
```

And `dispatch_score_is_better` requires both candidate and current to have utilization. Practical effect:

- Pure Python GPU setup → deferral works across GPUs.
- Pure Rust GPU setup → deferral never fires.
- **Mixed setup** → a Python worker may correctly defer to another Python worker, but a Rust worker on a low-load GPU is invisible to the deferral logic — a heavily loaded Python worker will happily claim while a fresh Rust GPU sits idle.

The AC says "If a GPU has heavy memory usage and/or high load, it should be skipped in favor of a GPU with more available memory and lower load when one is available." That's not quite honored in mixed setups. Either:
- Add nvidia-smi collection to the Rust worker (small `tokio::process::Command` wrapper, mirrors the Python helper).
- Or treat "no utilization data" as "unknown utilization, assume fully available" so a low-load Rust worker still wins over a heavily loaded Python worker. (Trade-off: now the deferral might incorrectly skip toward an unknown Rust worker that's actually busy.)

The first feels right — symmetry with the Python collection, no new heuristic gaps. Worth flagging on the story even if it lands as a follow-up.

### 2. `dispatch_score_is_better` thresholds are magic numbers
`crates/sceneworks-core/src/jobs_store.rs` — `dispatch_score_is_better` has a dense set of bounds:

```rust
let candidate_is_not_worse = candidate.free_memory_mb + 512.0 >= current.free_memory_mb
    && candidate.gpu_load_percent <= current.gpu_load_percent + 10.0
    && candidate.memory_usage_percent <= current.memory_usage_percent + 10.0;
let candidate_relief = free_delta >= 1024.0 || load_delta >= 15.0 || usage_delta >= 10.0;
...
(current.free_memory_mb < 2048.0 && candidate.free_memory_mb >= 4096.0)
    || (current.gpu_load_percent >= 85.0 && candidate.gpu_load_percent <= 75.0)
    || (current.memory_usage_percent >= 90.0 && candidate.memory_usage_percent <= 80.0)
```

These read reasonably but `512`, `1024`, `10`, `15`, `85`, `75`, `90`, `80`, `2048`, `4096` are all numbers with no explanation. A future engineer tuning this will need to reconstruct the intent. Named constants (`MEM_NOT_WORSE_TOLERANCE_MB`, `LOAD_NOT_WORSE_TOLERANCE_PCT`, `MEM_RELIEF_THRESHOLD_MB`, `HIGH_LOAD_THRESHOLD_PCT`, etc.) plus a one-paragraph comment explaining the policy ("prefer a candidate that's at least N MB freer or M% less loaded, with hysteresis so we don't ping-pong between equally-loaded GPUs") would make this maintainable.

### 3. No live-update test
The PR description claims "GPU utilization values update live without a full page refresh," and the AC repeats this. The new frontend test renders static workers prop with utilization populated and asserts the strings appear. It doesn't exercise: SSE worker update → React re-render → new utilization values in the card. The wiring is via existing worker SSE machinery, but a small test that mutates the `workers` prop between renders and asserts the new values appear would pin the live-update behavior. Cheap to add.

### 4. Memory meter shows 0%-width bar when memory data is unknown
```jsx
<div className="worker-meter" aria-label={`GPU memory usage ${utilizationLabel(memoryPercent)}`}>
  <span style={{ width: `${memoryPercent ?? 0}%` }} />
</div>
```

When `memoryPercent` is null (no utilization data), the bar renders at 0% width — looks identical to "empty / fully available." The `aria-label` correctly says "Unknown," but visually the user sees an empty bar and might think the GPU has plenty of memory free. Either hide the bar when unknown (`memoryPercent !== null && <div .../>`) or use a different visual treatment (striped, gray, "?" overlay).

### 5. `should_defer_auto_gpu_claim` runs an `active_gpu_job_exists` query per candidate
For typical setups (≤4 GPUs) this is fine, but it's N queries per claim. A single query that joins workers + active jobs would be cheaper if this ever runs hot. Optimization, not a correctness issue — flagging.

### 6. AC "Optional CPU/system display, if present" is not implemented
The AC explicitly allows this to be omitted ("optional ... if present"), so this isn't a miss. Just noting that we deliberately ship without it.

---

## Smaller observations

- **`nvidia-smi` invoked per heartbeat per worker** — Python `heartbeat()` calls `gpu_utilization()` which subprocess-spawns `nvidia-smi`. For 4 workers heartbeating every 10s, ~24 invocations per minute. Reasonable for now, but a 1-2s process-level cache on `query_nvidia_gpus()` would coalesce concurrent heartbeats.
- **`loads_optional` silently returns None on parse failure** (`jobs_store.rs`). For utilization specifically that's the right call (best-effort), but if a future caller uses this helper for a required field, the silent swallow would mask schema issues. A one-line comment scoping its intent to best-effort fields would prevent misuse.
- **`gpu_utilization(gpu_id)` falls through silently if the GPU id isn't in the list** — returns None. Probably the right default, but on a misconfigured `SCENEWORKS_GPU_ID` it silently disables utilization without telling anyone. A debug log when this happens would help diagnose future "why isn't dispatch working" puzzles.
- **`isGpuWorker` filter** requires `worker.gpuId && worker.gpuId !== "cpu"` AND `capabilities.includes("gpu")`. A future GPU worker that forgets to advertise the `gpu` capability would get filtered out silently. Worth a one-line comment about the filter's intent so future readers don't try to add their worker by setting `gpuId` alone.
- **The "Memory" stat shows `Unknown` only when BOTH used and total are non-finite** — looking at `WorkerCard`:
  ```jsx
  {Number.isFinite(usedMb) && Number.isFinite(totalMb) ? `${formatMemory(usedMb)} / ${formatMemory(totalMb)}` : "Unknown"}
  ```
  If only one is finite, you'd see "Unknown" rather than a partial value. Correct behavior; just noting that the predicate is "all-or-nothing."
- **`worker.assignedGpu` for active jobs in `QueueScreen.jsx:281`** is preserved via existing rendering (`workersById` map and `job.assignedGpu`). AC "Queue still makes it clear which GPU is assigned to active jobs" is honored by existing code; nothing changed here, just confirming it still works after the worker filter change.
- **No test for the `worker_supports_job` extraction** — the function is new (extracted from the inline filter in `choose_claimable_job`). The behavior matches the old inline version, but a unit test pinning the extraction would prevent regression if `WorkerCapability.as_str()` ever changes.

---

## What's good

- **Contract extension is clean** — `WorkerUtilizationSnapshot` with all-optional fields, `#[serde(default, skip_serializing_if = "Option::is_none")]` on the parent so older workers keep working.
- **Store migration via `ensure_column`** — idempotent ALTER TABLE on startup. Existing `jobs.db` files get the new column without manual intervention.
- **Rust worker preserves the contract**: explicitly sends `utilization: None` rather than omitting, so the typed `WorkerRegisterRequest` / `WorkerHeartbeatRequest` constructors stay symmetric across runtimes. Easy place to add real values later.
- **Frontend filter** correctly hides CPU utility workers and placeholder GPUs while preserving the assigned-GPU display in the job list. The negative assertions (`not.toContain("Rust CPU utility worker")`, `not.toContain("Rust placeholder GPU")`) are exactly the right shape — future regressions that bring CPU workers back will fail loudly.
- **`WorkerCard` component is well-factored** — three stat fields, two distinct meters (memory + load) with different colors, accessible labels. Graceful degradation when utilization is null (shows "Unknown" rather than crashing).
- **Dispatch test (`auto_gpu_claim_defers_to_less_loaded_compatible_worker`)** exercises the exact end-to-end path: register two workers with very different utilization, queue a job, assert the loaded worker's claim returns None and the idle worker gets it. Strong behavioral assertion.
- **Hysteresis baked in** — `dispatch_score_is_better` has +512 MB / +10% tolerance windows, so dispatch doesn't ping-pong between two GPUs with near-identical utilization. Right instinct.
- **`active_gpu_job_exists` excludes utility job types** — `type not in ('model_download', 'lora_import')` — so a model download on a GPU doesn't prevent the deferral logic from treating that GPU as idle for generation work.
- **Python parser handles both old and new `nvidia-smi` formats** — `if len(parts) < 3` still rejects malformed lines, `if len(parts) >= 6` adds utilization opportunistically. Backward compatible.
- **`parse_int` / `parse_float` swallow `TypeError`/`ValueError`** — defensive against unexpected nvidia-smi output formats.

---

## Suggested before merge

1. Decide whether the Rust worker should also collect utilization (or whether mixed setups are out of scope for this slice). At minimum, note on sc-1328 that smart dispatch is Python-only until the Rust collector lands.
2. Name the dispatch thresholds with constants + a one-paragraph policy comment.
3. Hide or visually differentiate the memory bar when utilization is unknown.
4. (Optional) Add a small live-update test that re-renders QueueScreen with updated worker props and asserts the new values appear.

PR #45, opened 2026-05-19
[codex] Keep LoRA upload form usable

michaeltrefry:
## Summary

Fixes Shortcut story sc-1363 by keeping the Models page LoRA upload flow usable after an import is accepted into the queue.

## What changed

- Reset the Models LoRA import fields and file input after a successful queue response so another import can be started immediately.
- Render queued/running/failed LoRA import jobs below the form as local progress cards, including independent failed states.
- Keep LoRA import progress visible through family filtering when family metadata is missing, and show a clearer filtered empty state when installed LoRAs are excluded.
- Refresh LoRAs with the active/project overlay after queueing and completion so project-scoped uploads do not disappear behind a global-only reload.
- Use LoRA name/id in local job cards instead of falling back to opaque job IDs.

## Root cause

The Models page tied in-flight upload state too closely to the form and only represented non-terminal LoRA imports inside the form. Catalog refreshes could also briefly replace project-aware LoRA data with a global-only list, making completed project LoRAs appear to vanish.

## Validation

- `npm test` in `apps/web` passed: 32 tests.
- `npm run build` in `apps/web` passed.
- Browser smoke opened the local Models page; API was not running, so queue behavior is covered by component tests.

### DIFF
--- a/apps/web/src/App.jsx
+++ b/apps/web/src/App.jsx
@@ -72,6 +72,7 @@ export function App() {
   const [previewAsset, setPreviewAsset] = useState(null);
   const [studioLaunch, setStudioLaunch] = useState(null);
   const [error, setError] = useState("");
+  const activeProjectRef = useRef(null);
   const activeViewRef = useRef(activeView);
   const localGenerationJobIdsRef = useRef(localGenerationJobIds);
   const selectedTimelineIdRef = useRef(null);
@@ -144,6 +145,10 @@ export function App() {
   useEffect(() => {
     activeViewRef.current = activeView;
   }, [activeView]);
+
+  useEffect(() => {
+    activeProjectRef.current = activeProject;
+  }, [activeProject]);
 
   useEffect(() => {
     localGenerationJobIdsRef.current = localGenerationJobIds;
@@ -228,11 +233,7 @@ export function App() {
         refreshData();
       }
       if (job.status === "completed" && job.type === "lora_import") {
-        refreshData();
-        // refreshData loads the global catalog; project imports need the project overlay too.
-        if (job.projectId) {
-          refreshLoras(job.projectId);
-        }
+        refreshDataWithLoraOverlay(job.projectId ?? activeProjectRef.current?.id);
       }
       if (job.status === "failed" && !hasVisibleLocalFailure(job)) {
         setError(failedJobNotice(job));
@@ -370,6 +371,14 @@ export function App() {
     }
   }
 
+  function refreshDataWithLoraOverlay(projectId = activeProjectRef.current?.id) {
+    refreshData().then(() => {
+      if (projectId) {
+        refreshLoras(projectId);
+      }
+    });
+  }
+
   async function refreshRecipePresets(projectId = activeProject?.id) {
     try {
       const query = projectId ? `?projectId=${encodeURIComponent(projectId)}` : "";
@@ -470,7 +479,7 @@ export function App() {
       setActiveView("Queue");
     }
     setError("");
-    refreshData();
+    refreshDataWithLoraOverlay(metadata.scope === "project" ? activeProject?.id : activeProjectRef.current?.id);
     return job;
   }
 
@@ -672,7 +681,7 @@ export function App() {
     if (active === "Video" && localIds.video.includes(job.id)) {
       return true;
     }
-    return active === "Models" && job.type === "model_download";
+    return active === "Models" && (job.type === "model_download" || job.type === "lora_import");
   }
 
   async function withCharacterApi(callback) {
--- a/apps/web/src/App.jsx
+++ b/apps/web/src/App.jsx
@@ -372,11 +372,13 @@ export function App() {
   }
 
   function refreshDataWithLoraOverlay(projectId = activeProjectRef.current?.id) {
-    refreshData().then(() => {
-      if (projectId) {
-        refreshLoras(projectId);
-      }
-    });
+    refreshData()
+      .then(() => {
+        if (projectId) {
+          refreshLoras(projectId);
+        }
+      })
+      .catch(() => {});
   }
 
   async function refreshRecipePresets(projectId = activeProject?.id) {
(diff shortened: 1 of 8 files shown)


michaeltrefry (commented):
Focused fix for sc-1363. The combination of (a) project-overlay refresh on completion, (b) keeping the progress block outside the form, (c) inline failed state, and (d) family-aware empty-state copy lands the AC list cleanly. Two small items worth noting, neither blocks merge.

**Recommendation: comment** — this is ready as-is; #1 and #2 are polish.

---

## Notable

### 1. Failed lora_imports hidden by the family filter get no visible signal
`hasVisibleLocalFailure(job)` now suppresses the global error banner for *any* `lora_import` failure when the user is on the Models view:

```js
return active === "Models" && (job.type === "model_download" || job.type === "lora_import");
```

But `localLoraImportJobs` is filtered by `matchesFamily(job, familyFilter)`. A failed qwen-image LoRA import while the user filters by z-image:
- Doesn't show as an inline card (filtered out).
- Doesn't show in the global banner (suppressed because user is on Models view).
- Counts toward `hiddenImportCount` ("1 queued LoRA import is hidden by this family filter") — which (a) calls it "queued" even though it's failed, and (b) the user might not connect the count to a failure.

Two ways to fix:
- Make `hasVisibleLocalFailure` aware of the family filter (requires lifting filter state, awkward).
- Only suppress the global banner when the failure is *actually* in `localLoraImportJobs`. App could pass a callback that returns whether the job is visible, called from `onJobUpdate`. Or simpler — keep showing the global banner for lora_import failures and accept the brief double-notification on the common case.

Practical recommendation: keep the global banner for lora_import failures. Inline failures are visible most of the time, but when they're not (filter mismatch or user on a different surface), the banner is the last line of defense.

### 2. `hiddenImportCount` copy says "queued" for non-queued statuses
```js
{hiddenImportCount} queued LoRA import{hiddenImportCount === 1 ? " is" : "s are"} hidden by this family filter.
```

`localLoraImportJobs` excludes only `completed`, so the filter includes `queued`, `running`, `preparing`, `downloading`, `failed`, `canceled`. The copy "X queued LoRA imports" is misleading when those are running or failed. A simple "X LoRA imports are hidden by this family filter" is accurate across all statuses.

---

## Smaller observations

- **`refreshDataWithLoraOverlay` chains `.then` without `.catch`.** If `refreshData()` rejects, you get an unhandled promise rejection. `refreshData` is supposed to swallow errors internally, so this is theoretical, but a defensive `.catch(() => {})` would be polite. Minor.
- **Asymmetric use of `activeProject` vs `activeProjectRef.current`** in `createLoraImportJob`:
  ```js
  refreshDataWithLoraOverlay(metadata.scope === "project" ? activeProject?.id : activeProjectRef.current?.id);
  ```
  Inside a click-handler-derived async function, `activeProject` and `activeProjectRef.current` are identical (the ref is updated synchronously via useEffect on every render). The ref is genuinely needed in `onJobUpdate` and `hasVisibleLocalFailure` (called from SSE callbacks where closures go stale), but here both forms work. Picking one for consistency (`activeProject?.id` in both branches) would make the asymmetry not invite head-scratching.
- **`loraFamilies(item)`** does a long fallback chain that mixes top-level fields and `payload.manifestEntry.*` fields. Reasonable defensive code given the dual purpose (LoRA records and lora_import jobs), but a one-line comment explaining "this helper accepts either a LoRA catalog entry or a lora_import job snapshot" would save a future reader from chasing the empty-family fallback in `matchesFamily`.
- **`matchesFamily` special-cases lora_import jobs with no family metadata as always-matching** — sensible because the job's family might not be on the snapshot yet. Worth a comment explaining why this isn't applied to LoRA catalog entries (because catalog entries should have family metadata after the import completes).
- **No new test for the App-level project-overlay refresh.** `refreshDataWithLoraOverlay` is a new helper that's exercised in two places (`onJobUpdate` and `createLoraImportJob`). A test that mocks `refreshData` and `refreshLoras` and asserts both fire with the right project id on lora_import completion would pin this against regression. Existing tests cover the screen behavior, not the App-level refresh wiring.
- **Moving the progress block outside the `<form>`** is mostly a layout cleanup since the JobProgressCard's "Open Queue" was already `type="button"` and wouldn't submit. The "form gets stuck" symptom is really fixed by the field-reset + finally-block from PR #43 (which I see is preserved); the moved progress block is an additional bit of cleanup so the failed-state card can sit independently of the form's lifecycle. Worth mentioning in the PR body so reviewers don't think the unparenting is the central fix.

---

## What's good

- **`refreshDataWithLoraOverlay`** correctly fires the project overlay on both initial submit and completion, with `job.projectId ?? activeProjectRef.current?.id` as the fallback chain. This addresses the "completed project LoRAs vanish behind a global refresh" failure mode named in the story.
- **`activeProjectRef` is the right tool** for the SSE-callback case where the captured `activeProject` would go stale. The useEffect pattern matches the existing `activeViewRef` / `localGenerationJobIdsRef` precedent from PR #42.
- **Inline failed state** with the error message rendered via `JobProgressCard` (which already had error-state styling and the "Open Queue" affordance from PR #42). Clean reuse.
- **The empty state now distinguishes** between "you have no LoRAs at all" and "your filter excludes the LoRAs you have" — the latter case explicitly names the active family. The hidden-import count is the same idea applied to in-flight jobs.
- **`jobTitle()` fallback chain** now prefers `payload.name` and `payload.loraId` for lora_import jobs, so the card shows a meaningful name instead of an opaque job id. Small but high-impact polish.
- **Test harness pattern** (an inline `<Harness>` component with its own jobs state) is the right approach for testing a parent-orchestrated flow without dragging the full App in. Cleanly tests the "first import → form clears → second import works" loop.
- **`hasVisibleLocalFailure` extended for lora_import on Models view** — matches the model_download precedent so the user doesn't get the double-notification PR #42 worked to eliminate.
- **Header count `visibleLoraCount = visibleLoras.length + localLoraImportJobs.length`** — the user sees pending imports counted in "X installed or pending" so the number doesn't temporarily drop while imports are in flight. Nice consistency.

---

## Suggested before merge

1. Decide on #1 — either keep the global banner for `lora_import` failures (simplest), or scope the suppression to actually-visible failures.
2. Drop "queued" from the `hiddenImportCount` message so the copy is accurate across statuses.
3. (Optional) Add a unit test for `refreshDataWithLoraOverlay` to pin the project-overlay wiring.

PR #46, opened 2026-05-19
[codex] Replace preset LoRA import with selector

michaeltrefry:
﻿## Summary
- Replace Preset Manager LoRA URL/file import controls with an Add LoRA selector backed by registered installed LoRAs.
- Keep selected preset LoRAs editable/removable with weight controls, plus missing/incompatible warnings.
- Point empty Preset Manager LoRA states to the Models page instead of offering uploads inline.
- Fix Docker rust-worker config path/mount so global LoRA upload jobs validate against the same manifest path queued by the API.

## Why
LoRA acquisition and management now belong on the Models page. Preset Manager should only reference LoRAs that are already uploaded or registered, avoiding duplicate import flows and confusing ownership.

## Root Cause For Upload Fix
The API queued global LoRA upload jobs with `/sceneworks/config/manifests/user.loras.jsonc`, but the Docker rust-worker service did not receive `SCENEWORKS_CONFIG_DIR` or the shared config mount. The worker fell back to `config/...` during manifestPath validation and rejected the otherwise-global upload.

## Validation
- `npm test -- --run src/main.test.jsx` from `apps/web`
- `npm run build` from `apps/web`
- `npm run check:compose`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,6 @@
+# SceneWorks Agent Notes
+
+## Pull Requests
+
+- For this repository, create pull requests with authenticated `gh pr create` directly.
+- Do not try the GitHub connector PR creation first; it repeatedly fails with `Resource not accessible by integration` and then requires falling back to `gh` anyway.
--- a/apps/web/src/App.jsx
+++ b/apps/web/src/App.jsx
@@ -1359,12 +1359,12 @@ export function App() {
         {activeView === "Presets" ? (
           <PresetManagerScreen
             activeProject={activeProject}
-            createLoraImportJob={createLoraImportJob}
             createRecipePreset={createRecipePreset}
             deleteRecipePreset={deleteRecipePreset}
             duplicateRecipePreset={duplicateRecipePreset}
             imageModels={imageModels}
             loras={loras}
+            onOpenModels={() => setActiveView("Models")}
             recipePresets={recipePresets}
             updateRecipePreset={updateRecipePreset}
             videoModels={videoModels}
--- a/apps/web/src/main.test.jsx
+++ b/apps/web/src/main.test.jsx
@@ -722,8 +722,8 @@ describe("SceneWorks app shell", () => {
     expect(importCalls).toBe(0);
   });
 
-  it("keeps Preset Manager LoRA imports in context", async () => {
-    const createdJobs = [];
+  it("keeps Preset Manager LoRA acquisition on the Models page", async () => {
+    let importCalls = 0;
     global.fetch.mockImplementation((url, options = {}) => {
       const path = new URL(url).pathname;
       if (path.endsWith("/health")) {
@@ -744,17 +744,11 @@ describe("SceneWorks app shell", () => {
         );
       }
       if (path.endsWith("/jobs")) {
-        return Promise.resolve(response(createdJobs));
+        return Promise.resolve(response([]));
       }
       if (path.endsWith("/loras/import") && options.method === "POST") {
-        const job = {
-          id: "lora-import-job-1",
-          type: "lora_import",
-          status: "running",
-          payload: { loraId: "preset_detail" },
-        };
-        createdJobs.unshift(job);
-        return Promise.resolve(response(job));
+        importCalls += 1;
+        return Promise.resolve(response({ id: "lora-import-job-1", type: "lora_import", status: "running" }));
       }
       return Promise.resolve(response([]));
     });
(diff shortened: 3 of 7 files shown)


michaeltrefry (commented):
Clean execution of sc-1365 — entire import panel ripped out, selector substituted in, empty state directs to Models. The auto-advance behavior on `selectedLoraToAdd` makes multi-LoRA flow smooth. The docker-compose fix is a real bug catch and the regression-asserting check-compose extension is the right shape. Couple of small items but nothing blocking.

**Recommendation: comment.**

---

## Notable

### 1. Mixed PR scope — sc-1365 + an unrelated rust-worker compose fix
sc-1365 is purely UI; the docker-compose env/mount addition is a runtime bug found during testing (per the PR body). They're co-located here, which is fine for shipping speed but worth noting:
- Future archaeology on "when did global LoRA upload start working in Docker?" lands on a UI PR.
- Reverting the UI change (if needed) would also revert the compose fix.

You've called this out in the PR body, so I'll just acknowledge it. If this lands at a slower cadence, splitting into two commits (or two PRs) keeps the history cleaner.

### 2. Empty state can render alongside the selected-LoRA list
Looking at the render block:

```jsx
{!availableLoras.length ? (
  <div className="empty-panel compact-panel">
    <span>{loraEmptyMessage}</span>
    {onOpenModels ? <button onClick={onOpenModels} type="button">Open Models</button> : null}
  </div>
) : null}
{form.loras.length ? (
  <div className="lora-choice-list">
    ...
  </div>
) : null}
```

If a preset has selected LoRAs but none are currently "available" (e.g., all selected LoRAs are missing or incompatible), the user sees both the "No installed LoRAs match X. Open Models" empty panel AND the warning-state list of their selected-but-broken LoRAs.

It's not strictly wrong — both signals are accurate — but it reads as "you have no LoRAs (but also here are some)." Either:
- Suppress the empty panel when `form.loras.length` is non-zero, since the warning cards already convey "broken selection."
- Or reword the empty state to say "No new LoRAs to add" when there are existing selections to keep the messages consistent with each other.

### 3. Weight input stays editable for missing/incompatible selected LoRAs
```jsx
<input ... disabled={!editable} ... value={selected?.weight ?? ""} />
```

A missing LoRA can have its weight changed even though the LoRA itself is broken. Tweaking a weight on a not-installed LoRA is meaningless. Disabling the weight input when `missing || incompatible` would nudge the user toward Remove or fixing the install.

---

## Smaller observations

- **No test for the Remove button.** The PR introduces a "Remove" affordance per selected LoRA but no test asserts that clicking Remove deletes the LoRA from `form.loras`. The existing CRUD test exercises Add + Weight; a one-liner adding + removing would pin the inverse path.
- **No test for the warning state** on missing/incompatible selected LoRAs. The render adds `.warning` class and the "Missing or still importing" / "incompatible with X" copy, but no assertion pins it. A small render-test fixture with a stale LoRA reference would lock the warning UX.
- **`addableLoras` filter auto-advance is a nice UX detail** — worth a one-line comment on the useEffect explaining "after adding a LoRA, pick the next one so the user can keep adding without re-opening the dropdown."
- **Disabled `<select>` with `<option value="">No LoRAs available</option>` when `addableLoras` is empty** — works but the option is decorative since the select can't be interacted with. Either an inline empty-state inside the picker container, or just rely on the Open Models call-to-action below, would be a slightly cleaner presentation.
- **`assertServiceMountsTarget` assumes volumes are object-form (`{target: ...}`).** Docker compose normalizes short-form to long-form during parsing, so this should be fine, but if anyone hand-writes short-form volumes (`./config:/sceneworks/config`) before parsing, the assertion would miss. The current compose file uses short-form (`${SCENEWORKS_CONFIG_BIND:-./config}:/sceneworks/config`) — confirm `loadComposeConfig` (whatever that does in the script) normalizes consistently.
- **No end-to-end regression test for the docker-compose fix** — `check-compose-config` validates the env and mount are present, but doesn't exercise a real global LoRA upload through rust-worker. Acceptable for a compose-only change, but worth a follow-up integration test once the harness exists.
- **`hasPendingCompatibleLoras` branch in `loraEmptyMessage`** is now reachable only when `installedLoras.length > 0 && availableLoras.length === 0`. Subtle precedence but correct.
- **`onOpenModels` defensively rendered** — good defensive coding, but every consumer in this PR passes it. Worth deciding if it's truly optional or a required prop with a defensive null check; small inconsistency that future readers might puzzle over.

---

## What's good

- **Full removal of the import panel** — no orphaned `importingLora` / `importForm` / `fileInputKey` state, no dead `importLora` handler, no leftover JSX. Clean excision.
- **`addSelectedLora()` + `removeLora()`** are tight, focused functions; the slice-to-3 cap is preserved.
- **Auto-advance of `selectedLoraToAdd`** after adding via the useEffect dep on `addableLoras` — adding multiple LoRAs is a click-click-click loop rather than a click-pick-click loop. Nice.
- **Missing/incompatible warnings are now per-LoRA in the selected list**, not just an aggregate validation message. The colored border (`.lora-choice.warning`) makes the broken entry visually distinct. Easier to identify what to remove.
- **Empty state directs to Models page** with an "Open Models" button rather than just text — matches the story's "path to manage/upload LoRAs."
- **Test coverage updated comprehensively** — the existing "keeps Preset Manager LoRA imports in context" test was renamed and rewritten to assert the absence of import controls, the presence of the empty-state copy, and that "Open Models" navigates. Strong shape — future regressions that bring import controls back will fail loudly.
- **`docker-compose.yml` fix addresses a real runtime issue** — the symmetric env/mount between API and rust-worker for the config dir prevents the "queued path doesn't validate against the worker's path" failure mode. The compose-check extension prevents this from regressing.
- **PR description honestly separates the UI work from the compose fix** with a "Root Cause For Upload Fix" subsection. Easy to understand what's what.
- **`SCENEWORKS_CONFIG_BIND` defaulting to `./config`** matches the existing `SCENEWORKS_DATA_BIND` pattern — no surprise for users.

---

## Suggested before merge

1. Decide on #2 — either suppress the empty panel when there are selected LoRAs, or reword it to be consistent with the warning cards.
2. Add a test for the Remove button (small) and ideally one for the missing/incompatible warning state.
3. (Optional) Disable the weight input for selected LoRAs that are missing or incompatible.

michaeltrefry:
﻿Addressed the review items except #1 as requested.

Changes pushed in `5bdc557`:
- Suppressed the LoRA empty panel when selected LoRAs already exist, so broken selections show only their warning cards.
- Disabled weight editing for missing or incompatible selected LoRAs.
- Added coverage for Remove and selected-LoRA warning/disabled-weight behavior.
- Added a short comment explaining add-selector auto-advance.
- Hardened the compose mount assertion to tolerate normalized object volumes and raw short-form volume strings.
- Added the repo AGENTS.md note to use `gh pr create` directly for SceneWorks PRs.

Validation:
- `npm test -- --run src/main.test.jsx` from `apps/web`
- `npm run build` from `apps/web`
- `npm run check:compose`

PR #47, opened 2026-05-19
[codex] Restore model download size estimates

michaeltrefry:
## Summary
- Add manifest-backed fallback download size estimates for model downloads when live Hugging Face lookup is disabled or unavailable.
- Populate built-in model download entries with current Hugging Face tree-size estimates.
- Show estimated sizes on the Models page with a `~` marker and show `Unavailable` when no size is known.

## Validation
- `cargo test -p sceneworks-rust-api model_`
- `cargo build -p sceneworks-rust-api`
- `python -m pytest tests/test_rust_api_contract_snapshots.py::test_system_manifest_and_http_contracts`
- `npm --prefix apps/web test -- src/main.test.jsx`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -212,6 +212,7 @@ apps/web/node_modules/
 apps/web/dist/
 
 # SceneWorks local runtime data
+config/manifests/user.loras.jsonc
 data/projects/*
 !data/projects/.gitkeep
 data/recent-projects.json
--- a/apps/rust-api/src/lib.rs
+++ b/apps/rust-api/src/lib.rs
@@ -4331,6 +4331,7 @@ fn model_download_context(model: &Value) -> Result<Option<DownloadContext>, ApiE
 }
 
 fn manifest_download_size_bytes(model: &Value, download: &Value) -> Option<u64> {
+    // Prefer the selected download entry, then fall back to legacy model-level metadata.
     ["estimatedSizeBytes", "downloadSizeBytes", "sizeBytes"]
         .iter()
         .find_map(|field| download.get(*field).and_then(json_size_to_u64))
--- a/apps/rust-api/src/lib.rs
+++ b/apps/rust-api/src/lib.rs
@@ -3025,6 +3025,7 @@ fn publish<T: Serialize>(state: &AppState, event: &str, data: &T) {
 struct DownloadContext {
     repo: String,
     files: Vec<String>,
+    fallback_size_bytes: Option<u64>,
 }
 
 async fn model_catalog(state: &AppState) -> Result<Vec<Value>, ApiError> {
@@ -3052,6 +3053,12 @@ async fn model_catalog(state: &AppState) -> Result<Vec<Value>, ApiError> {
         .iter_mut()
         .zip(download_contexts.into_iter().zip(download_size_bytes))
     {
+        let fallback_size_bytes = download_context
+            .as_ref()
+            .and_then(|context| context.fallback_size_bytes);
+        let effective_download_size_bytes = download_size_bytes.or(fallback_size_bytes);
+        let download_size_estimated =
+            download_size_bytes.is_none() && fallback_size_bytes.is_some();
         let (downloadable, installed_path, installed) =
             if let Some(download_context) = download_context {
                 let installed_path = state
@@ -3070,16 +3077,20 @@ async fn model_catalog(state: &AppState) -> Result<Vec<Value>, ApiError> {
         object.insert("downloadable".to_owned(), Value::Bool(downloadable));
         object.insert(
             "downloadSizeBytes".to_owned(),
-            download_size_bytes
+            effective_download_size_bytes
                 .map(|value| json!(value))
                 .unwrap_or(Value::Null),
         );
         object.insert(
             "downloadSizeLabel".to_owned(),
-            download_size_bytes
+            effective_download_size_bytes
                 .map(format_bytes)
                 .map(Value::String)
                 .unwrap_or(Value::Null),
+        );
+        object.insert(
+            "downloadSizeEstimated".to_owned(),
+            Value::Bool(download_size_estimated),
         );
         object.insert(
             "installState".to_owned(),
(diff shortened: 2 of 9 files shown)


michaeltrefry (commented):
Tight, well-scoped restoration of the pre-Rust-conversion behavior. Fallback precedence (live HF → download-level manifest → model-level manifest) is sensible, the `downloadSizeEstimated` flag lets the UI mark uncertainty honestly, and the empty state matches the AC's "Unavailable" requirement. Two items worth thinking about.

**Recommendation: comment.**

---

## Notable

### 1. Hardcoded estimates in `builtin.models.jsonc` have no refresh path
Six manually curated byte counts:
- `32899667397` (Z-Image-Turbo)
- `57704594653` (Qwen-Image)
- `57720467613` (Qwen-Image-Edit)
- `157004895813` (LTX-2.3)
- `254107005436` (LTX-Video)
- `34203021834` (Wan2.2)

These are point-in-time snapshots of HF tree sizes. When upstream models change (revisions, weight reformatting, fp8 quantizations land), these go stale silently — the UI keeps showing the old estimate prefixed with `~`. Two ways to mitigate:
- A short comment near the first entry: `// Refresh with: huggingface-cli scan-cache --repo <repo> | grep size` (or whatever your team's process is). Low cost, high signal.
- A CI check that runs `estimate_huggingface_download_size` against the manifest values and warns if they drift by >10%. Overkill for this PR, but worth a follow-up story.

Right now the manifest is the only source of truth for these numbers and there's no breadcrumb pointing to how they were derived.

### 2. Multi-download models show only the first download's size
`model_download_context` returns the first download entry (`download_list.first()` semantics from the existing code), so the LTX-2.3 model — which now has two downloads (`Lightricks/LTX-2.3` at ~146 GB + `Lightricks/LTX-Video` at ~236 GB) — shows ~146 GB to the user.

When they click Download and it pulls ~382 GB, the surprise is on us. Either:
- Sum across all downloads when computing the displayed estimate (`fallback_size_bytes`s combined, or `Σ live_lookup`).
- Show the count alongside the first size: "~146 GB (first of 2 downloads)".

The summing option is the right call for honest disk-space communication.

---

## Smaller observations

- **No frontend test for the `downloadSizeEstimated: false` (live size succeeded) path.** The new test covers `estimated: true` and `no size` cases; the no-tilde case for an exact size isn't pinned. A third entry in the fixture would close the trio.
- **Manifest size won't reflect a `files: [...]` filter.** If a manifest declares `estimatedSizeBytes: 30_000_000_000` for the full repo but `files: ["*.safetensors"]` filters the actual download to half, the user sees an overestimate. The contract-test fixture (`estimatedSizeBytes: 12884901888` + `files: ["*.safetensors"]`) implicitly trusts the manifest author to have accounted for the filter. Worth a one-line schema comment: "`estimatedSizeBytes` should reflect the bytes actually downloaded after the `files` filter, not the full repo size."
- **`downloadSizeEstimated` is always inserted, even for non-downloadable models.** Minor API noise (defaults to `false`). Fine to keep for shape consistency.
- **`Download` button falls back to plain "Download"** when `downloadSizeLabel` is falsy, even though the metadata row shows "Unavailable" — slight asymmetry but reasonable. Including the unavailable hint on the button ("Download (size unknown)") would be one option if you want full consistency.
- **`~` prefix is conventional tech shorthand** for "approximately" — readable. `≈` is more precise but `~` works.
- **`manifest_download_size_bytes` precedence chain** — download-level fields first, then model-level, with the most-specific name (`estimatedSizeBytes`) preferred over the more generic ones (`downloadSizeBytes`, `sizeBytes`). Sensible ordering; could be a one-line comment explaining the precedence so future readers don't have to derive it from the code.
- **`json_size_to_u64`** already handles string inputs (per the existing helper test), which is useful for hand-written manifests where someone might quote the number. Good defensive design.
- **Snapshot test update is one entry** — if other snapshots include model-like entries with `downloads`, they'd want the same field additions and might not be covered. Worth a `grep` for other model snapshots in the fixture before merging.

---

## What's good

- **Fallback precedence is the right shape**: live HF tree-size wins, manifest is the safety net. Lets the live path be authoritative when it works and the manifest provide a baseline when it doesn't (offline, rate-limited, repo gated).
- **`downloadSizeEstimated` boolean** is the honest way to communicate uncertainty: live values aren't prefixed (they're accurate), fallback values get `~`. The UI handles both cases declaratively.
- **"Unavailable" empty state** matches the AC explicitly — "Missing or unknown sizes are handled gracefully" + "show a clear unavailable state rather than silently omitting the field." Both checked.
- **`manifest_download_size_bytes` checks multiple field names** (`estimatedSizeBytes` / `downloadSizeBytes` / `sizeBytes`) — forward-compat for whatever naming convention shakes out across future manifests.
- **All six built-in HF downloads now have estimates** — the empty `files: []` (full repo) downloads consistently get a size, which is the easy case. The contract-test fixture exercises the `files` filter case explicitly.
- **Frontend test covers both estimated and unavailable** in a single render fixture — efficient and asserts the specific strings (`~30.6 GB`, `Unavailable`, `Download ~30.6 GB`) the user will see.
- **`downloadSizeText` helper** centralizes the formatting decision so the metadata row and the button text stay consistent automatically.
- **No download job required for the estimate** — directly satisfies the AC bullet.

---

## Suggested before merge

1. Sum sizes across multiple downloads (or annotate "first of N") so LTX-2.3 doesn't underrepresent the actual ~382 GB combined footprint.
2. Add a refresh-process comment or follow-up story for the hand-curated byte counts.
3. (Optional) Add a no-tilde test case to pin the live-size code path.

michaeltrefry:
Retracting finding #2 from my review — I assumed the multiple `downloads` entries on the LTX-2.3 model represented separate files to be combined, but the author clarified they're **alternate versions of the same model**, only the first is actually used. So showing only the first download's size is correct, not an underestimate.

This also means **don't** sum across the downloads array (that would be the wrong direction). If anything, a future polish would be to document this in the manifest schema — that the `downloads` array represents alternates ranked by preference, with the first being the canonical pick — so this convention is discoverable. But that's well out of scope for this PR.

Net: only finding #1 (refresh path for hand-curated sizes) and the smaller items remain. Sorry for the noise.

PR #49, opened 2026-05-19
Show image batch results incrementally

michaeltrefry:
## Summary
- persist partial image batch results on worker progress updates as each asset is saved
- render Image Studio review slots with completed images plus pending/failed placeholders while a batch is still running
- recover in-progress image jobs for the active project from refreshed job state and support result-sidecar asset URLs before asset listing refreshes

## Tests
- npm test
- npm run build
- python -m pytest tests/test_worker_runtime.py
- Browser smoke: opened Image Studio at http://127.0.0.1:5173 and confirmed an active 4-image job shows Pending #1-#4 slots in the review panel

## Note
- Broader `python -m pytest tests/test_worker_runtime.py tests/test_rust_api_worker_smoke.py` hit an unrelated existing 400 Bad Request in `test_rust_worker_claims_and_completes_lora_import_against_rust_api_binary` while creating a LoRA import job.

Shortcut: sc-1366

### DIFF
--- a/apps/web/src/App.jsx
+++ b/apps/web/src/App.jsx
@@ -40,6 +40,14 @@ function failedJobNotice(job) {
   return `${label}: ${detail}`;
 }
 
+function isImageGenerationJob(job) {
+  return ["image_generate", "image_edit"].includes(job.type);
+}
+
+function hasImageReviewResult(job) {
+  return Boolean(job.result?.generationSetId || job.result?.assetIds?.length || job.result?.assets?.length);
+}
+
 function isLoraImportNotice(message) {
   return String(message ?? "").startsWith("lora import: ");
 }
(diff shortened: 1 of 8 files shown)


michaeltrefry (commented):
## Review

Nice slice: persisting partial results on the worker, hydrating from project state on reload, and using the sidecar `file.path` to render images before the assets list refreshes all hang together cleanly. A few items I'd want addressed before merging.

### Requested changes

**1. `apps/web/src/App.jsx:115-122` — recovery filter retains failed/canceled jobs with partial results indefinitely.**

The PR summary describes recovering "in-progress" jobs, but the filter
```js
(!terminalStatuses.has(job.status) || (job.status !== "completed" && hasImageReviewResult(job)))
```
also pulls in any `failed`/`canceled`/`interrupted` job that has a `result.assets` blob. After refresh, those project-scope jobs flow into `imageLocalJobs`, and in `ImageStudio` the local-job filter
```js
job.status !== "completed" || (!resultVisible(job) && !completedWaitExpired(job))
```
has no clause for failed/canceled — so they stay in the Review stack forever with no dismiss UI. Bounded to 4 by `slice(0, localJobStackLimit)`, but still a behavior change beyond the stated scope. Either:
- restrict recovery to `!terminalStatuses.has(job.status)` (only active jobs), or
- add a TTL/dismiss path for terminal recovered jobs (mirror `completedWaitExpired` for non-completed terminal states).

**2. `apps/web/src/screens/ImageStudio.jsx:28-33` — raw status leaks into the slot label.**

```js
return `${job.status} #${index + 1}`;
```
A failed batch renders as `failed #2`/`failed #3` in the slot label. Use a human-readable string per status (`"Failed"`, `"Canceled"`, `"Interrupted"`) so the placeholder reads naturally. The `.review-placeholder.failed` styling is already there.

**3. `apps/worker/scene_worker/image_adapters.py:227-240` — new progress-callback contract has no Python coverage.**

The contract that `progress(...)` may receive a `result` kwarg containing `assets`, `expectedCount`, `generationSetId`, etc. is the core of this change but isn't asserted anywhere in `tests/test_worker_runtime.py`. The JSX test exercises the consumer side only. A small adapter test (capture progress calls during `write_outputs`, assert the result-bearing saving call shape) would lock the wire contract and prevent silent regressions on the worker side.

- `apps/web/src/screens/ImageStudio.jsx:240-263` — `localJobs` is recomputed every render (it's a plain `.filter` outside `useMemo`). Because it's in the `reviewSlots` dep array, `reviewSlots` memoizes nothing. Wrap `localJobs` in `useMemo` so the slot calc actually caches.
- `apps/web/src/screens/ImageStudio.jsx:23-26` — fallback chain `expectedCount ?? count ?? payload.count` is partially redundant: the worker now always writes `expectedCount`. Trim once you trust the contract.
- `apps/worker/scene_worker/image_adapters.py:38-47` vs `apps/worker/scene_worker/video_adapters.py:36` — `ProgressCallback` is now a Protocol with optional `result` in image, but still `Callable[[str, str, float, str], None]` in video. Fine today, but worth aligning when video gets the same incremental treatment.
- `apps/web/src/components/assetMedia.jsx:8-10` — the fallback URL does not URL-encode path segments. Matches the server side (`crates/sceneworks-core/src/project_store.rs:1969-1985`), and current slugified filenames are safe, but worth a note if filenames ever loosen.

### Verified

- Worker SSE/result contract: `update_job_progress` in `apps/rust-api/src/lib.rs:2726` and `crates/sceneworks-core/src/jobs_store.rs:665` persist `result_json` via `coalesce(?6, result_json)` and republish — incremental updates flow through correctly. `saving` is a valid `JobStatus` (`crates/sceneworks-core/src/contracts.rs:185`).
- Client-side `assetUrl` fallback mirrors server-side `normalize_asset` exactly — projectId + file.path → `/api/v1/projects/{id}/files/{path}`. Path traversal protection is enforced by the Rust `project_file` handler (test at `apps/rust-api/src/lib.rs:7295`).
- Dedup logic in `imageLocalJobs` correctly preserves local-first ordering via Map insertion order.
- `handleJobUpdated` only sets `latestGenerationSetId` and refreshes assets on `status === "completed"` — incremental "saving" updates don't prematurely trigger the post-completion flow.

PR #50, opened 2026-05-19
[codex] Fix queued job cancellation display

michaeltrefry:
## Summary
- Upsert job action responses immediately so queued cancellations render as canceled even if a follow-up list refresh is stale.
- Add an API regression test proving queued job cancellation finishes without worker acknowledgement and cannot be claimed afterward.
- Add a frontend regression test for stale refresh behavior after canceling a queued job.

## Validation
- npm --prefix apps/web test -- main.test.jsx
- cargo test --workspace canceling_queued_job_finishes_without_worker_acknowledgement

## Notes
- `cargo fmt --all -- --check` currently fails before formatting because several Rust files in the checkout report incorrect newline style.
- Leaves the existing unstaged `config/manifests/user.recipe-presets.jsonc` change out of this PR.

Shortcut: SC-1370

### DIFF
--- a/apps/rust-api/src/lib.rs
+++ b/apps/rust-api/src/lib.rs
@@ -5479,6 +5479,75 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn canceling_queued_job_finishes_without_worker_acknowledgement() {
+        let temp_dir = tempfile::tempdir().expect("temp dir creates");
+        let app = create_app(test_settings(&temp_dir)).expect("app creates");
+
+        let (status, created) = request(
+            app.clone(),
+            "POST",
+            "/api/v1/jobs",
+            json!({
+                "type": "image_generate",
+                "projectId": "project-1",
+                "projectName": "Project 1",
+                "payload": { "prompt": "mist over hills" },
+                "requestedGpu": "auto"
+            }),
+        )
+        .await;
+        assert_eq!(status, StatusCode::CREATED);
+
+        let job_id = created["id"].as_str().expect("job id is string");
+        let (status, canceled) = request(
+            app.clone(),
+            "POST",
+            &format!("/api/v1/jobs/{job_id}/cancel"),
+            Value::Null,
+        )
+        .await;
+        assert_eq!(status, StatusCode::OK);
+        assert_eq!(canceled["status"], "canceled");
+        assert_eq!(canceled["stage"], "canceled");
+        assert_eq!(canceled["progress"], 1.0);
+        assert_eq!(canceled["cancelRequested"], true);
+        assert_eq!(canceled["message"], "Canceled before a worker started.");
+        assert!(canceled["canceledAt"].is_string());
+        assert!(canceled["completedAt"].is_string());
+        assert_eq!(canceled["workerId"], Value::Null);
+
+        let (status, queue) = request(app.clone(), "GET", "/api/v1/queue", Value::Null).await;
+        assert_eq!(status, StatusCode::OK);
+        assert_eq!(queue["counts"]["canceled"], 1);
+        assert_eq!(queue["counts"]["queued"], 0);
+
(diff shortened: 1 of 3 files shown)


michaeltrefry (commented):
## Review

Bug is real and the Rust regression test (`canceling_queued_job_finishes_without_worker_acknowledgement`) is exactly the right contract guarantee — both the cancel response shape and the post-cancel claim-returns-null are pinned. Frontend test mirrors the stale-list scenario well. A few concerns about the client fix.

### Requested changes

**1. `apps/web/src/App.jsx:1196-1200` — post-refresh re-apply can clobber fresher SSE state for non-terminal actions.**

Timeline that breaks:
- T0: `await apiFetch(.../{action})` returns `updatedJob` (status=running, progress=0.1)
- T0+10ms: optimistic `setJobs(updatedJob)`
- T0+15ms: `refreshData()` starts
- T0+40ms: SSE `job.updated` arrives with progress=0.4 (worker advanced during the refresh)
- T0+90ms: refresh resolves with stale server list, `setJobs(server list)` (progress=0.4 or 0.1 depending on timing)
- T0+91ms: **second** `setJobs(updatedJob)` re-applies the captured closure value with progress=0.1, clobbering 0.4

For `cancel` of a queued job (the reported bug) the job is terminal so this is fine, but the same `jobAction` handler runs `retry`/`restart`/`duplicate` — those produce mutable jobs the worker can advance during the refresh window. Either:
- gate the re-apply on `updatedJob` being terminal (e.g., `if (terminalStatuses.has(updatedJob.status))`), or
- use a functional updater that only takes `updatedJob` when it has a newer `updatedAt` than the current entry.

**2. Symptom-level fix — consider fixing `refreshData` instead.**

The root cause is `setJobs(jobsResult.value.sort(sortNewest))` at `apps/web/src/App.jsx:350` replaces the array unconditionally. Every caller of `refreshData` after a mutation is exposed to the same staleness window. A merge-based `refreshData` (keep current entry when its `updatedAt` is newer than the server's) would eliminate the need for the local double-set here and protect future call sites for free. Worth at least filing as a follow-up — the current fix lives in one handler and the pattern will be re-discovered the next time someone wires up a mutation.

**3. Brief flicker between optimistic-set → refresh-set → optimistic-set.**

If the refresh response is stale, the user sees: canceled (optimistic) → queued (stale refresh) → canceled (re-apply). Sub-50ms in practice, but visible on slow networks. The merge-in-refresh approach (item 2) avoids this entirely.

- `apps/rust-api/src/lib.rs:5478-5544` — the test could also assert `publish_queue` semantics (e.g., a follow-up `GET /api/v1/queue` returns counts with `canceled: 1, queued: 0`) to cover the queue-summary side of the cancel, which is part of the same handler path.
- `apps/web/src/main.test.jsx:1302-1366` — consider extending the assertion to verify the test would have failed without the post-refresh re-apply (e.g., by also asserting it's NOT `"queued"`). The current `expect(container.textContent).toContain("canceled")` plus the negative `not.toContain("Waiting for an available GPU worker.")` covers most of it — making the test fail-loudly when the fix is reverted is just a small explicit check.
- The PR note flags that `cargo fmt --all -- --check` is failing on unrelated files — worth tracking that separately so it doesn't keep blocking future PRs.

### Verified

- Rust cancel/retry/duplicate handlers (`apps/rust-api/src/lib.rs:2680-2724`) all publish `job.updated` + `publish_queue` before returning, so SSE and HTTP carry consistent state.
- `apiFetch` returns parsed JSON regardless of 2xx code (`apps/web/src/api.js:14-18`); the new `updatedJob` capture works for `duplicate` (201) and `cancel`/`retry` (200) alike.
- The Rust test correctly proves a worker that registers *after* the cancel cannot claim the job — covers the worker-acknowledgement-not-required guarantee in the title.

PR #51, opened 2026-05-19
[codex] Fix preset PEFT LoRA failures

michaeltrefry:
## Summary
- Add PEFT to the Python worker runtime dependencies so Diffusers LoRA methods used by image presets have the required backend.
- Wrap raw PEFT backend failures with SceneWorks-friendly worker messages that identify the preset LoRA path and recovery step.
- Block Image Studio preset submission when a managed preset LoRA is incompatible with the selected model.

## Root Cause
Preset image generation can apply hidden managed LoRAs. Diffusers LoRA loading and adapter selection can require PEFT, but the worker image did not install `peft`, so jobs could fail after queueing with the raw `PEFT backend is required for this method.` exception.

## Validation
- `python -m pytest tests\test_worker_runtime.py -q`
- `npm test -- src/main.test.jsx`
- `cargo test -p sceneworks-rust-api recipe_preset --lib`

Shortcut: https://app.shortcut.com/trefry/story/1373/fix-preset-image-generation-peft-backend-error

### DIFF
--- a/apps/web/src/main.test.jsx
+++ b/apps/web/src/main.test.jsx
@@ -1913,6 +1913,136 @@ describe("SceneWorks app shell", () => {
     );
   });
 
+  it("blocks image presets whose managed LoRAs do not match the selected model", async () => {
+    const createImageJob = vi.fn();
+    root = createRoot(container);
+    await act(async () => {
+      root.render(
+        <ImageStudio
+          activeProject={{ id: "project-1", name: "Noir" }}
+          assets={[]}
+          characters={[]}
+          createImageJob={createImageJob}
+          deleteAsset={() => {}}
+          gpuOptions={["auto"]}
+          imageModels={[
+            { id: "z_image_turbo", name: "Z-Image", type: "image", family: "z-image" },
+            { id: "qwen_image", name: "Qwen Image", type: "image", family: "qwen-image" },
+          ]}
+          latestAssets={[]}
+          loras={[
+            {
+              id: "qwen_detail",
+              name: "Qwen Detail",
+              family: "qwen-image",
+              scope: "builtin",
+              presetManaged: true,
+            },
+          ]}
+          onPreview={() => {}}
+          purgeAsset={() => {}}
+          recipePresets={[
+            {
+              id: "cinematic",
+              name: "Cinematic",
+              workflow: "text_to_image",
+              builtInLoras: [{ id: "qwen_detail", weight: 0.4 }],
+            },
+          ]}
+          requestedGpu="auto"
+          selectedAsset={null}
+          setRequestedGpu={() => {}}
+          updateAssetStatus={() => {}}
+        />,
+      );
+    });
+
+    const generate = [...container.querySelectorAll("button")].find((button) => button.textContent === "Generate");
+    expect(container.textContent).toContain("Preset cannot run with Z-Image");
+    expect(container.textContent).toContain("qwen_detail");
+    expect(generate.disabled).toBe(true);
+
+    await act(async () => {
+      generate.click();
+    });
+
+    expect(createImageJob).not.toHaveBeenCalled();
+
+    await act(async () => {
+      [...container.querySelectorAll("button")].find((button) => button.textContent === "Advanced").click();
+    });
+    await changeField(field(container, "Model"), "qwen_image");
+    await settle();
+
+    expect(container.textContent).not.toContain("Preset cannot run with Qwen Image");
+    expect(generate.disabled).toBe(false);
+
+    await act(async () => {
+      generate.click();
+    });
+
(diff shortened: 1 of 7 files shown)


michaeltrefry (commented):
## Review

Root-cause-style fix on three layers — install the missing backend, give the worker a friendly message when it still surfaces, and block the UI from queueing jobs that are guaranteed to fail. The Python and Rust pieces are clean. A couple of issues on the frontend and a parity gap with Video Studio.

### Requested changes

**1. `apps/web/src/screens/ImageStudio.jsx:158-163` — duplicates `presetValidation()` from `presetUtils.js:113-127`.**

The new inline block:
```js
const presetIncompatibleLoras = presetLoraDetails.filter((detail) => {
  const lora = loras.find((item) => item.id === detail.id);
  return lora && !loraMatchesModel(lora, selectedModel);
});
const presetBlocked = Boolean(presetMissingLoras.length || presetIncompatibleLoras.length);
```
is the same logic as `presetValidation()` (already used by `PresetManagerScreen.jsx:110`). Reuse it so the three surfaces (manager, image, video) can't drift again:
```js
const presetValidationResult = useMemo(
  () => presetValidation(selectedRecipePreset, loras, selectedModel),
  [selectedRecipePreset, loras, selectedModel],
);
```
Then `disabled` checks `!presetValidationResult.ok` and the two `<p className="inline-warning">` blocks read from `presetValidationResult.missing` and `.incompatible`. Lets you drop the local `presetMissingLoras` and `presetIncompatibleLoras` and keep the wording for both cases.

**2. `apps/web/src/screens/VideoStudio.jsx:77, 218, 556-558` — Video Studio still ships the bug for the incompatible case.**

`VideoStudio` blocks only on `presetMissingLoras`:
```js
const presetMissingLoras = presetLoraDetails.filter((lora) => lora.missing);
// ...
const canSubmit = Boolean(... && !presetMissingLoras.length);
```
If a video preset's managed LoRA family doesn't match the selected video model, Video Studio will happily queue the job and the worker will fail it for the same reason. Apply the same guard there (which falls out for free if item 1 is via `presetValidation`). Per project history, Video Studio has repeatedly diverged from Image on these details — worth closing the gap in the same PR.

**3. `apps/worker/scene_worker/lora_adapters.py:166-167` — substring match is fragile.**

```python
def is_peft_backend_error(exc: Exception) -> bool:
    return "peft backend is required" in str(exc).lower()
```
The Diffusers message is currently `"PEFT backend is required for this method."` — if it ever rewords to "PEFT is required" or "Install peft", this guard silently stops catching. At minimum widen the substring set (`"peft backend"`, `"requires peft"`, `"peft is required"`), or — better — also catch `ImportError`/`ModuleNotFoundError` mentioning peft. Today's narrow string is what fires for the exact in-tree bug, so this is hardening, not a correctness bug.

- `apps/worker/scene_worker/lora_adapters.py:117-129` — the nested try inside `set_adapters` block is hard to follow. The inner `except TypeError as exc: if is_peft_backend_error(exc): raise` exists to short-circuit the kwarg-fallback retry; flatten it by checking once after capturing exc or by extracting a small helper. Outer `except Exception` then handles wrap. Same behavior, easier to read.
- `apps/web/src/main.test.jsx:1916-1966` — consider extending the test to assert that switching `imageModels` to a Qwen model unblocks `Generate` (or that switching the preset away does the same). Locks the recovery path.
- `tests/test_worker_runtime.py:315-323` — would be worth also asserting `from exc` chain is preserved (e.g., `pytest.raises(RuntimeError) as info; assert isinstance(info.value.__cause__, ValueError)`) so the original PEFT exception isn't lost from debug logs.
- The wrapped message says "Rebuild or reinstall the SceneWorks worker dependencies with peft" — clearer to name the concrete recovery command (`pip install -r apps/worker/requirements.txt` for bare-metal, `docker compose build worker --no-cache` for compose) since "rebuild" is ambiguous depending on how the user set up.

### Verified

- `peft>=0.17,<1` is a reasonable pin: 0.17 is recent and the `<1` cap is a sane SemVer guard. Adds ~modest size to the worker image; within the project's ~1.2 GB worker image budget.
- Docker pipeline picks up the new dep automatically: `docker/worker.Dockerfile:18-21` installs from `apps/worker/requirements.txt`, no separate manifest to keep in sync.
- Two-layer friendly_failure guard covers both raw (`"peft backend is required"`) and wrapped (`"requires the peft backend"`) substrings, so the message reaches users whether the wrap fires or not.
- Rust side untouched and unaffected: Diffusers/PEFT only run in the Python worker; LoRA loading isn't on the Rust runtime path.
- Frontend test correctly bypasses App and renders `ImageStudio` directly with curated props, so the assertions are about the component's own guard rather than incidental App state.

PR #52, opened 2026-05-19
[codex] Prevent CPU generation jobs

michaeltrefry:
## Summary
- prevent image/video generation jobs from targeting `requestedGpu: cpu`
- ensure auto-dispatch never assigns GPU-required jobs to CPU workers, even if a worker is misregistered
- require a real torch inference backend before Diffusers image/video adapters load, and build the Compose worker with CUDA PyTorch by default
- update Queue eligibility messaging and docs/env examples for the CUDA worker build knobs

## Root Cause
The Python worker image could be built with CPU-only PyTorch while still appearing eligible for GPU generation work. That allowed Z-Image jobs to be claimed and kept alive with progress/heartbeats while no CUDA inference was happening.

## User Impact
After this change, GPU-required jobs stay on real GPU-capable workers. Direct API requests with `requestedGpu: cpu` for image/video generation return a 400 instead of entering the queue.

## Build Note
Compose now defaults the Python inference worker to CUDA PyTorch wheels (`cu128`) so local GPU inference actually uses NVIDIA GPUs. This materially increases worker build/download size compared with the old CPU-only default. Users on slow connections or intentional CPU-only setups can set `SCENEWORKS_PYTORCH_INDEX_URL=https://download.pytorch.org/whl/cpu`, but those workers will not advertise image/video generation capability.

## Validation
- `python -m pytest -q tests\\test_worker_runtime.py tests\\test_worker_gpu.py`
- `cargo test -p sceneworks-core --test jobs_store`
- `cargo test -p sceneworks-rust-api create_job`
- `npm --prefix apps/web test`
- `npm run check`
- `npm run check:compose`

## Notes
`cargo fmt --all -- --check` is currently blocked by existing Rust newline-style issues across multiple files, including files outside this change. I left line endings alone to avoid unrelated churn.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -48,6 +48,12 @@ HF_TOKEN=
 SCENEWORKS_HF_HOME=/sceneworks/data/cache/huggingface
 SCENEWORKS_HUGGINGFACE_HUB_CACHE=/sceneworks/data/cache/huggingface/hub
 
+# Compose worker PyTorch build. The default CUDA wheel lets the Python
+# inference worker actually use NVIDIA GPUs; set the index URL to
+# https://download.pytorch.org/whl/cpu only for an intentional CPU-only build.
+SCENEWORKS_PYTORCH_INDEX_URL=https://download.pytorch.org/whl/cu128
+SCENEWORKS_PYTORCH_SPEC=torch>=2.7,<2.8
+
 # Worker image adapter override: auto, procedural/procedural_preview, z_image_diffusers, or qwen_image.
 SCENEWORKS_IMAGE_ADAPTER=auto
 
--- a/README.md
+++ b/README.md
@@ -30,6 +30,11 @@ container and is exposed on the same host port.
 `SCENEWORKS_WEB_PORT` controls the host port for the Vite web service. The web
 service receives `VITE_API_BASE_URL=http://localhost:${SCENEWORKS_API_PORT}`,
 and workers call `http://api:${SCENEWORKS_API_PORT}` on the compose network.
+Compose builds the Python inference worker with CUDA PyTorch wheels by default
+using `SCENEWORKS_PYTORCH_INDEX_URL=https://download.pytorch.org/whl/cu128`.
+Set `SCENEWORKS_PYTORCH_INDEX_URL=https://download.pytorch.org/whl/cpu` before
+building only when you intentionally want a CPU-only worker; CPU-only PyTorch
+workers do not advertise image or video inference capabilities.
 
 API volume contracts:
 
--- a/apps/rust-api/src/lib.rs
+++ b/apps/rust-api/src/lib.rs
@@ -5178,6 +5178,7 @@ impl From<JobsStoreError> for ApiError {
             JobsStoreError::InvalidNumber(field) => {
                 Self::bad_request(format!("Invalid numeric value for {field}"))
             }
+            JobsStoreError::InvalidRequestedGpu(detail) => Self::bad_request(detail),
             JobsStoreError::RetryLimit { max_attempts } => Self {
                 status: StatusCode::BAD_REQUEST,
                 detail: format!("Job retry limit reached after {max_attempts} attempts."),
--- a/apps/web/src/main.test.jsx
+++ b/apps/web/src/main.test.jsx
@@ -1554,6 +1554,15 @@ describe("SceneWorks app shell", () => {
           setRequestedGpu={() => {}}
           workers={[
             {
+              id: "misregistered-cpu",
+              gpuId: "cpu",
+              gpuName: "CPU worker",
+              status: "idle",
+              currentJobId: null,
+              capabilities: ["placeholder", "cpu", "video_generate"],
+              loadedModels: [],
+            },
+            {
               id: "python-gpu-0",
               gpuId: "0",
               gpuName: "Fixture GPU 0",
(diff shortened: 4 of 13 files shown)


michaeltrefry (commented):
## Review

Four well-layered guards (DB-level rejection, dispatch-level filter, capability advertisement, adapter-level fast-fail) plus matching UI explanation and a Compose-build default that actually wires CUDA up — exactly the right shape for this class of bug. A handful of small issues worth addressing.

### Requested changes

**1. `apps/worker/scene_worker/runtime.py:57, 62-70` — CUDA-only gate forecloses MPS (Epic 1330 Phase 4).**

```python
if "cpu" not in gpu_capabilities and "gpu" in gpu_capabilities and torch_cuda_available():
    capabilities |= set(SUPPORTED_JOB_TYPES)
```
`torch_cuda_available()` returns `False` on Apple Silicon even when `torch.backends.mps.is_available()` is true. Per the Epic 1330 plan, MPS workers come in Phase 4 — this check will need to be widened anyway, and the rename now is cheap. Rename to `torch_inference_backend_available()` (or similar) and OR in the MPS branch:
```python
def torch_inference_backend_available() -> bool:
    try:
        torch = importlib.import_module("torch")
    except Exception:
        return False
    try:
        if bool(torch.cuda.is_available()):
            return True
        mps = getattr(torch.backends, "mps", None)
        return bool(mps and mps.is_available())
    except Exception:
        return False
```
Same change in `require_cuda_for_gpu_worker` — rename to `require_inference_backend_for_gpu_worker` and check `cuda.is_available() or mps.is_available()`. The error wording can stay CUDA-centric since today's GPU workers are NVIDIA-only; the function signature just won't lock Phase 4 out.

**2. `crates/sceneworks-core/src/jobs_store.rs:1373-1380` — case-sensitive "cpu" check leaks past the guard.**

```rust
fn normalize_requested_gpu(value: &str) -> String {
    let trimmed = value.trim();
    if trimmed.is_empty() { "auto".to_owned() } else { trimmed.to_owned() }
}
```
The check at 792 compares `requested_gpu == "cpu"` against the trimmed-but-case-preserving value. `" CPU "` becomes `"CPU"` and slips through. Lowercase the normalized value before the comparison (or use `eq_ignore_ascii_case`). Same concern for `worker.gpu_id == "cpu"` at line 1232 — that depends on whoever registers the worker normalizing. Worth confirming Rust workers register lowercase.

**3. Triple-sourced GPU-required job-type list will drift.**

- `apps/web/src/screens/QueueScreen.jsx:7` — `gpuRequiredJobTypes` set
- `crates/sceneworks-core/src/jobs_store.rs:1382-1391` — `job_requires_gpu`
- `apps/worker/scene_worker/runtime.py:25` — `SUPPORTED_JOB_TYPES` (used as the implicit list)

Three sources, no shared schema. Add a comment on each pointing to the others, or — better — derive the JSX set from a contract emitted by the Rust schema (the project already has a contract-fixture pattern for cross-language types per the existing `packages/schemas/` and `crates/sceneworks-core/src/contracts.rs` setup).

**4. `apps/worker/scene_worker/image_adapters.py:686` — docker command word order is inconsistent with the rest of the codebase.**

```
`docker compose build --no-cache worker`
```
vs `lora_adapters.py:192` and `runtime.py:206` which use `docker compose build worker --no-cache`. Both run, but pick one (recommend the `worker --no-cache` form to match the existing tests that grep for it).

- `apps/web/src/screens/QueueScreen.jsx:25-27` — the new CPU-blocking branch has no UI test confirming the explanation strings still flow correctly (e.g., "Blocked: no active worker supports image generate."). The existing "explains queued GPU jobs..." test setup covers similar terrain; a small extension that registers a CPU worker with `image_generate` cap and asserts the block message would lock the new branch.
- `crates/sceneworks-core/tests/jobs_store.rs:184-219` — solid pair of tests; consider also covering `auto` + `assigned_gpu = "cpu"` directly (i.e., the case where assignment was somehow allowed previously) to pin the worker-side guard symmetrically with the create-side guard.
- Compose default flip from CPU to CU128 is a significant size/bandwidth change (cu128 torch wheel is ~3 GB; image will balloon past the project's ~1.2 GB historical worker baseline). Worth noting explicitly in the PR description and/or CHANGELOG so users running `docker compose build` after pulling don't get surprised on slow connections.
- `apps/worker/scene_worker/image_adapters.py:680-687` — `require_cuda_for_gpu_worker` exempts `gpu_id in {"auto", "cpu"}`. `"auto"` shouldn't reach the adapter path (children spawn with concrete ids), but the defensive exemption could mask a regression where a misconfigured worker passes "auto" through. Either drop the exemption (let it raise) or assert in tests that the parent never spawns a child with `gpu_id="auto"`.

### Verified

- Layered defense is sound: DB rejection at create (`jobs_store.rs:791-797`), dispatch filter at claim (`jobs_store.rs:1232-1234`), capability removal at register (`runtime.py:54-59`), adapter fast-fail at load (`image_adapters.py:680-687`, `video_adapters.py:433`). Each layer is independently testable and the existing test suite exercises three of the four directly.
- Frontend `gpuOptions` already excludes CPU workers via `isSelectableGpuWorker` (`App.jsx:33-35, 179-182`), so the UI never offers `"cpu"` as a `requestedGpu`. Direct-API calls are the only way to hit the new 400; the regression test pins that contract.
- `JobsStoreError::InvalidRequestedGpu → bad_request` mapping is correct (no leaking 500s).
- Old jobs with `requested_gpu="cpu"` for GPU types weren't reachable via UI before this PR (the dropdown already excluded CPU), so no DB cleanup is needed.
- Compose build args (`docker-compose.yml:32-34`) cleanly thread the env vars to Dockerfile, and `.env.example` plus the worker README cover the override path for intentional CPU-only builds.

PR #61, opened 2026-05-19
[codex] Make model and preset selection first-class

michaeltrefry:
## Summary
- Move model and preset selection to the top of Image and Video generation forms.
- Add an explicit `None` preset option for no-preset generation.
- Move GPU and Image Studio LoRA selection into Advanced while preserving existing generation behavior.

## Behavior
Selecting `None` submits with `recipePresetId: null`, skips preset-managed LoRAs and prompt/default effects, and restores preset-applied defaults when they have not been manually changed.

Shortcut: https://app.shortcut.com/trefry/story/1383/make-model-and-preset-selection-first-class-on-generation-pages

## Validation
- `npm test`
- `npm run build`
- Browser sanity check on Image and Video Studio at `http://127.0.0.1:5173`

### DIFF
--- a/apps/web/src/main.test.jsx
+++ b/apps/web/src/main.test.jsx
@@ -2530,6 +2530,14 @@ describe("SceneWorks app shell", () => {
     expect(generate.disabled).toBe(true);
 
     await act(async () => {
+      [...container.querySelectorAll("button")].find((button) => button.textContent === "Hide advanced").click();
+    });
+    await settle();
+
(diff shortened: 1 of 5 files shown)


michaeltrefry:
Addressed the remaining review UX nit: Image Studio now reopens Advanced while an incompatible manually selected LoRA is blocking generation, and the regression test covers collapsing Advanced in that state. Validation: npm test, npm run build.

PR #98, opened 2026-05-21
[sc-1410] Add Rust training dataset store

michaeltrefry:
## Summary

Adds Rust-owned persistence and CRUD APIs for SceneWorks training datasets.

- Adds a `training_store` core module that creates project-managed dataset folders and persists `TrainingDataset` manifests.
- Copies image dataset sources into `training/datasets/<dataset_id>/images` and records stable item/caption metadata.
- Wires project store methods and Axum routes for create/list/get/update/delete under `/api/v1/projects/:project_id/training/datasets`.
- Rejects unsafe item paths, missing assets, non-image inputs, and assets outside the current project.

## Why

Story sc-1410 is the storage/API layer for the native LoRA Training Studio. Rust needs to own dataset files, versions, validation, and project reload behavior before the Train UI and later job planning work can build on it.

## Validation

- `cargo check --workspace`
- `git diff --check`

### DIFF
--- a/apps/rust-api/src/lib.rs
+++ b/apps/rust-api/src/lib.rs
@@ -7146,6 +7146,44 @@ mod tests {
             updated["items"][0]["caption"]["text"],
             "miraStyle close portrait"
         );
+        let dataset_image_path = project_path
+            .join("training")
+            .join("datasets")
+            .join(&dataset_id)
+            .join("images")
+            .join("item_0001.png");
+        assert_eq!(
+            std::fs::read(&dataset_image_path).expect("dataset image remains"),
+            b"png-bytes"
+        );
+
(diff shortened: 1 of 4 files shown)


michaeltrefry:
Addressed the actionable review feedback in 3ad507e. Kept the training_datasets index because it belongs to the 1410 persistence surface, but list_datasets now reads from the index instead of walking manifests and writing during GET. Dataset item replacement now stages into a temp media dir, rejects duplicate item ids before copy, rolls media back if manifest write fails, and cleans orphan create dirs on failure. Validation: cargo test -p sceneworks-core; cargo test -p sceneworks-rust-api; cargo check --workspace; git diff --check.

PR #105, opened 2026-05-21
[codex] Add training config UI

michaeltrefry:
## Summary
- expose the Rust-owned training target registry through the API
- build the Training Studio Configure Job tab from target defaults, including visible fields, Advanced overrides, validation, and a prepared config snapshot
- extend target registry defaults and coverage for config UI values

## Tests
- npm --prefix apps/web test
- npm run web:build
- cargo fmt --all -- --check
- cargo test -p sceneworks-rust-api training_targets_route_returns_builtin_registry
- cargo test -p sceneworks-core --test training_contracts

Shortcut: https://app.shortcut.com/trefry/story/1415

### DIFF
--- a/apps/rust-api/src/lib.rs
+++ b/apps/rust-api/src/lib.rs
@@ -38,7 +38,9 @@ use sceneworks_core::project_store::{
     CharacterReferenceUpdateInput, CharacterUpdateInput, ProjectStore, ProjectStoreError,
     UploadAsset,
 };
-use sceneworks_core::training::TrainingDataset;
+use sceneworks_core::training::{
+    builtin_training_targets, TrainingDataset, TrainingTargetRegistry,
+};
 use sceneworks_core::training_store::{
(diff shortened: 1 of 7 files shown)


michaeltrefry:
Addressed review feedback in commit 5683622.

Changes:
- Prevented Configure Job drafts from resetting when `gpuOptions` is recomputed by worker/queue SSE updates; config defaults now reset only when the selected target or dataset changes.
- Added a separate GPU reconciliation path that only changes `requestedGpu` when the selected GPU disappears.
- Treat empty numeric drafts as invalid/null instead of serializing them as `0`, and validate checkpoint cadence plus other required numeric fields.
- Made the output-name/warning label derive from `outputKind` instead of hardcoding LoRA everywhere.
- Added coverage for the GPU-options recompute regression and validation-blocked prepare path.

Verification:
- `npm --prefix apps/web test` (71 passed)
- `npm run web:build`
- `git diff --check`

PR #380, opened 2026-05-30
fix(character-studio): Z-Image strict pose — fix pose count routing + clean pose-only default; reference img2img-init opt-in (sc-2328)

michaeltrefry:
## Why

The shipped **strict Z-Image pose tier (sc-2257)** locks the *pose* via the Fun-Controlnet-Union skeleton but **dropped the character reference entirely** — identity came from the prompt only ([`image_adapters.py:3120`](../blob/main/apps/worker/scene_worker/image_adapters.py)/`:3128`). So it never met the Character Studio "same character, new pose" requirement. The best-effort tier (sc-2263) was the mirror image: identity via img2img init, but no skeleton.

Investigating "is best-effort redundant now strict shipped?" surfaced that strict didn't preserve identity at all — so they were complementary, not redundant.

## What

mflux's ported `ZImageControl.generate_image` **already** accepts `image_path` / `image_strength` and threads them into the img2img **init** latents, *independently* of the skeleton `control_context`. So one Z-Image pass can do skeleton-accurate pose **and** reference identity. This unifies them — **SceneWorks-only wiring; mflux unchanged** (the pinned branch already has the params):

- **`mlx_flux_runner._run_z_image_control`** forwards spec `imagePath`/`imageStrength` into `generate_image` (`None` → pose-only, the prior sc-2257 behaviour).
- **`MlxZImageAdapter`** strict pose path resolves the reference via `find_asset_media_path` into a **single shared img2img init** across the pose set (identity is constant; only the per-pose skeleton changes). Stops silently dropping it.
- New **`_reference_strength`** helper (default `0.5`, `advanced.referenceStrength` override, clamp `[0.05, 1.0]`). ⚠️ mflux `image_strength` is **inverted** vs diffusers (higher = more init kept = stronger identity, less reposing room), so a dedicated key avoids the footgun.

This **supersedes the best-effort tier sc-2263** (archived — not deferred).

## Tests

475 worker tests pass (3 new):
- ref-init dispatch (strict pose + reference → shared `imagePath` + default strength; no-reference path asserted unchanged)
- `_reference_strength` default / override / clamp / junk-input
- runner pass-through into `generate_image` (mocked mflux), with/without a reference

## Owed (real-Mac E2E)

Sweep `referenceStrength` for the default that holds identity **without** breaking the pose lock — the few-step Turbo init↔control tension. If it can't hold both, sc-2263's best-effort is the documented fallback.

### DIFF
--- a/apps/worker/scene_worker/image_adapters.py
+++ b/apps/worker/scene_worker/image_adapters.py
@@ -7096,9 +7096,14 @@ def _should_route_z_image_to_mlx(payload: dict[str, Any]) -> bool:
       2. Platform == darwin.
       3. Model in MlxZImageAdapter._supported_models.
       4. mode != "edit_image".
-      5. No referenceAssetId (Z-Image has no reference path on either
-         backend; the flag is checked symmetrically with the other mflux
-         families).
+      5. No referenceAssetId — UNLESS the request carries advanced.poses. A
+         plain reference (no poses) has no Z-Image path on either backend, so it
+         falls to torch; but a POSE request must route here regardless of a
+         reference, because the strict Fun-Controlnet-Union pose tier (sc-2257)
+         exists ONLY on MLX (the torch ZImageDiffusersAdapter has no pose
+         ControlNet and would honour count while silently ignoring advanced.poses
+         → "1 of 1" for N poses). MLX consumes the reference as an identity
+         img2img-init (sc-2328) or ignores it (sc-2257).
       6. Sidecar venv exists.
     """
     if os.getenv("SCENEWORKS_DISABLE_MLX_FLUX", "").strip().lower() in {"1", "true", "yes"}:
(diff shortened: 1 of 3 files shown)


michaeltrefry:
Added a second commit fixing the root cause of a bug Michael hit in testing: **Z-Image pose generated only one image ("1 of 1") for N selected poses.**

`_should_route_z_image_to_mlx` had a blanket `referenceAssetId → torch` gate. The Character Studio pose panel always sends a reference (button disabled without one) + `count:1`, relying on the worker to emit one image per pose — but that gate diverted the job to the **torch** `ZImageDiffusersAdapter`, which has **no pose-ControlNet path**, so it honoured `count:1` and silently dropped `advanced.poses`.

The strict Fun-Controlnet-Union pose tier lives **only** on MLX, so pose requests must route there regardless of a reference. Fix: divert to torch on `referenceAssetId` only when there are **no** poses. This is also what makes this PR's identity feature *reachable* — reference+pose jobs were never hitting MLX before, so the new img2img-init code path was effectively dead. Good catch on Michael's part; it surfaced a gap in the original wiring.

michaeltrefry:
**Final state of this PR after testing.** The reference img2img-init identity approach failed in Michael's test (no pose adherence + reference/skeleton bleed): the Fun-Controlnet-Union branch denoises the pose from noise, so a reference init fights the pose lock on few-step Turbo. So it's now **opt-in (`advanced.referenceStrength > 0`), off by default**, and the default Z-Image strict pose tier is **clean pose-only**.

Net deliverables:
1. **Routing fix** — `referenceAssetId` no longer diverts pose requests to the torch adapter (was the "1 of 1 for N poses" bug). ✅ verified by Michael.
2. **Clean pose-only default** for reference+pose jobs.
3. Reference img2img-init retained as an off-by-default experimental knob (per Michael).

Identity for Z-Image pose is **not** achievable via a reference image (no IP-Adapter; img2img can't repose on Turbo). The real path is a **character LoRA** (epic 2221, already threaded through the MLX pose path) — validation deferred per Michael. 476 worker tests pass.

PR #1065, opened 2026-07-02
fix(core): route training_store project.db opens through connect_project_db (sc-8815)

michaeltrefry:
## What

Five sites in `crates/sceneworks-core/src/training_store.rs` opened `project.db` via raw `Connection::open(...)`, inheriting rusqlite's default **0 ms** `busy_timeout`. Under any cross-connection overlap (a worker embedding write racing an API dataset read on the same `project.db`) these paths failed immediately with `database is locked` instead of queueing — intermittent 500s on the training-dataset surface. `resolve_asset_source` additionally skipped migrations and queried `assets` directly.

Fixes [sc-8815](https://app.shortcut.com/trefry/story/8815) ([F-013], Bug).

## Change

- Add a module-private `connect_project_db` helper in `training_store.rs`, matching `character_store`'s **migrating** variant: opens with the shared 5 s `busy_timeout` **and** runs the version-gated comprehensive `apply_project_migrations` (a no-op read once the schema is current).
- Route **all five** sites through it:
  - `resolve_asset_source` (queries `assets` directly → now migrations-first, table guaranteed present)
  - `ensure_training_dataset_table` (drops its now-redundant explicit `apply_project_migrations` call)
  - `list_dataset_summaries_from_index`
  - `index_dataset`
  - `remove_dataset_index`

### Why the migrating variant everywhere

Two `connect_project_db` helpers exist in the crate: `project_store`'s (busy_timeout only) and `character_store`'s (busy_timeout + migrations). `resolve_asset_source` and `list_dataset_summaries_from_index` query tables (`assets`, `training_datasets`) without a preceding `ensure_*` migration call, so they need the migrating variant to guarantee the tables exist. The migration is version-gated, so for sites that already migrated (`ensure_training_dataset_table`, `index_dataset`, `remove_dataset_index`) the second call is a cheap no-op read. One consistent helper avoids a footgun.

## Tests

- New regression test `connect_project_db_migrates_and_queues_on_contention`: asserts the helper migrates a fresh db (`assets` / `training_datasets` tables present) and that a second writer **queues** behind an in-flight write transaction rather than erroring instantly with `SQLITE_BUSY`. (rusqlite exposes no `busy_timeout` getter, so the observable queueing contract is asserted instead.)
- `cargo test -p sceneworks-core` — green (347 lib + integration tests pass).
- `npm run rust:check` — green (fmt, clippy, build).

### DIFF
--- a/crates/sceneworks-core/src/training_store.rs
+++ b/crates/sceneworks-core/src/training_store.rs
@@ -1,5 +1,6 @@
 use std::fs;
 use std::path::{Path, PathBuf};
+use std::time::Duration;
 
 use rusqlite::{params, Connection};
 use serde::{Deserialize, Serialize};
@@ -1266,6 +1267,22 @@ fn resolve_item_source(
     })
 }
 
+/// Open `project.db` for a training-store read/write with the same 5s
+/// `busy_timeout` every other project.db opener uses, so cross-connection
+/// overlap (worker embedding writes racing an API dataset read) queues instead
+/// of failing immediately with `database is locked` (sc-8815). Runs the
+/// version-gated comprehensive migration too — matching `character_store`'s
+/// migrating variant — so the `assets`/`training_datasets` tables are
+/// guaranteed present before we query them (the migration is a no-op read once
+/// the schema is current).
+fn connect_project_db(project_path: &Path) -> ProjectStoreResult<Connection> {
+    fs::create_dir_all(project_path)?;
+    let connection = Connection::open(project_path.join("project.db"))?;
+    connection.busy_timeout(Duration::from_millis(5000))?;
+    apply_project_migrations(&connection)?;
+    Ok(connection)
+}
+
 fn resolve_asset_source(
     project_path: &Path,
     project_id: &str,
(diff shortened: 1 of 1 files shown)


michaeltrefry:
Closing without merge. A focused tie-breaker verification (plus the original re-validation on sc-8815) confirms F-013's premise is empirically false: the pinned rusqlite 0.32.1 (bundled) sets `sqlite3_busy_timeout(db, 5000)` on every connection inside `InnerConnection::open` (inner_connection.rs:119), so the raw `Connection::open` sites already carry the same 5s busy timeout as `connect_project_db` and queue identically under contention (measured: raw open reports PRAGMA busy_timeout=5000; a raw-open writer queued ~369ms behind a BEGIN EXCLUSIVE and succeeded, no SQLITE_BUSY). This PR's `busy_timeout(5000)` is therefore redundant, and its regression test only fails pre-fix on its migration-adds-tables assertion (orthogonal to F-013). The added `apply_project_migrations` on the `resolve_asset_source` read path is a benign-but-unnecessary behavior change. sc-8815 stays closed as invalid/no-action. See sc-8815 for the full evidence trail.

PR #1066, opened 2026-07-02
fix(core): serialize credential-store read-modify-write (sc-8813)

michaeltrefry:
## Summary

`CredentialFileStore::set` and `delete` each did an **unlocked** load → mutate → save of `credentials.json`. Two concurrent calls could interleave and one write would silently clobber the other — a saved credential could vanish after a racing set/delete (**F-011**). This was the one store in `sceneworks-core` that didn't serialize its read-modify-write window (every other store uses the striped `lock_project_files` / `JobsStore.lock`).

**Impact:** a credential could silently disappear after a concurrent set/delete; the user re-enters a token that "didn't stick" and a worker download later fails with an auth error that looks like a token problem.

## Fix

Hold the crate's existing process-wide **striped reentrant lock** (`lock_project_files`, sc-1633) keyed on the credentials **file path** across `load`+`save` in both `set` and `delete`. This matches how `ProjectStore`/`JobsStore` serialize their RMW. Because the lock is path-keyed (not a bare `Mutex<()>`), multiple `CredentialFileStore` instances pointing at the same file also serialize.

`set` and `delete` are the only RMW methods; `list`/`load` are read-only.

## Tests

Two concurrency regression tests added, both of which **fail without the guard** and pass with it:
- `concurrent_sets_do_not_lose_updates` — 8 threads × 25 distinct hosts; asserts all 200 survive with correct tokens.
- `concurrent_set_and_delete_do_not_lose_survivors` — a set thread and a delete thread churn the same hosts while untouched survivors must never vanish.

`cargo test -p sceneworks-core` → 348 pass. `npm run rust:check` (fmt + clippy `-D warnings` + build + tests) → green.

## Scope

Implements **only** F-011's RMW serialization. Sibling findings F-012 (sc-8814, corrupt-file-treated-as-empty) and F-147 (sc-8949, atomic_write fsync) in the same file are separate stories, untouched here.

Story: https://app.shortcut.com/trefry/story/8813

### DIFF
--- a/crates/sceneworks-core/src/credentials.rs
+++ b/crates/sceneworks-core/src/credentials.rs
@@ -15,7 +15,7 @@ use std::path::{Path, PathBuf};
 
 use serde::{Deserialize, Serialize};
 
-use crate::store_util::random_hex;
+use crate::store_util::{lock_project_files, random_hex};
 
 /// Filename of the credential store within the config dir. Shared so the rust-api
 /// (writer) and the worker (reader) never drift.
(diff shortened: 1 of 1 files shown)


michaeltrefry:
Closing as a duplicate of #1064 (same F-011 fix), which merged to main as 9d210cfe. Both PRs independently implemented credential-store RMW serialization; #1064's approach (generalized lock_store_path extraction) was selected and landed. This branch used a different lock wiring but is now redundant and conflicts with main.

PR #1076, opened 2026-07-02
fix(web): keep WebView2 shell painting through native file dialogs (sc-9261)

michaeltrefry:
## Problem (sc-9261)

On **Windows desktop only**, opening a native file dialog from an `<input type="file">` (Models → Import LoRA/model → **Choose**) blanks the whole app and it stays blank on select or cancel. macOS is unaffected.

## Root cause (re-diagnosed on-device)

Reproduced live on the installed 0.5.1 build with instrumentation:

- **#1042's flag is applied but insufficient.** The running WebView2 processes carry `--disable-features=…,CalculateNativeWinOcclusion` (the env var is *merged* into Tauri/wry's args, not ignored), yet the app still fully blanks. So occlusion tracking is not the whole cause.
- **The entire window blanks**, not just `.workspace` — the shell has three independently-promoted composited layers (`.sidebar` and `.workspace` scroll containers, `.topbar` `backdrop-filter`), and they drop together.
- **The blank does not self-recover.** After the dialog closes it stays black; resize and minimize/restore don't fix it. Only an *internally-scheduled* frame (DOM/rAF) forces WebView2 to re-present. That's why the earlier focus-return nudge (#1028) worked at all, and why external window messages don't.

Mechanism: the modal dialog disables + deactivates the webview's owner window; Chromium stops presenting the promoted DirectComposition layers and won't rebuild them from window messages. Windows-only because macOS WKWebView keeps its CoreAnimation layers composited and `NSOpenPanel` doesn't disable the owner window.

## Changes (flag-free, complementary)

1. **Repaint pump during the dialog** (`App.jsx`, desktop-shell-only): while a file picker is pending, pump a sub-pixel transform on `.app` every frame (rAF + timer fallback for when rAF is throttled while occluded) so the webview keeps presenting *during* the dialog — not just after it closes, which was the visibly-flashing stopgap. Armed on a file-input click (capture phase), disarmed on focus/visibility return or `change`, with a 60s safety valve. No-op on web and macOS.
2. **Shed two composited layers** (`styles.css`): drop `backdrop-filter` from `.topbar` and the shared `.modal-backdrop`, compensated with more opaque backgrounds. Fewer promoted layers = less surface for WebView2 to strand. (Also de-risks routing the picker through the shared modal, whose backdrop was itself promoted by the blur.)

The #1042 occlusion flag is intentionally **kept** as harmless defense-in-depth.

## Verification

- `npm run lint` — 0 errors (pre-existing warnings only), `npm test` — 992 passing, `npm run build` — clean.
- ⚠️ **Needs a Windows desktop rebuild to verify on-device** — the bug is WebView2-specific and does not reproduce on web or macOS.

## Note on overlap

Supersedes the approach on `claude/sc-9261-restore-upload`, which reverts #1042 and adds a focus-return-only nudge (the stopgap that flashes blank *during* the dialog). This PR keeps painting throughout instead.

### DIFF
--- a/apps/web/src/App.jsx
+++ b/apps/web/src/App.jsx
@@ -1052,6 +1052,112 @@ export function App() {
       .catch(() => setSetupCompleted(true));
   }, []);
 
+  // Desktop (WebView2 / Windows) only: opening a native file dialog from an
+  // `<input type="file">` can strand the composited app shell painted blank.
+  // When the modal dialog takes the foreground it disables the owner window and
+  // Chromium stops presenting the promoted DirectComposition layers (`.sidebar`,
+  // `.topbar`, `.workspace`); it does NOT restore them from ordinary window
+  // messages (resize / minimize-restore) — only an internally-scheduled frame
+  // recovers it, which is why disabling `CalculateNativeWinOcclusion` alone
+  // doesn't help. So while a file picker is pending we pump a cheap, sub-pixel
+  // transform on `.app` every frame to keep WebView2 presenting *during* the
+  // dialog (not just after it closes, which was the visibly-flashing stopgap),
+  // then do a final promote/demote once focus returns to guarantee recovery.
+  // No-op off the desktop shell and on macOS WebKit, which never drops layers.
+  useEffect(() => {
+    if (!isDesktopShell) {
+      return undefined;
+    }
+
+    let pumping = false;
+    let rafId = 0;
+    let timerId = 0;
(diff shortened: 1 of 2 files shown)


michaeltrefry:
Validated on-device: rebuilt the Windows desktop bundle and confirmed the app no longer blanks when the native LoRA/model file dialog opens (select or cancel). Moving the story to In Review.

PR #1194, opened 2026-07-04
fix(rust-api): quant tier install state must be per-tier, not repo-marker (sc-9909)

michaeltrefry:
## Problem
Follow-up to #1193 (sc-9907). After that fix removed the false *"incomplete + Fix"*, the Model Manager showed **all three tiers (q4/q8/bf16) as installed** for quant-matrix models even when only **one** tier was downloaded.

Verified live against the running app (`GET /api/v1/models`): e.g. `flux2_klein_9b` has only `q4` in the HF cache, yet `q8` and `bf16` reported `installed/complete` with `installedPath = data/models/SceneWorks__flux2-klein-9b-mlx`.

## Root cause (`apps/rust-api/src/models.rs`, `model_variant_states`)
Per-tier `installed` was `cache_health.installed || managed_installed`, where `managed_installed = model_is_installed(&managed_path)` checks the **repo-level** app-managed dir `data/models/<repo>`. The download flow writes a repo-level completion marker `.sceneworks-download-complete.json` there — **one marker per repo**, tier-agnostic (it records only repo/modelId/jobId/completedAt). So once any tier is downloaded, the marker exists and every tier OR's to `installed = true`.

sc-9907's *absent-tier → missing* change is what made this fully visible: those tiers flipped from `cacheState: incomplete` to `complete`.

## Fix
New `managed_tier_installed(managed_path, files)`: the managed dir only certifies a tier when the marker is present **AND** — for a tier with a `files` filter — the tier's own files actually exist under the managed dir. A quant tier whose files live only in the HF cache (or nowhere) no longer inherits the repo marker. Single-variant turnkey (empty `files`) is still certified by the marker alone, preserving the pre-matrix contract.

## Tests
`quant_matrix_model_with_single_tier_reads_installed_not_incomplete` now also writes the repo-level managed marker (the realistic on-disk state) and asserts the un-downloaded tiers stay `missing`/`missing` while the downloaded tier reads `installed`/`complete`. **Fails without this fix.**

Full `sceneworks-rust-api` lib suite green (187 passed); fmt + clippy + build clean.

## Known follow-up (separate, filed)
The coarse `<tier>/*` glob still marks a tier installed when it holds ≥1 file, so a *partially* downloaded tier (e.g. z-image-turbo q8 present but missing `transformer/` + `vae/`) reads installed. Tracked as **[sc-9910](https://app.shortcut.com/trefry/story/9910)** — needs per-tier required-file patterns in the manifest.

Shortcut: [sc-9909](https://app.shortcut.com/trefry/story/9909) (epic 8506).

### DIFF
--- a/apps/rust-api/src/models.rs
+++ b/apps/rust-api/src/models.rs
@@ -1131,10 +1131,14 @@ fn model_variant_states(model: &Value, data_dir: &FsPath) -> Vec<ModelVariantSta
                 .as_ref()
                 .map(|health| health.missing_files.clone())
                 .unwrap_or_default();
-            // The managed dir mirrors the default-download install path; a variant that is present
-            // there (a directly-downloaded turnkey) counts as installed too.
+            // The managed dir mirrors the default-download install path; a variant present there (a
+            // directly-downloaded turnkey) counts as installed too — but the check must be TIER-aware.
+            // A quant-matrix repo writes ONE repo-level completion marker no matter which tier was
+            // fetched, so keying a per-tier "installed" on the bare marker made EVERY tier report
+            // installed after any single tier's download (sc-9909). Require the tier's own files to
+            // actually exist under the managed dir, not just the marker.
             let managed_path = data_dir.join("models").join(safe_download_dir(&repo));
-            let managed_installed = model_is_installed(&managed_path);
+            let managed_installed = managed_tier_installed(&managed_path, &files);
             let installed_path = if installed || cache_incomplete {
                 cache_path
             } else if managed_installed {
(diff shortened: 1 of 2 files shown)


michaeltrefry:
Added commit 04616dc8 (same sc-9909 theme). While diagnosing a follow-up report, found the underlying cause is a **content gap**: the `SceneWorks/*-mlx` quant repos are only partially uploaded (sc-8513 rollout). Verified live on HF — e.g. sdxl-base-mlx has bf16 (6.9GB) but q4=configs-only and q8=empty. Installing an unpublished tier resolves zero files → empty cache skeleton + stale marker, which read as `incomplete: snapshots/<revision>` + phantom-installed.

This commit adds two accuracy fixes so unpublished/empty tiers show a clean **not-installed** state instead:
- `huggingface_cache_health`: empty repo skeleton (no snapshots) → `missing`, not repairable `incomplete`.
- `install_state_for`: matrix top-level derives from the tier-aware variant aggregate, no longer trusting the repo-level marker.

Test: `quant_matrix_empty_cache_skeleton_reads_missing_not_incomplete`. 188 green; fmt+clippy clean.

Separately filing a worker fix (a download resolving zero files should fail, not report success), and the content rollout is sc-8513.

PR #1492, opened 2026-07-13
chore(catalog): rename Krea 2 Turbo edit hint scene/person → image 1/image 2 (sc-11816)

michaeltrefry:
## What

Fixes the last straggler of the sc-11797 rename. The Krea 2 **Turbo** edit second-image slot still showed the "scene/person" misnomer in the UI:

> **Person image** — "Optional — a person to place into the scene. Fixed order: the Source image is the scene (image 1), this is the person (image 2)."

sc-11797's scope covered the worker + candle-gen engine and assumed the UI was already correct, but it **missed `config/manifests/builtin.models.jsonc`**. The `krea_2_raw` `editReferences` block had already been corrected; `krea_2_turbo` was left behind, so the two Krea edit surfaces were inconsistent.

## Change

Mirror the already-correct `krea_2_raw` wording into the `krea_2_turbo` block — `secondaryLabel` → `"Image 2 (optional)"`, the neutral "combine with Image 1" hint, and the adjacent comment — keeping the Turbo-specific bits (`sc-11640`, `krea_2_turbo_edit`). The two blocks are now identical wording.

No behavior change — the fixed image-1-then-image-2 trained order is preserved. Other manifest `scene`/`person` hits (multi-person scenes, person replacement, scene-flexible resemblance) are legitimate and untouched.

## Verification

- `node scripts/check-scaffold.mjs` → **green**; manifest JSONC parses.
- No `scene (image …`/`person (image …`/`Person image`/`place into the scene` left in the manifest.
- Turbo and Raw edit slots now render identical neutral wording.

Reported by @michaeltrefry1818 (Turbo edit still showing the old hint). Relates to sc-11797.

### DIFF
--- a/config/manifests/builtin.models.jsonc
+++ b/config/manifests/builtin.models.jsonc
@@ -2546,13 +2546,14 @@
         "img2img": true,
         "img2imgStrength": { "label": "Reference strength", "default": 0.5, "min": 0.0, "max": 1.0, "step": 0.05 },
         // Two-reference Kontext edit (sc-11640, mirrors Krea 2 Raw): the `krea2_identity_edit` edit
-        // optionally takes a SECOND source — scene = image 1 (the required Source image), person = image 2
-        // (optional) — in a FIXED order. Presence of this object makes the Studio render the optional
-        // person slot in Edit mode; the worker + engine (`krea_2_turbo_edit` MultiReference) cap at two and
-        // preserve order. Ignored outside `edit_image` (t2i/img2img).
+        // optionally takes a SECOND source — any two images, image 1 (the required Source image) + image 2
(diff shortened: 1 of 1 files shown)


michaeltrefry:
Added a second commit expanding the hint to say users can reference each subject **by position or by description**.

This is backed by a GPU A/B on the Krea 2 edit path (two synthetic subject crops, same seed, only the instruction's reference swapped):
- "red cap on **the person in image 2**" → cap on subject 2
- "red cap on **the person in image 1**" → cap on subject 1
- appearance phrasing ("the bald bearded man" / "the red-haired woman") targeted identically.

The index reliably steers which reference is edited (both directions, everything else held constant), so ordinal phrasing is valid — the hint now surfaces both options.

PR #1547, opened 2026-07-15
feat(worker): NVFP4 as a distinct, Blackwell-gated quant tier (sc-11042)

michaeltrefry:
Worker half of **sc-11042** (epic 11037). The runtime half shipped in `runtime-2026.07.3` (sc-12006); this makes the tier actually reachable.

**Decision: Option A** — NVFP4 is its own user-selectable tier, **never an auto-swap of q4 on Blackwell**. NVFP4's numerics differ from int4-affine q4 (E2M1 elements + FP8-E4M3 block scales, W4A4 regime), so auto-selecting it for a q4 pick would silently change what that tier renders — the creative-choice violation **SC#5** forbids.

## Two required fixes (found by the sc-12006 review, [activity-12150](https://app.shortcut.com/trefry/story/11042#activity-12150))

Both were unreachable until now — nothing produced a `Quant::Nvfp4`. This story's tier-select is exactly what arms them.

1. **`video_jobs.rs` `video_quant_label` aliased NVFP4 → `"q4"`.** It derived the label via `format!("q{bits}")` from `q.bits()`, and `Quant::Nvfp4.bits()` returns **4** — so an NVFP4 video render was stamped `"q4"` + bits 4 in Stats telemetry. **The compiler could not catch it** (reading `.bits()` raises no E0004 on a new variant), so it compiled silently on the `backend-candle` lane. Now matched per-variant → a future `Quant` variant is a hard compile error here instead of a silent mislabel. NVFP4 reports **no** bit count (~4.5 *effective* bits/weight); `Some(4)` would repeat the aliasing in the numeric column.
   - **Audit:** repo-wide, this was the **only** `.bits()`-derived tier label (one hit across all crates).
2. **`flux2_comfyui_candle.rs` stale `quantTier` key.** The insert was conditional, so a caller-supplied `advanced.quantTier` survived on the q4/q8 path (`raw` is a clone of `request.advanced`). Now `else { raw.remove("quantTier"); }` — the symmetric twin of the deliberately-unconditional `mlxQuantize` insert. This matters more now that `quantTier` is a *selection* input: a labeled request that resolved to q4/q8 anyway (off-Blackwell, or tier not converted) must not record the tier it did not run. **q4/q8 record shape stays byte-identical.**

## Tier-select

NVFP4 rides **`advanced.quantTier: "nvfp4"`** — the label sc-12006 established as write-only telemetry, now *consumed* — and **not** `mlxQuantize`, which is bits-valued and names a tier (`<=0`⇒bf16, `<=4`⇒q4, else q8): no integer is honest for NVFP4, so `mlxQuantize` stays **null** for it. Same shape as the sc-9300 `convRot` precedent.

Selection requires **both** an explicit label **and** an sm_120 host. The Blackwell gate is defence-in-depth, mirroring ConvRot (capability + engine floor):
- **Picker-side:** the worker advertises a new `nvfp4` capability only on the candle lane with compute cap ≥ **12.0**; the web hides the tier otherwise.
- **Worker-side:** `nvfp4_host_eligible()` re-checks the probed cap at tier-select, because `advanced` is free-form pass-through — a hand-crafted API call can put the label on a request to any worker.

The 12.0 floor deliberately excludes **datacenter Blackwell sm_100** (probes 10.0): it *has* FP4 hardware but is explicitly out of scope and the lane isn't built (`CUDA_COMPUTE_CAP=120`) or validated for it.

**Fallback is clean, never an error:** off-Blackwell, or before sc-11043's converter produces an `nvfp4/` dir (no shipping model packs one today), the pick rejoins the existing clean-tier chain. That fallback isn't silent — `reconcile_resolved_tier_quant` still fires `quant_tier_downgraded`.

Wired into `standard_tier_subdir` and `krea_model_subdir` (epic 11037's named SC#1/SC#2 vehicle per the sc-12110 Sana→Krea redirect; its subdir shape takes an `nvfp4/` dir with no new logic). The three MLX-family resolvers (anima/ideogram/boogu) pass `false` with a documented rationale.

## SC#5 guard

Structural, not just review:
- `preferred_tier`'s bits map is **unchanged**, and NVFP4 is proven **unreachable** from it for every `(bits, floor)` pair.
- The resolvers are pinned to their **pre-sc-11042 answers on a Blackwell-eligible host with an `nvfp4/` dir present** — hardware selects nothing; only a deliberate pick does.
- Web: `nvfp4` is excluded from `GENERATION_QUALITY_TIERS`, so it can never become anyone's default; a q4 sticky stays q4.

## SC#6 — measured, not claimed

**Not asserted here, and not this story's to fix.** sc-11045 measured resident VRAM on real Sana/sm_120: blanket W4A4 **0.2822×** ✅, but the **shipping mixed policy 0.6988×** ⚠️ (95/163 layers hold dense bf16), blanket W4A16 1.0000×. Root cause is `Nvfp4Linear::new_dequant` materializing a resident dense bf16 weight — an **inference-repo** defect tracked as **sc-12121**, needing its own fix + a new runtime tag. No NVFP4 weights exist on disk yet (sc-11043), so the tier's footprint is **not observable from the worker today**.

## Verification

- `scripts/check-gen-core-skew.sh` green — exactly one `sceneworks-gen-core` at `runtime-2026.07.3#8c35cd2a`
- `cargo build --locked --features backend-candle` (vcvars64/MSVC 14.44, `CUDA_COMPUTE_CAP=120`) — **0 errors / 0 warnings**
- `cargo clippy --locked --features backend-candle -p sceneworks-worker --all-targets -- -D warnings` — clean
- **Neither build** (`npm run rust:check:neither`) — clean (`nvfp4_host_eligible` has both cfg arms; the CONTRIBUTING cfg trap)
- `cargo fmt --all -- --check` — clean
- **612** worker lib tests pass (**603 before + 9 new**, 0 failed)
- **1814** web tests pass (135 files)

New tests: NVFP4 resolves on sm_120; falls back cleanly off-Blackwell and when unconverted; **existing q4/q8/bf16 selection unchanged on both host classes** (the SC#5 regression guard); `video_quant_label(Nvfp4) != "q4"`; the sm_120 cap floor incl. sm_100 exclusion; label parsing rejects every `mlxQuantize` value; web payload/gating/default-seeding.

### DIFF
--- a/apps/web/src/imageJobAdvanced.js
+++ b/apps/web/src/imageJobAdvanced.js
@@ -1,5 +1,5 @@
 import { buildStructuredPromptRecipe } from "./ideogramCaption.js";
-import { tierQuantize, isConvRotTier } from "./quantTier.js";
+import { tierQuantize, isConvRotTier, isNvfp4Tier, NVFP4_TIER } from "./quantTier.js";
 
 // sc-8854 (F-052): pure builder for the Image Studio job's `advanced` payload. Extracted
 // verbatim from ImageStudio.submit() — the ~110-line object literal that assembled the
@@ -158,6 +158,18 @@ export function buildImageJobAdvanced(state) {
     // (weights = Dir(bf16 snapshot) + text_encoder = File(convrot DiT)). Emitted only when the picker
     // is shown AND the picked tier is int8-convrot — disjoint from the mlxQuantize spread above.
     ...(showTierPicker && isConvRotTier(quantTier) ? { convRot: true } : {}),
+    // NVFP4 tier (sc-11042, epic 11037): like int8-convrot, NVFP4 is NOT a bits-based quant — E2M1
+    // elements + FP8-E4M3 block scales, ~4.5 effective bits/weight — so it can't ride `mlxQuantize`
+    // (its `tierQuantize` is null, so the spread above omits it and mlxQuantize stays out of the
+    // payload entirely). Instead send the distinct `quantTier: "nvfp4"` label the worker's tier-select
+    // reads (`nvfp4_requested` → the `nvfp4/` tier subdir + `Quant::Nvfp4`). sc-12006 already
+    // established this exact key/value as the tier's identity on the asset record, so the signal the
+    // studio sends and the telemetry the worker stores are the same string.
+    //
+    // Emitted ONLY when the picker is shown AND the user explicitly picked nvfp4 — disjoint from both
+    // spreads above. This is the SC#5 opt-in at the payload layer: absent this deliberate pick no
+    // request ever carries the label, so no q4 render can be silently upgraded to NVFP4 on Blackwell.
+    ...(showTierPicker && isNvfp4Tier(quantTier) ? { quantTier: NVFP4_TIER } : {}),
     // PiD decoder (epic 7840, sc-7851): emit usePid:true only when the toggle is shown
     // (model PiD-eligible AND checkpoint installed) AND on. The worker swaps the native
     // VAE for the PiD decode + 2K/4K super-resolve pass; it rides `advanced` (opaque
--- a/apps/web/src/imageJobAdvanced.test.js
+++ b/apps/web/src/imageJobAdvanced.test.js
@@ -223,6 +223,35 @@ describe("buildImageJobAdvanced", () => {
     expect(buildImageJobAdvanced(offState({ showTierPicker: true, quantTier: "q4" }))).not.toHaveProperty("convRot");
   });
 
+  it("emits quantTier (not mlxQuantize) for the NVFP4 tier (sc-11042)", () => {
+    const advanced = buildImageJobAdvanced(offState({ showTierPicker: true, quantTier: "nvfp4" }));
+    expect(advanced.quantTier).toBe("nvfp4");
+    // NVFP4 is not a bits-based quant (~4.5 effective bits/weight), so it must NOT leak an
+    // mlxQuantize — `4` there would select the int4-affine q4 tier (epic 11037 SC#5 aliasing).
+    expect(advanced).not.toHaveProperty("mlxQuantize");
+    // …and it is a different tier from int8-convrot, so it never sends that signal either.
+    expect(advanced).not.toHaveProperty("convRot");
+    // quantTier is only emitted when the tier picker is shown AND nvfp4 is picked.
+    expect(
+      buildImageJobAdvanced(offState({ showTierPicker: false, quantTier: "nvfp4" })),
+    ).not.toHaveProperty("quantTier");
+  });
+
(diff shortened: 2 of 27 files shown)


michaeltrefry:
## Audit follow-up: a **third** aliasing site — `effective_quant_label` stamped NVFP4 as `"bf16"`

Pushed in `2d53401d`. The story asked me to *"audit for any other `.bits()`-derived tier label/map on a candle path"*. The literal `.bits()` grep returned exactly **one** hit repo-wide (`video_quant_label`, fixed above) — but the audit shouldn't stop at the literal call, because the bug is the *class*, not the spelling.

**`image_jobs/base.rs` `effective_quant_label`** is the image lane's twin of `video_quant_label`, and it aliases too — it just derives the label from `resolve_quant`'s **bit count** rather than from `.bits()` directly, so no grep for `.bits()` could find it:

```rust
match resolve_quant(request).1 {   // <- the BITS, not the variant
    Some(8) => ("q8", Some(8)),
    Some(4) => ("q4", Some(4)),
    _       => ("bf16", None),     // <- NVFP4's bits are None → lands HERE
}
```

NVFP4's bit count is deliberately `None` (~4.5 *effective* bits/weight — `Some(4)` would alias it onto q4, which is the whole reason the video fix reports `None`). So the bits-only match dropped NVFP4 into the `_` arm and stamped a **4-bit NVFP4 render as full-precision `"bf16"`**.

This is the **inverse** of the video lane's mislabel — `bf16` instead of `q4` — but the identical SC#5 violation: the stored creative choice is not the one that ran. It's also a nice illustration of why "match the variant, never derive a tier label from bits" is the right rule rather than "don't call `.bits()`": the two sites fail in opposite directions from the same root cause, and *fixing* the video one honestly (bits → `None`) is what armed this one.

Fixed the same way — match the variant, with the NVFP4 arm alongside `int8-convrot`'s (both are tiers with no honest integer width, both report `None` bits for that reason). Added `effective_quant_label_gated` mirroring the `resolve_quant_gated` seam so the mapping is testable on both host classes — necessary because **the candle CI lane runs on the sm_120 rig**, so a test probing the live cap would assert different things there than on a dev box.

**Re-verified:** 613 worker lib tests pass (603 baseline + 10 new, 0 failed) · clippy `-D warnings` clean · neither build clean · fmt clean.

**Audit now complete** — three sites total, all matched per-variant:
| site | lane | was | now |
|---|---|---|---|
| `video_quant_label` | video | `"q4"` (bits-derived) | `"nvfp4"`, no bits |
| `effective_quant_label` | image | `"bf16"` (bits-derived) | `"nvfp4"`, no bits |
| `flux2_comfyui_raw_settings` | image (comfy) | stale key survived | `else { remove }` |

Issue #1544, opened 2026-07-15
problems with new version


Pall13:
Hello,

I am experiencing an issue with SceneWorks version 0.7.5.

I have an iMac with an Apple M3 chip and 8 GB of unified memory. With SceneWorks version 0.7.3 everything was working correctly and I was able to generate images without any problems.

After updating to version 0.7.5, image generation no longer works. The MLX worker fails to start and I receive this error:

```
MLX error: [metal::set_wired_limit] Setting a wired limit larger than the maximum working set size is not allowed.
```

The SceneWorks log shows:

```
MLX worker unavailable.
The MLX GPU worker is required on macOS but no live worker claimed this job.
```

I checked my MLX installation and it seems to be working correctly:

* MLX version: 0.32.0
* GPU detected correctly: Apple M3
* MLX test from Python returns: Device(gpu, 0)

I also tried changing the SceneWorks GPU memory setting. I tested 7 GB, 6 GB, 5 GB, and 4 GB, but the error remains the same.

Since the only change was updating from 0.7.3 to 0.7.5, I suspect there may be a regression in the new version, possibly related to MLX memory handling on Macs with 8 GB unified memory.

Could you please check if this is a known issue or if there is a workaround?

I can provide additional logs or system information if needed.

Thank you!

michaeltrefry:
Can you tell me what model(s) were working for you? I didn't expect anything to work on 8GB

Thank you for filing this issue, you did find a real regression

Pall13:
z-image turbo works fine with 0.7.3 version, new version got errors

michaeltrefry:
OK, I'm pretty sure this will be fixed in the next release. Look for 0.7.6 this evening. If you still have problems, don't be afraid to open another issue. Thank you!

Pall13:
thanks for all, sure I let you know if its ok with new version.

PR #1552, opened 2026-07-15
fix(worker): restore image generation on 8/16 GB Macs (sc-12178, sc-12179) (#1544)

michaeltrefry:
Fixes the 0.7.5 regression in GitHub #1544: an 8 GB iMac M3 that generated images fine on 0.7.3 can no longer generate on 0.7.5. Two **independent** worker-only bugs both hit small-RAM Macs; both are fixed here (one commit each).

## Bug 1 — `set_wired_limit` crashes the worker at startup (sc-12178, epic 7819)
`set_gpu_memory_limit_inner` passes the raw user GPU cap to `mlx_rs::memory::set_wired_limit`, which **throws** when the cap exceeds the device `recommendedMaxWorkingSetSize`. MLX's default error handler answers that throw with `exit(-1)` — an uncatchable libc exit (not a Rust panic `catch_unwind` could contain) that kills the worker before it claims a job. That is the exact error the user pasted:
```
MLX error: [metal::set_wired_limit] Setting a wired limit larger than the maximum working set size is not allowed.
```
On an 8 GB Mac the ceiling is ~5.3 GB, so a 6–7 GB cap (the natural "leave RAM for the OS" choice) crashed the worker → "no live worker claimed this job".

**Fix:** derive the device ceiling from MLX's default memory limit (documented as 1.5× the recommended working set), read once before the first `set_memory_limit`, and clamp the wired limit to it. A cap above the ceiling is meaningless anyway — the device already bounds wired residency there.

## Bug 2 — the fit-gate categorically rejects small-RAM Macs (sc-12179, epic 10834)
The 0.7.5 fit-gate predicts `peak = Σweights + HEADROOM_GB(18)` and rejects when it exceeds **total unified memory**. On an 8 GB Mac that is `> 8` for *every* measurable model → every model rejected; 16 GB Macs are hit for most models too (the most common Mac config). The flat 18 GB headroom bundles a worst-case 1024² activation transient that is **pageable** (swaps, not a wired SIGKILL) and resolution-bound — the wrong thing to hard-reject on. Lowering the GPU-cap slider can't help: the gate budgets on real `hw.memsize`, not the cap (matches the user's report that 7/6/5/4 GB all failed).

**Fix:** a weights-fit floor — a would-be reject is admitted (Sequential where the provider stages, else best-effort Resident) whenever the resident weights fit `budget − OS_RESERVE_GB` (3 GiB). Reject only when the weights themselves can't be held. Restores small-Mac generation and, as a bonus, lets the epic's own qwen-image-on-32 GB unlock finally fire (the flat headroom had defeated it: staged peak 38 > 32).

### ⚠️ Trade-off to review (sc-11924)
The floor is strictly **more permissive**, so a transient-heavy bf16 tier (lens-turbo class) on a *mid-size* Mac can now reach a Metal-OOM where it previously pre-rejected. That is the pre-fit-gate behavior and IMO the correct default — a machine that ran a model must not be newly wall-rejected; it should degrade (page / Sequential), not categorically exclude. The honest fix for those tiers is in-memory (materialized) component sizing, already tracked as **sc-11924**. Flagging explicitly for your call.

## Verification
- `mlx_fit_gate` + `generator_cache` unit tests green (new: `weights_fit_floor_admits_small_model_on_a_small_mac`, `clamp_wired_limit_never_exceeds_the_device_ceiling`; the peak-selection test is preserved as `decide_residency_by_peak_...`).
- Full `npm run rust:check` (fmt + clippy `-D warnings` + whole-workspace test + build) — exit 0.
- **On-device probe** (ignored test, ran on this 128 GiB Mac): `get_memory_limit() = 130.5 GB` = exactly 1.5× the derived ceiling `87.0 GB` (63% of RAM), and `set_wired_limit(ceiling)` does not throw. By the same 63% ratio an 8 GB Mac's ceiling is ~5.3 GB — confirming the crash and the clamp.

Not verified end-to-end on real 8 GB hardware (I don't have one); the emulated-budget path (`SCENEWORKS_MLX_MEMORY_CAP_GB`) is exercised by the unit tests, and on-device real-weights A/B for the fan-out families is existing verification debt (sc-11922).

Both changes are worker-only — no gen-core / mlx-gen / candle-gen pin lockstep.

### DIFF
--- a/crates/sceneworks-worker/src/generator_cache.rs
+++ b/crates/sceneworks-worker/src/generator_cache.rs
@@ -171,7 +171,11 @@ fn generator_cache_idle_timeout(raw: Option<&str>) -> Option<Duration> {
 ///   working set genuinely needs more will thrash/swap or hit a Metal OOM — already contained by the
 ///   `catch_unwind` guard above).
 /// - `set_wired_limit` — caps pinned (non-pageable) residency so the OS can reclaim the rest of
-///   unified memory for other apps. macOS 15+.
+///   unified memory for other apps. macOS 15+. **Clamped to the device wired ceiling** — MLX throws
+///   if asked for more than the device `recommendedMaxWorkingSetSize`, and its default error handler
+///   answers that throw with `exit(-1)`, killing the worker at startup (sc-12178, GitHub #1544: an
+///   8 GB Mac's ceiling is ~5.3 GB, so a 6–7 GB user cap crashed the worker). See
+///   [`clamp_wired_limit`].
 ///
 /// We deliberately leave `set_cache_limit` at its default: forcing it low causes reallocation storms
 /// between steps (the fork's own doc warns about this).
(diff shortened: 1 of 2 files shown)


michaeltrefry:
## Revised per review feedback (gate was too conservative) — force-pushed

**1. z-image-turbo is the 0.7.3 baseline → must run on 8 GB.** Measured the real tiers on disk:

| tier | total | largest single component (wired under Sequential) |
|---|---|---|
| **q4** | 5.49 GiB | **3.38 GiB** (transformer) |
| bf16 | 19.11 GiB | 11.62 GiB |

On 8 GB the capability-clamp picks q4. Lowered `OS_RESERVE_GB` 3 → **2 GiB** (flat, not a fraction) → 6 GiB weight budget on 8 GB → q4's 3.38 GiB admits **Sequential** with ~2.6 GiB margin. Locked in with two tests, incl. an ignored one that drives the **live `residency_for_dir` seam** against the real q4 weights under `SCENEWORKS_MLX_MEMORY_CAP_GB=8`:
```
live gate on …/z-image-turbo-mlx/…/q4 @ 8 GB → Sequential
```

**2. "No reason to hold 40 GB in reserve on a 128 GB Mac."** Agreed, and it doesn't. The ~63% figure (81 GiB) is your Metal device's `recommendedMaxWorkingSetSize` (you're on the default — `iogpu.wired_limit_mb: 0`), and it's used in **exactly one place**: the sc-12178 wired *clamp*, which caps **pinning** (non-pageable residency), not usage. `set_wired_limit` above 81 GiB is literally the crash. Clamping to 81 GiB gives you the *maximum* MLX can pin; the ~47 GiB above it stays usable as pageable memory. The **fit-gate budgets against your full 128 GiB** (`probe_total_unified_memory_gib`), so model admission is never limited to 81 GiB. The clamp only engages if you set a GPU cap above ~63% of RAM; with no cap it's not applied at all. (If you ever want more *wired* memory for very large models, `sudo sysctl iogpu.wired_limit_mb=<mb>` raises the OS ceiling — separate axis; happy to make the clamp read the live sysctl if that becomes useful.)

Full `npm run rust:check` green; fresh CI running on the force-push.

PR #1564, opened 2026-07-16
fix(worker): fail a LoRA download that fetched nothing (sc-12288)

michaeltrefry:
Closes sc-12288.

> **Stacked on #1562** (sc-12283) because it touches the same function family and extracts a shared helper from it. Base retargets to `main` automatically when #1562 merges. **Review #1562 first** — the diff here is only the LoRA commit.

## Why

`run_lora_download_job` had **no zero-file check at all**. The sibling `run_model_download_job` got one in sc-9909; this never did.

So a LoRA whose declared `source.file` is absent from the repo — renamed upstream, typo'd entry, unpublished — resolves zero files, downloads nothing, and reports the job **completed**. The API's install probe (`lora_huggingface_cached_file`) then finds an empty cache and the catalog still reads "not installed". The user gets a job that claims success and a LoRA that never appears, with no error anywhere.

## What changed

- **Fail when the resolve yields zero files.** An empty `files` list is *not* an omission here — it deliberately means "fetch the (small) whole repo" (`create_lora_download_job`) — but that only *widens* the filter, so resolving nothing still means there was nothing to fetch.
- **Also check per pattern.** `create_lora_download_job` accepts an explicit `source.files` **list** and the filter ORs across it, so a multi-file adapter (a high/low-noise pair) that lands one half and not the other passes the aggregate check and completes as a torn install — the sc-12283 gap, LoRA side. No shipping LoRA declares a list today; this guards the case rather than waiting for it. **Happy to cut this back to the literal aggregate check if you'd rather.**
- **Extract `unmatched_patterns`** so the model and LoRA paths share one implementation.
- **`check-download-patterns.mjs` now covers `builtin.loras.jsonc`** (`source.file` / `source.files`). The LoRA path hard-fails now, so it needs the same authoring-time pre-flight. 224 patterns total (217 model + 7 LoRA).

## Scope — what this deliberately does *not* touch

**LoRAs from non-HF sources are unaffected.** `create_lora_download_job` requires a `repo` and only queues this job for a `provider: "huggingface"` source; URL imports and uploads go through the import path.

**`ensure_hf_files_cached` is unchanged, deliberately.** A zero-match filter there is a designed no-op — its callers keep their own "resolve the tier subdir, else fall back" presence checks, so it degrades rather than breaks.

## Why hard-failing is safe

Gated the same way as sc-12283: swept every entry in `builtin.loras.jsonc` against its repo, read through this workspace's own `strip_jsonc_comments` — **7 entries, all declaring an explicit `source.file`, all 7 present, zero whole-repo entries.** Nothing shipping relies on the missing check passing silently.

## Verified

- **639/639** `sceneworks-worker --lib --features backend-candle`; fmt + clippy(candle) clean.
- Both new tests are real guards: with the checks disabled, each fails on **its own assertion**, not on a stub gap.
- Script negative control: renaming a LoRA's `file` to one that doesn't exist upstream is caught, named, exit 1 — the exact scenario the fix guards.

### DIFF
--- a/crates/sceneworks-worker/src/model_jobs.rs
+++ b/crates/sceneworks-worker/src/model_jobs.rs
@@ -147,24 +147,12 @@ pub(crate) async fn run_model_download_job(
     // load-side half of that failure is sc-12279, which now falls back to a complete sibling tier;
     // this stops the torn tier being created in the first place.)
     //
-    // Testing each pattern against the RESOLVED files is sound: `snapshot.files` is every remote file
-    // matching ANY pattern, so a pattern with no match among them matched none in the repo.
-    //
     // Safe to hard-fail: swept all 217 `downloads[].files` patterns across the 53 HF repos that
     // declare one (builtin.models.jsonc, comment-stripped through this workspace's own
     // `strip_jsonc_comments`) — every pattern matches at least one file today, so no download that
     // works now begins to fail. A future entry that trips this is a real publishing gap, and failing
     // loudly at download beats a phantom install the user only discovers at generation time.
-    let unmatched = files
-        .iter()
-        .filter(|pattern| {
-            !snapshot
-                .files
-                .iter()
-                .any(|file| pattern_matches(pattern, &file.path))
-        })
-        .cloned()
-        .collect::<Vec<_>>();
+    let unmatched = unmatched_patterns(&files, &snapshot);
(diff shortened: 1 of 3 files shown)


michaeltrefry:
Accidental merge to a branch instead of main

PR #1569, opened 2026-07-16
feat(worker): run Mochi 1 video jobs — both backends, tier dirs, fit gate (sc-11992)

michaeltrefry:
B2 of [epic 1788](https://app.shortcut.com/trefry/epic/1788) — [sc-11992](https://app.shortcut.com/trefry/story/11992). B1 (`a59cada7`) declared Mochi to the app; B3 (`0c9061df`) shaped the request. **This wires the worker so a Mochi text-to-video job actually runs.**

## Scope vs each AC

| AC | Status |
|---|---|
| A Mochi t2v job produces frames | ✅ **GPU smoke run on real weights** — all 3 tiers render 848×480 × 7 frames; at the engine's full 64 steps q4 renders a recognizable subject |
| …reports monotonic progress | ✅ Smoke asserts monotonicity + that `Decoding` is reached; the shared `generate_video` funnel bridges it to job progress/heartbeat |
| …cancels promptly | ✅ Inherited from `generate_video` (`CancelFlag` threaded, stall watchdog, abandon deadline) — not re-implemented |
| Selecting q4/q8/bf16 loads the right tier | ✅ `mochi_tier_order` + `resolve_mochi_model_dir`; **all three tiers load in the smoke** |
| …fetched on demand if toggled but not downloaded | ✅ `ensure_mochi_{q8,bf16}_present` off the pinned revision |
| gpu smoke test passes on macOS | ✅ `mochi_mlx_smoke.rs` — **actually run**, exit 0, all 3 tiers, incl. a real **motion floor** (a frozen clip now fails it) |
| 🔴 Candle lane | ✅ All three traps fixed + a fourth found |
| 🔴 Fit gate | ✅ `mlx_fit_gate::mochi_fit_check`, frame-aware — **called via the `mochi_preflight` seam and pinned there** (reject@151 / admit@19 on a 64 GB Mac) |
| 🔴 Frame seam | ✅ Centralized as `video_frame_count`, and pinned **at the caller** via `mochi_preflight` |

## The candle lane

The story text is MLX-only, but B1 already routed Windows (`VideoModelCaps::new("mochi_1", true, true, false, false)`) and A6 CUDA-validated the lane against the *same* hosted tiers (the manifest carries no `platforms` tag). Shipping the MLX half alone would leave Windows **routed but unserved**. Three traps, each now pinned by a test:

1. **`candle_video_engine_id` didn't resolve `mochi_1`** → `resolve_candle_video_route` never reached the generic arm → `CandleVideoRoute::Stub` → a **procedural fake video**, not an error. The worst failure mode of the three.
2. **`candle_video_adapter_label`'s `_` arm is the Wan default** → `mochi_1` labelled a Wan adapter. Wrong provenance, no error.
3. **`candle_video_default_repo`** — same fall-through shape.

**A fourth, not in the brief:** the same else-arm routed Mochi into `candle_wan_sampling`, whose dense-5B tail forces `CANDLE_WAN5B_INTERIM_STEPS` (**20**) — silently overriding the AsymmDiT's own **64**-step default with a Wan tuning constant. Mochi gets its own sampling arm (true CFG: negative prompt + guidance, `None` ⇒ engine defaults).

**Tier dirs serve both lanes.** `resolve_mochi_model_dir` + `ensure_mochi_{q8,bf16}_present` + `mochi_tier_quant` are compiled for the **superset** cfg, so `advanced.mlxQuantize` works off-Mac and the candle arm calls them instead of `candle_video_snapshot_dir` (which has no notion of the tier layout).

**Does Windows actually work end-to-end?** Routing/labelling/tier-resolution: yes, and covered. **But** see sc-12306 — the candle video lane has **no VRAM gate at all**, and Mochi's untiled decode needs ~60 GiB at the 5 s default, which no consumer GPU has. Short clips will work; the default duration will CUDA-OOM (a catchable error, so it maps to a job error rather than a crash — the epic AC holds, but the UX is poor). A6 validated candle at a 64×64×7 fixture geometry, so the lane has never been exercised at a production bucket either.

**I could not compile the candle lane locally** — it is `cfg`'d out on macOS and cross-compiling needs `nvcc`. `windows-candle.yml` runs `cargo check`, `cargo clippy -D warnings` **and** `cargo test --features backend-candle` on every same-repo PR, so the arms + the candle-gated tests (**2 new** — `candle_mochi_resolves_the_engine_and_never_falls_to_the_stub`, `candle_mochi_default_repo_is_the_shared_mochi_turnkey` — plus `candle_video_adapter_labels_are_per_family`, which is **pre-existing and extended** here, not new) are genuinely gated there.

## The fit gate

Mochi holds **~18.7 GiB resident** for the whole run (`supports_sequential_offload: false` ⇒ nothing drops) and its AsymmVAE decode is **untiled**, so the peak grows **linearly with clip length**.

**Why the generic `mlx_fit_gate` cannot cover it** — two independent reasons:
- `predicted_peak_gb` = Σweights + a **1024²-calibrated constant**. It is resolution-blind *by construction* (the generator is cached across resolutions), so a 7-frame and a 151-frame request are indistinguishable to it. A constant cannot express a term that spans ~55 GiB.
- `supports_sequential_offload: false` ⇒ `weights_fit_floor` (sc-12179) admits on the **weights alone** (18.7 GiB fits almost any Mac). That floor is correct for its own purpose but is exactly what makes the generic gate unable to protect Mochi.

**It must be pre-flight.** MLX's default error handler is `exit(-1)` (sc-12178/12179, #1544) — a hard process kill that **cannot be mapped** to a job error afterwards.

**Formula** (`mochi_decode_peak_gb`), derived from the architecture with every constant named:
```
decode_peak = live(2) × C(128) × (frames + 5) × W × H × 4 B
needed      = resident_weights + decode_peak + OS_RESERVE_GB
```
- `C = 128` = `decoder_block_out_channels[0]` — the final `block_out` stage runs 128 ch at **full** output resolution; every earlier stage is strictly smaller.
- `+5` = the causal decode's extra leading frames, paid before `drop_last_temporal_frames` trims them.
- **`4 B` = f32, NOT bf16** — 🔴 this **deliberately disagrees with sc-12291 and B1's manifest**, which both derived `× 2 B`. The port pins `Dtype::Float32` (`mlx-gen-mochi/src/vae.rs:270`) and the crate says so: *"the reference decode ran bf16 — the parity residual reflects that bf16 rounding."* **sc-12291's decode table is ~2× low** (19→7, 61→19, 151→45 GiB). Corrected table filed there.

  **`mlx.minMemoryGb: 96` is wrong in its DERIVATION only — the value itself survives, so the manifest is unchanged.** B1 reached 96 from the bf16 (halved) peak, so the reasoning is void; but re-derived against the shipped f32 decode the requirement still clears 96 on every tier. Measured against the real hosted weights (`.safetensors` byte sums), + the 60.56 GiB decode at the 151-frame default + 2 GiB `OS_RESERVE_GB`:

  | tier | resident | total needed | vs 96 |
  |---|---|---|---|
  | q4 | 18.73 GiB | **81.29 GiB** | clears |
  | q8 | 22.10 GiB | **84.66 GiB** | clears |
  | bf16 (worst case) | 28.41 GiB | **90.97 GiB** | clears |

  So 96 remains a correct advisory floor and needs no change. An earlier revision of this PR body claimed "96 inherits" the 2× error — that overstated it and contradicted this PR's own `MOCHI_DECODE_LIVE_TENSORS` doc, which says 96 is consistent. The doc was the accurate one.
- `live = 2` = B1's own concrete "residual + intermediate live at once" (it then hedged 2–3×; 2 is the defensible claim).

Sanity, all in **GiB** (an earlier revision of this list mixed in decimal GB — the code was always correct, it divides by `BYTES_PER_GIB`): 7 frames ⇒ ~4.66 decode; 19 ⇒ ~9.32; 61 ⇒ ~25.62; 151 ⇒ ~60.56; 163 ⇒ ~65.21. The 5 s default needs **~79 GiB active** (18.73 weights + 60.56 decode) — consistent with B1's declared 96 GiB floor.

**Message** follows the `too_big_error` convention (names the model, "unified memory", need vs budget) and adds the lever unique to Mochi — the generic advice ("smaller tier, lower resolution") is near-useless here, since Mochi has one trained bucket and q4→bf16 is ~11 GiB against a ~60 GiB decode:
> `mochi_1 needs ~81 GB of unified memory to render a 151-frame 848x480 clip (~19 GB of weights, held resident for the whole run, plus an untiled VAE decode whose peak grows with clip length) but this machine has ~64 GB. Shorten the clip — the decode peak scales roughly linearly with duration — or run on a Mac with more memory.`

**Tested**: same machine, 19 frames admitted / 151 rejected (a frame-blind gate cannot pass this); message shape; linearity in frames *and* pixels; no-signal ⇒ never block; and that the shared parent T5/VAE are folded in (summing only the tier dir would under-count by ~9.7 GiB and admit a job that then dies).

## The frame seam — and a correction

`video_jobs.rs:5216` was a binary `if is_ltx { ltx } else { wan }`, so **every non-LTX model including `mochi_1`** inherited Wan's stride. Rather than add a third inline branch, the ladder is centralized as **`video_frame_count(model, raw)`**, called by **both** lanes — which is what makes the mapping unit-testable *by model id* and keeps the lanes from drifting.

The other `wan_frame_count` sites (`~5445, 5671, 5771, 5818, 5939, 5990, 6058, 6193`) are all inside **family-specific** functions (scail2 / bernini / wan-vace / comfyui) that Mochi never reaches — verified; left alone.

🔴 **The brief's characterization was arithmetically wrong**, and the error originates in B1's analysis (it propagated to sc-11992's comment and sc-11993's). The claim: *"`wan_frame_count(150) = 145`. And `145 % 6 == 1`, so the runtime accepts it → a 5 s request silently renders 4.83 s."*

- `wan_frame_count(150)` is **149**, not 145 — `150 − ((150−1) % 4)` = 150 − 1. (145 is unreachable: it needs `149 % 4 == 5`.)
- `149 % 6 == 5`, so the engine **hard-rejects** it.

The bug is real but **loud**: every default-duration Mochi job would die on *"num_frames must be 1 + 6·k (got 149)"*. Stronger — a silent wrong length is **provably impossible** on this pairing: whenever Wan's answer lands on Mochi's lattice it *equals* Mochi's snap (if `wan(raw) = W ≡ 1 mod 6` then `raw − W ∈ 0..=3`, so `W` is Mochi's `lower` and nearest-with-ties-low picks it). Verified exhaustively over raw 1..1999 (0 silent-wrong cases) and pinned as an invariant in the test.

## Tests

**27 added** (counted as `#[test]` attributes in the diff vs `origin/main`): **17** in `video_jobs` — incl. the 3 `mochi_preflight` seam tests and **2 new** candle-gated ones (`candle_mochi_resolves_the_engine_and_never_falls_to_the_stub`, `candle_mochi_default_repo_is_the_shared_mochi_turnkey`) — **5** fit-gate, **4** `smoke_support` motion-metric, **1** GPU smoke (`#[ignore]`d). `candle_video_adapter_labels_are_per_family` is **pre-existing and extended** here, so it is not counted as new. Deliberately placed on the **superset** cfg where the code is shared, so the tier resolver / stride / fetch gating run on the macOS lane **and** the windows-candle lane.

**Mutation-checked** (real exit codes; each restored after).

🔴 **Correction — an earlier revision of this section claimed every mutation was caught. That was not accurate.** A review round found **three more that survived 835/0 green**, all at the *production seam* rather than in the pure functions: the tests pinned the FUNCTIONS, not the CALL SITE. `generate_mochi` is `async` and needs a live `ApiClient`/`JobSnapshot`, so every decision left inline in it was unreachable from a unit test — which left the epic AC *"an unsupported environment shows an actionable error, not a crash"* **unenforced by any test**. Fixed by extracting `mochi_preflight` (below); all three are now RED, re-run and pasted.

| Mutation | Before | Now |
|---|---|---|
| `video_frame_count(...)` → `wan_frame_count(...)` **at the caller** | 🔴 **survived** 835/0 | **caught** — 2 tests, exit 101 |
| Fit gate's `frames` argument **hardcoded to `7`** | 🔴 **survived** 835/0 | **caught** — 1 test, exit 101 |
| **`mochi_fit_check(...)?` call deleted** | 🔴 **survived** 835/0 | **caught** — 1 test, exit 101 |
| **Frozen clip** (temporal path → N copies of one still) | 🔴 **passed the whole smoke** | **caught** — smoke fails, exit 101 |
| Route Mochi to `wan_frame_count` (function-level) | survived first → drove the `video_frame_count` extraction | caught |
| Fit gate ignores frame count (function-level) | — | caught (3 tests) |
| Candle adapter arm removed → Wan | — | caught |
| Candle default-repo arm removed → Wan | — | caught |
| Candle `engine_id` arm removed → Stub | — | caught |
| `mochi_available` drops the t2v mode gate | — | caught |
| `resolve_mochi_model_dir` ignores the tier order | — | caught |
| `ensure_video_engine_weights` loses the fail-loud arm | — | caught |
| Fit gate misses the shared T5/VAE | — | caught |

### The `mochi_preflight` seam

The fix for the first three. The frame lattice, the fit gate, and the tier quant move out of `generate_mochi` into a **pure, synchronously testable** seam:

```rust
fn mochi_preflight(model, engine_id, tier_dir, raw_frames, width, height)
    -> WorkerResult<MochiPreflight>   // { frames, quant }
```

`generate_mochi` becomes a thin caller. `frames` and `quant` are returned as **one unit** on purpose: both are things the generation arm cannot build its `VideoGenInput` without, and `mochi_preflight` is the only thing that produces them — so it cannot obtain a frame count or a quant marker on a path that skipped the gate. The three tests assert **by model id**, through the real on-disk weight scan (sparse fixtures at the true hosted byte sizes) and the live `SCENEWORKS_MLX_MEMORY_CAP_GB` budget:

- `mochi_preflight_coerces_the_frame_count_on_mochis_lattice_by_model_id` — 150 ⇒ **151**, `== mochi_frame_count(150)`, `!= wan_frame_count(150)`.
- `mochi_preflight_rejects_the_5s_default_on_a_64gb_mac` — 151 frames ⇒ 81.29 GiB ⇒ **Err**, message names the model + the clip-length lever.
- `mochi_preflight_admits_a_short_clip_on_the_same_64gb_mac` — 19 frames ⇒ 30.05 GiB ⇒ **Ok**. Without this the reject test would pass against a gate that blanket-refuses; the pair is what proves the gate is **frame-sensitive**.

**Residual, disclosed and tracked as [sc-12318](https://app.shortcut.com/trefry/story/12318):** a unit test still cannot reach `generate_mochi` itself, so a deliberate multi-line rewrite of its single `mochi_preflight` call could still bypass the gate. The seam raises that from a one-line slip (what actually happened) to an obvious rewrite, but it does not make it impossible — closing it fully needs integration coverage of the async arm. The candle lane's `video_frame_count` call in `generate_candle_video` has the same shape and the same residual.

### The smoke's motion floor

The smoke's comment promised *"the clip must actually MOVE"* — then only **printed** the means. **A fully frozen clip passed every assertion**, and the 6× temporal path is Mochi's distinguishing feature. Now asserted with `mean_abs_frame_delta` (0 **iff** bit-identical; unlike an `image_mean` delta, which a pan preserves — so the old print-only metric could read ~0 on a live clip and be cleared by brightness drift on a frozen one), on **every consecutive pair** (strictly stronger than first-vs-last: it also catches a clip that moves once then freezes).

ε = **1.0**, from the observed run — not invented. Real per-pair deltas at 848×480 × 7f @ 4 steps, seed 42:

| tier | per-pair deltas | min |
|---|---|---|
| q4 | 7.58, 5.58, 3.60, 3.41, 2.60, 5.21 | 2.597 |
| q8 | 30.04, 6.11, 2.57, 7.23, 14.93, 11.89 | **2.573** |
| bf16 | 30.87, 6.47, 2.75, 7.19, 16.24, 12.67 | 2.752 |

Tightest real observation **2.573**; 1.0 sits ~2.6× below it (margin against a different prompt/steps/seed) while a frozen pair scores **exactly 0.0**. Proven load-bearing: freezing the clip fails the smoke with *"frames 0->1 are FROZEN (mean abs delta 0.000 <= floor 1)"*, exit 101. Backed by 4 GPU-free unit tests on the metric.

The candle mutations were run against verbatim copies of the real match statements in a scratch harness (the arms can't execute on macOS); the pre-B2 state fails 7 assertions.

`npm run rust:check` **exit 0** (fmt + `clippy -D warnings` + test + build, 0 warnings); worker lib suite **844 passed / 0 failed**; GPU smoke **exit 0** on real weights, all 3 tiers, now including the motion floor.

## Not the stub

`ensure_video_engine_weights` (the sc-4176 fail-loud gate) gets a Mochi arm — without it, unresolvable weights hand the user a **procedural fake video** instead of the resolver's precise error. Tested.

## Follow-ups filed (not buried)

- **[sc-12318](https://app.shortcut.com/trefry/story/12318)** — the async video generation arms are unreachable from a unit test, which is what let 3 mutations survive here. `mochi_preflight` closes the Mochi case as far as a unit test can; this tracks the remaining residual.
- **[sc-12306](https://app.shortcut.com/trefry/story/12306)** — the candle video lane has **no VRAM gate at all** (for any model); Mochi at the 5 s default will OOM every consumer GPU.
- **[sc-12291 comment](https://app.shortcut.com/trefry/story/12291)** — the bf16-vs-f32 correction: its decode table and B1's `mlx.minMemoryGb: 96` are ~2× low. Also flags a possibly cheaper lever than tiling (the decoder already takes a dtype; the reference *was* bf16 — the question is whether the `vae_parity` golden holds).

## Known limitations (disclosed)

- The fit-gate constants are **derived, not measured**. The dtype is read from shipped code (a clean 2×); the liveness multiplier (2 vs 3) is ~1.5× uncertainty. B5/sc-12291 settle it — `mochi_decode_peak_gb` is the one place to update.
- Candle **not compiled locally** (no `nvcc`); relies on `windows-candle.yml`'s PR gate.
- The smoke renders **7 frames** by default — the untiled decode is why (~60 GiB at 151).

### DIFF
--- a/crates/sceneworks-worker/src/lib.rs
+++ b/crates/sceneworks-worker/src/lib.rs
@@ -330,6 +330,15 @@ mod pid_tier_mlx_smoke;
 // the already-packed weights (packed-detected) and bf16 loads dense (Quant::None).
 #[cfg(all(test, target_os = "macos"))]
 mod sana_mlx_smoke;
+// Real-weight MLX smoke for the Mochi 1 quant-matrix video lane (epic 1788, sc-11992). macOS-only;
+// drives `crate::inference_runtime::load("mochi_1")` with a per-tier LoadSpec against the
+// pre-quantized q4/q8 + dense bf16 tier subdirs of the SceneWorks/mochi-1-mlx turnkey. On-device
+// evidence that the worker tier path loads (`WeightsSource::Dir` = the TIER dir, with the shared
+// T5-XXL/tokenizer/AsymmVAE resolved from that dir's PARENT — the A6 sibling layout) and renders a
+// non-degenerate, MOVING clip at every downloaded tier, with monotonic progress that reaches decode
+// (the video job lane has no background heartbeat during a generation).
+#[cfg(all(test, target_os = "macos"))]
+mod mochi_mlx_smoke;
 // On-device per-tier memory-footprint measurement harness (sc-8516, epic 8506). Test-only + macOS-only;
 // #[ignore]d real-weight smokes that drive `crate::inference_runtime::load(id)` + ONE generation while sampling the MLX
 // process-global memory counters (mlx_rs::memory::{reset_peak_memory, get_active_memory, get_peak_memory})
--- a/crates/sceneworks-worker/src/mlx_fit_gate.rs
+++ b/crates/sceneworks-worker/src/mlx_fit_gate.rs
@@ -341,6 +341,214 @@ fn sysctl_total_unified_memory_gib() -> Option<f64> {
     None
 }
 
+// ---------------------------------------------------------------------------------------------
+// Mochi 1: the FRAME-DEPENDENT decode fit gate (epic 1788 / sc-11992)
+// ---------------------------------------------------------------------------------------------
+//
+// Why Mochi needs its own gate rather than riding the generic one above:
+//
+//  1. The generic gate is deliberately RESOLUTION-BLIND. `predicted_peak_gb` = Σweights +
+//     HEADROOM_GB, where HEADROOM_GB is a 1024²-calibrated CONSTANT — the load-time seam cannot see
+//     the request geometry (the generator is cached across resolutions, see the HEADROOM_GB note).
+//     That structure is right for image models, whose transient is roughly request-independent once
+//     calibrated. It is WRONG for Mochi: its AsymmVAE decode is UNTILED (`vae.decode(latents)`
+//     materializes the whole clip — sc-12291), so the peak grows LINEARLY IN CLIP LENGTH. A 7-frame
+//     and a 151-frame request differ by ~55 GiB on the same model. A constant cannot express that.
+//
+//  2. Mochi's `supports_sequential_offload: false` ⇒ `weights_fit_floor` admits on WEIGHTS ALONE
+//     (18.73 GiB fits almost any Mac), so the generic gate would happily admit a 151-frame job that
+//     then needs ~79 GiB. The floor is correct for its own purpose (sc-12179: never wall-reject a
+//     machine that used to work) but it is precisely what makes the generic gate unable to protect
+//     Mochi.
+//
+//  3. This MUST be a pre-flight rejection, not a caught error. MLX's default error handler is
+//     `exit(-1)` — a hard process kill (sc-12178/12179, GitHub #1544). An `exit(-1)` cannot be
+//     mapped to a job error after the fact, so the only place to honor the epic's "actionable error,
+//     not a crash" AC is BEFORE the decode allocates.
+//
+// The gate is therefore a per-REQUEST admission check (it sees frames + geometry), layered beside —
+// not inside — the generic per-LOAD one, and it budgets on a DERIVED architectural formula rather
+// than a calibration constant, because nothing has been measured on-device yet (B5/sc-11995 backfills
+// `footprint.residentMemoryBytes`/`peakMemoryBytes`; sc-12291 tiles the decode and should then cut
+// the frame term sharply).
+
+/// The AsymmVAE decoder's final-stage channel count — `decoder_block_out_channels[0]` (config
+/// default `[128, 256, 512, 768]`). The last `block_out` mid-block runs THIS many channels at FULL
+/// output resolution, which is the decode's peak stage: every earlier stage is either fewer pixels
+/// (pre-upsample) or, at 256/512/768 channels, at 1/4/1/16/1/64 of the spatial extent, so their
+/// products are strictly smaller.
+const MOCHI_DECODE_CHANNELS: f64 = 128.0;
+
+/// Bytes per element in the MLX AsymmVAE decode. **f32 (4), NOT bf16 (2)** — `MochiVaeDecoder::
+/// from_weights` pins `Dtype::Float32` and `decode_denormalized` casts the latent to that dtype, so
+/// the whole decode flows f32 ("the reference decode ran bf16 — the parity residual reflects that
+/// bf16 rounding", mlx-gen-mochi/src/vae.rs).
+///
+/// ⚠️ This DELIBERATELY disagrees with the sc-12291 / B1-manifest derivation, which assumed **bf16**
+/// (`128 × 156 × 480 × 848 × 2 B`) and therefore under-predicts the real MLX peak by 2×. Surfaced as
+/// a follow-up on sc-12291 — that story's table (19→7, 61→19, 151→45 GiB) and B1's
+/// `mlx.minMemoryGb: 96` are both bf16-derived and want re-deriving against the shipped f32 decode.
+/// This gate uses the dtype the code actually runs.
+const MOCHI_DECODE_BYTES_PER_ELEM: f64 = 4.0;
+
+/// Full-resolution tensors live simultaneously at the decode peak. `MochiResnetBlock3D` is
+/// `GroupNorm → silu → CausalConv3d → GroupNorm → silu → CausalConv3d → + residual`, so the residual
+/// and the in-flight intermediate are both live across the block — the "residual + intermediate live
+/// at once" claim B1's manifest derivation makes. B1 then hedged "2–3×"; 2 is the concrete,
+/// defensible liveness (3 assumes an extra un-freed temporary, which MLX's allocator need not hold).
+/// Combined with the f32 correction above this lands at ~79 GiB for the shipped 5 s / 151-frame
+/// default — consistent with B1's declared `mlx.minMemoryGb: 96` floor.
+const MOCHI_DECODE_LIVE_TENSORS: f64 = 2.0;
+
+/// Extra leading frames the causal decode materializes before `drop_last_temporal_frames` trims them:
+/// `temporal_ratio − 1` = 5. The peak is paid on the UNTRIMMED length, so a 151-frame clip decodes
+/// 156 frames wide.
+const MOCHI_DECODE_EXTRA_FRAMES: u32 = 5;
+
+/// Predicted AsymmVAE decode peak (GiB) for a `frames` × `width` × `height` clip.
+///
+/// `live × C × (frames + 5) × H × W × 4 B`. Linear in BOTH clip length and pixel count, because the
+/// peak tensor is `[1, C, T, H, W]` and the decode is untiled. This is the term the generic gate's
+/// flat `HEADROOM_GB` structurally cannot represent.
+///
+/// DERIVED, not measured. Sanity points at the native 848×480 bucket: 7 frames ⇒ ~5.0 GiB, 19 ⇒
+/// ~10.1, 61 ⇒ ~27.5, 151 ⇒ ~60.6, 163 ⇒ ~65.2 GiB.
+pub(crate) fn mochi_decode_peak_gb(frames: u32, width: u32, height: u32) -> f64 {
+    let decoded_frames = f64::from(frames.saturating_add(MOCHI_DECODE_EXTRA_FRAMES));
+    MOCHI_DECODE_LIVE_TENSORS
+        * MOCHI_DECODE_CHANNELS
+        * decoded_frames
+        * f64::from(width)
+        * f64::from(height)
+        * MOCHI_DECODE_BYTES_PER_ELEM
+        / BYTES_PER_GIB
+}
+
+/// Predicted whole-generation peak (GiB) for a Mochi request: the ALL-RESIDENT weights
+/// (`supports_sequential_offload: false` ⇒ T5-XXL + AsymmDiT + VAE are held for the whole run, so
+/// nothing drops) + the frame-dependent decode peak + [`OS_RESERVE_GB`]. `None` when the weights are
+/// unmeasurable (`weight_bytes == 0`) ⇒ no signal ⇒ the gate never blocks, matching
+/// [`predicted_peak_gb`].
+pub(crate) fn mochi_needed_gb(
+    weight_bytes: u64,
+    frames: u32,
+    width: u32,
+    height: u32,
+) -> Option<f64> {
+    (weight_bytes > 0).then(|| {
+        weight_bytes as f64 / BYTES_PER_GIB
+            + mochi_decode_peak_gb(frames, width, height)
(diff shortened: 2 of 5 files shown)


michaeltrefry:
**CI green on all lanes** — read from the API rollup, not `gh pr checks --watch`: web ✅ · build-windows ✅ · **candle-worker ✅** · nax-worker ✅ · parity ✅ (package-windows skipped by design).

### The candle lane caught a real defect — and then proved itself

The first push **failed `candle-worker`**, which is exactly the lane I flagged as un-runnable locally. `cargo clippy --features backend-candle -D warnings` found `mochi_engine_id` as **dead code**: it carried the superset cfg the shared tier helpers use, but every caller (`mochi_available`, `resolve_video_route`, `ensure_video_engine_weights`) is macOS-only — the candle lane resolves the same id through its own `candle_video_engine_id` table.

Fixed in `5bbcc872` by keying the candle arm off the already-resolved `engine_id` (`engine_id == "mochi_1"`), mirroring the `is_ltx` binding immediately below it rather than re-deriving the family from the model string — one source of truth per lane — and gating `mochi_engine_id` to macOS. I then audited every remaining superset symbol to confirm each is reachable from `generate_candle_video`'s call graph.

This is precisely the sc-12223 failure mode, caught by the gate that exists to catch it.

### The candle half is now verified by EXECUTION, not inspection

`windows-candle.yml` compiled the arms **and ran the tests** on the real CUDA box. All three candle-gated tests — the ones only that lane can execute — passed:

```
test video_jobs::candle_video_label_tests::candle_mochi_resolves_the_engine_and_never_falls_to_the_stub ... ok
test video_jobs::candle_video_label_tests::candle_mochi_default_repo_is_the_shared_mochi_turnkey ... ok
test video_jobs::candle_video_label_tests::candle_video_adapter_labels_are_per_family ... ok
```

plus the 5 fit-gate tests and B1's `mochi_resolves_one_engine_id_to_the_gen_core_descriptor` against the linked **candle** catalog.

So the PR body's "could not compile the candle lane locally" caveat still describes my local loop, but it is no longer an open risk on this change: the arms compile, clippy is clean at `-D warnings`, and the traps are pinned by tests that actually ran off-Mac.

**Not merging** — coordinator merges.

PR #1582, opened 2026-07-17
test(worker): pin the post-download Mochi gate at the floor-vs-full seam (sc-12318)

michaeltrefry:
Meets **sc-12318 AC #1**, which was not met on main. Follow-up to #1573 / #1575 / #1577.

## The gap

sc-12318's stated AC:

> *"a mutation that removes the `mochi_preflight` call from `generate_mochi` … must be caught by the default `cargo test -p sceneworks-worker` run"*

Measured on main — inlining the seam (`video_frame_count` + `mochi_tier_quant`, no `mochi_fit_check`):

```
cargo test -p sceneworks-worker            → 856 passed; 0 failed     ← NOT caught
cargo clippy --all-targets -D warnings     → `mochi_preflight` is never used   ← caught
```

### Why it regressed: the pre-check masks the preflight

sc-12322's `mochi_precheck` runs the **same** `mochi_fit_check` earlier. Every `_using` over-budget test stages a **complete** tier, so the pre-check measures the full weights and refuses first — making `mochi_preflight` unobservable. The tests passed because *something* refused, just not the thing under test.

Nothing exercised the one case where the two gates **disagree**: floor admits, full weights refuse.

## That case is real, reachable, and a SIGKILL if the preflight goes

The pre-check only sees a *floor* when the requested tier is absent. On a **64 GB Mac** requesting a not-yet-installed tier:

| frames | floor (pre-check) | full q4 (preflight) | outcome |
|---|---|---|---|
| 109 | 55.98 | 64.99 | admits → **refuses** |
| 115 | 58.31 | 67.32 | admits → **refuses** |
| 127 | 62.97 | 71.97 | admits → **refuses** |

**2.8s–4.2s** — well under `hardMaxDuration`'s 151 frames. Delete the preflight and those jobs walk into MLX's `exit(-1)`, with `cargo test` green.

## The test — no download, no network

Getting the two gates to disagree normally needs a real ~13 GiB fetch to materialize the tier. This avoids that entirely:

- `mlxQuantize: 8` ⇒ the pre-check names the **absent** `q8/` dir, so `mochi_resident_bytes` folds only the shared `text_encoder/` + `vae/` siblings — a 9.73 GiB floor (`q4/` is *not* a folded component).
- `drive_mochi_arm`'s `data_dir` has no HF cache ⇒ `huggingface_snapshot_dir` is `None` ⇒ `ensure_mochi_tier_present` early-returns `Ok(())` instead of fetching.
- `resolve_mochi_model_dir` falls back to the installed `q4/` ⇒ the preflight measures the **full** 18.73 GiB and refuses.

Kills the AC #1 mutation — **856 passed / 1 failed**, and the failure is this test, for the stated reason.

## Why every geometry term is explicit

`VideoRequest`'s defaults are none of Mochi's, and the window is only ~9 GiB wide (the q4 tier). Each default slides the request off the seam into the region where both gates agree — a green test proving nothing. The premise assertions caught two of these while I was writing it:

| term | default | Mochi | effect if left default |
|---|---|---|---|
| `fps` | 25 | 30 | 95 raw ⇒ 97 frames — outside the window |
| frame | 768×512 | 848×480 | decode term wrong |
| stride | 64 | 16 (declared) | 848 floors to **832**, decode ~2% low |

Hence the explicit `modelManifestEntry: { limits: { requiresDimensionsMultipleOf: 16 } }` — production passes the manifest entry, and a fixture that omits it isn't running the geometry it claims to.

## Verification

`npm run rust:check` (fmt + lint + test + build) — exit **0**. **857 passed / 0 failed**.

## ⚠️ AC #2 still unverified — by construction

`generate_candle_video` is `#[cfg(all(not(target_os = "macos"), …))]`, so the candle `video_frame_count` line **is not compiled on a Mac**. I ran that mutation locally and got a green result; it is meaningless, and I'm not claiming it. #1575's `generate_candle_video_using_…` test is built for it and runs on `candle-worker` — someone on that lane (or a CI push) should confirm it kills the swap before this story closes.

### DIFF
--- a/crates/sceneworks-worker/src/video_jobs.rs
+++ b/crates/sceneworks-worker/src/video_jobs.rs
@@ -11904,6 +11904,105 @@ mod tests {
         );
     }
 
+    /// **The floor-vs-full seam** — the one case where sc-12322's two gates DISAGREE, and the only
+    /// thing that pins the post-download `mochi_preflight` call now that `mochi_precheck` runs the same
+    /// `mochi_fit_check` earlier.
+    ///
+    /// Why the sibling tests can't cover this: they stage a COMPLETE tier, so the pre-check measures the
+    /// full weights and refuses first — the preflight becomes unobservable, and inlining it away stays
+    /// green (caught only by clippy's dead-code cascade, not by `cargo test`; sc-12318 AC #1).
+    ///
+    /// This makes the two verdicts diverge, with **no download and no network**:
+    ///   * `mlxQuantize: 8` ⇒ the pre-check names the **absent** `q8/` dir, so `mochi_resident_bytes`
+    ///     folds only the shared `text_encoder/` + `vae/` siblings — a 9.73 GiB FLOOR (`q4/` is not a
+    ///     folded component). At 115 frames: 9.73 + 46.58 + 2 = **58.31 ⇒ ADMITS** a 64 GB budget.
+    ///   * `ensure_mochi_q8_present` then no-ops rather than fetching: `drive_mochi_arm`'s `data_dir`
+    ///     has no HF cache, so `huggingface_snapshot_dir` is `None` and `ensure_mochi_tier_present`
+    ///     early-returns `Ok(())`. That is what keeps this hermetic — the divergence normally needs a
+    ///     real ~13 GiB download to materialize the tier.
+    ///   * `resolve_mochi_model_dir` falls back to the installed `q4/`, so `mochi_preflight` sees the
+    ///     FULL 18.73 GiB: 18.73 + 46.58 + 2 = **67.32 ⇒ REFUSES**.
+    ///
+    /// So the job is admitted by the floor and refused by the full check — exactly the case the
+    /// post-download gate exists for. It is REACHABLE in production (a 64 GB Mac, bf16/q8 not yet
+    /// installed, ~2.8-4.2 s clip — well under `hardMaxDuration`'s 151 frames), and without the
+    /// preflight those jobs run into MLX's `exit(-1)`.
+    ///
+    /// Kills the sc-12318 AC #1 mutation — inline `video_frame_count` + `mochi_tier_quant` in place of
+    /// `mochi_preflight` — which passes 856/0 on main: the pre-check admits here, so nothing else
+    /// refuses, the engine loads, and both assertions below fail.
+    #[cfg(target_os = "macos")]
+    #[test]
+    fn generate_mochi_using_refuses_after_the_fetch_when_the_precheck_floor_admitted() {
+        // q4 installed at its real hosted size; q8 is NOT — the fixture stages no other tier.
+        let root = mochi_root_real_sized("arm_floor_vs_full");
+        let probe = ArmProbe::default();
+        // 3.8 s x 30 fps = 114 raw ⇒ 115 frames at 848x480, inside the 109..127 divergence window on a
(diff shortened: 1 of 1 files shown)


michaeltrefry:
**✅ AC #2 verified on the candle lane — the mutation is killed. The ⚠️ caveat can come off.** — claude, 2026-07-16

Picked this up from a Windows/CUDA box, which is the thing the PR body correctly says a Mac cannot do: `generate_candle_video` is `#[cfg(all(not(target_os = "macos"), …))]`, so the mutated line isn't compiled there.

## Result

Applied the story's AC #2 verbatim — `video_frame_count(&request.model, request.raw_frame_count())` → `wan_frame_count(request.raw_frame_count())` at `video_jobs.rs:5437` — and reverted it after:

| Step | Result |
|---|---|
| Baseline | `generate_candle_video_using_hands_the_engine_the_mochi_lattice_frame_count ... ok` |
| **AC #2 mutation** | **FAILED — `left: 149, right: 151`** |
| After revert | **671 passed / 0 failed** / 35 ignored; clippy `-D warnings` exit **0**; `git diff` empty |

The failure is the right one for the right reason: 149 is `wan_frame_count(150)`, off Mochi's `6k+1` lattice (`149 % 6 == 5`) — exactly what the engine's `validate_request` hard-rejects. #1575's second assertion (`assert_ne!(engine_frames(), wan_frame_count(150))`) bites on the same mutation, so both halves hold.

```
cargo test  -p sceneworks-worker --lib --features backend-candle
cargo clippy -p sceneworks-worker --all-targets --features backend-candle -- -D warnings
```
MSVC 14.44 (CUDA 12.9's nvcc rejects VS18/14.51), `CUDA_COMPUTE_CAP=120`.

## ⚠️ I tested the MERGE, not this branch — and that mattered

This branch is based on `15b53e01` and does **not** contain `e9277420` (sc-12347, merged ~an hour later), which rewired how `fps` resolves. That is not incidental here: the candle test's expected **151 is `duration × fps`**, so sc-12347 lands directly on this test's premise, and a green run on the stale branch would not have said anything about what actually merges.

Tested `5237ff90` = `771ed62e` (this PR) + `e9277420` (main). Re-derived after the merge: `duration` defaults to **6.0** (`safe_float(…, 6.0, …)`) and `resolve_fps` with no `modelManifestEntry` still yields **25** ⇒ 150 raw ⇒ **151**. sc-12347 leaves it intact. Merge was clean (additive both sides).

Minor: #1575's assertion message calls this *"the 5 s default"*, but the arithmetic is 6.0 s × 25 fps. The number is right, the label isn't — worth a word-fix if anyone's in there, not worth a PR.

## Scope of this claim — read it precisely

**This is not a GPU run, and I'm not claiming one.** The test drives `ArmProbe`'s stub loader — no weights, no device, **0.02 s**. What it needs is the candle *build*, and that's the whole reason the macOS lane can't reach it. So this validates **compilation + arm logic on the candle cfg**, not hardware behavior. That is exactly what AC #2 asks for, but it should not be read as a Blackwell inference validation.

Also worth stating plainly: the PR's `candle-worker` CI (IN_PROGRESS at the time of writing) will confirm the test **passes**. It cannot confirm it **kills the mutation** — that needs a mutation applied by hand, which is this run. Green CI is necessary, not sufficient, for AC #2.

## sc-12318 status

With AC #1 met by this PR's floor-vs-full test and AC #2 now measured rather than asserted, both ACs are met at the **`cargo test`** level rather than the clippy dead-code level — which was the story's whole point.

PR #1696, opened 2026-07-22
fix(training): point SDXL LoRA trainer at the installed turnkey repo (#1694)

### LINKED ISSUE #1694: Cannot start Traning
### What happened?

When i click train, it says sdxl not installed, butit is

Image
Image

### Steps to reproduce

goto train, select data set, switch from zimage turbo to sdxl, click train.
error

### Platform

Windows (NVIDIA / candle)

### SceneWorks version

0.8.0

### Model / LoRA involved (if any)

_No response_

### Logs / screenshots

```shell
21:26:17
api
debug
api_error status=400 detail=Base model 'sdxl' is not installed. Install it from the model catalog before starting a real training run (dry runs work without it).
```

michaeltrefry:
Fixes #1694.

## Problem

Selecting **SDXL** in the Training Studio and clicking **Train** fails with:

> `Base model 'sdxl' is not installed. Install it from the model catalog before starting a real training run (dry runs work without it).`

…even when SDXL *is* installed.

## Root cause

The stock `sdxl` training target named its base as the flat upstream repo `stabilityai/stable-diffusion-xl-base-1.0`, but **nothing in the app downloads that repo**. The catalog `sdxl` entry and the worker engine both install the SceneWorks quant-matrix turnkey `SceneWorks/sdxl-base-mlx` (`config/manifests/builtin.models.jsonc`, `crates/sceneworks-worker/src/engines.rs`).

The pre-flight gate `training_base_model_installed` checks the HF cache for the target's `base_model_repo`, finds the never-downloaded `stabilityai/...` repo missing, and 400s — even though the real base is on disk. A dry run skips the gate, which is why only a real "train" click trips it.

## Fix

Point the SDXL target's `base_model_repo` at `SceneWorks/sdxl-base-mlx`, matching the catalog + engine and the Illustrious pattern. Training resolves the dense `bf16/` tier through the existing `tiered_turnkey_train_dir` path — no resolver changes needed.

## Tests

- `stock_sdxl_target_points_at_installed_turnkey` (new): pins that the SDXL target names the turnkey, resolves training to its `bf16/` tier, and passes the install gate only once the dense tier is present.
- `flat_diffusers_snapshot_is_not_treated_as_a_tiered_turnkey`: the flat-resolution guard's former SDXL exemplar is now a turnkey, so it's rebased onto a synthetic flat repo — it exercises the resolution branch itself, independent of registry drift.
- Core contract test asserts the new repo; registry snapshot fixture regenerated (one line).

All `sceneworks-core` training-contract and `sceneworks-rust-api` training tests pass; fmt clean.

## This is one instance of a broader drift

The epic 8506 MLX re-host migration moved the **catalog** to `SceneWorks/*-mlx` turnkeys but left several older **training targets** pointing at upstream repos the app no longer downloads. Illustrious/Krea/LTX/Anima were updated (e.g. sc-10617 for Illustrious); these were not:

| training target | points at (upstream) | catalog installs |
|---|---|---|
| `z_image_turbo` | `Tongyi-MAI/Z-Image-Turbo` | `SceneWorks/z-image-turbo-mlx` |
| `lens` | `SceneWorks/Lens` | `SceneWorks/lens-mlx` |
| `sd3_5_large` | `stabilityai/stable-diffusion-3.5-large` | `SceneWorks/sd3.5-large-mlx` |
| `sd3_5_medium` | `stabilityai/stable-diffusion-3.5-medium` | `SceneWorks/sd3.5-medium-mlx` |
| `wan_2_2` | `Wan-AI/Wan2.2-TI2V-5B-Diffusers` | `SceneWorks/wan2.2-ti2v-5b-mlx` |
| `kolors` | `Kwai-Kolors/Kolors-diffusers` | `SceneWorks/kolors-mlx` |

Each likely trips the same false "not installed" (Kolors is structurally identical to SDXL). They need **per-target install-path validation** before the same change is applied blindly — Wan is a `requiresConversion` model and SD3.5 is native-MLX/Apple-only, so their install + resolve paths differ from the packed-turnkey SDXL/Kolors case. Tracked separately so this fix (the reported issue) isn't blocked on that audit.

## Note (not part of this fix)

Training requires the **`bf16/`** tier specifically. Image *generation* installs the **q4** tier by default, so a user who only installed SDXL for generation may — correctly, after this fix — still be told to install the training tier. That's expected, distinct from the false "not installed" this PR removes.

### DIFF
--- a/apps/rust-api/src/tests/training.rs
+++ b/apps/rust-api/src/tests/training.rs
@@ -2579,24 +2579,25 @@ fn sdxl_family_tiered_turnkey_trains_on_its_bf16_unet_tier() {
 }
 
 #[test]
-fn flat_sdxl_snapshot_is_not_treated_as_a_tiered_turnkey() {
-    // Backward-compat guard for sc-10613: `unet/` at the snapshot root marks a FLAT diffusers tree
-    // (stabilityai/stable-diffusion-xl-base-1.0, Kwai-Kolors/Kolors-diffusers). It must resolve
-    // unchanged, never descend into a `bf16/` subdir — even if a stray tier dir sits alongside it.
+fn flat_diffusers_snapshot_is_not_treated_as_a_tiered_turnkey() {
+    // Backward-compat guard for sc-10613: `unet/` at the snapshot root marks a FLAT diffusers tree.
+    // It must resolve unchanged, never descend into a `bf16/` subdir — even if a stray tier dir sits
+    // alongside it. Driven off a synthetic flat repo rather than a shipped target's `base_model_repo`
+    // so it exercises the resolution branch itself, independent of which targets ship a flat repo
+    // (the stock `sdxl` target moved to its tiered turnkey in issue #1694 — see
+    // `stock_sdxl_target_points_at_installed_turnkey`).
     let _env = isolate_hf_cache(); // seed under the tempdir, never a developer's real HF cache (sc-13834)
     let temp = tempfile::tempdir().expect("tempdir");
     let data_dir = temp.path().join("data");
 
-    let target = crate::builtin_training_targets()
+    // A real SDXL-family target struct with its repo overridden to a flat upstream diffusers repo,
+    // so only the resolution branch under test — not the shipped repo string — drives the assertion.
+    let mut target = crate::builtin_training_targets()
         .targets
         .into_iter()
         .find(|t| t.base_model == "sdxl")
         .expect("sdxl target");
-    assert_eq!(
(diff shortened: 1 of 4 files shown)


michaeltrefry:
The broader drift (the table above) is tracked as **sc-13860** — z_image_turbo / lens / sd3_5_large / sd3_5_medium / wan_2_2 / kolors, each with per-target install-path validation (conversion / MLX-only / packed-turnkey differ). This PR stays scoped to the reported SDXL issue.

PR #1795, opened 2026-07-24
fix(web): send the Simple UI's Text-mode reference image

michaeltrefry:
Follow-up to #1791. The Simple Image Studio's **Reference tile silently did nothing on the Text tab.**

It armed, showed the `SET` badge and a thumbnail, and toasted "Reference attached" — then the id was routed to `sourceAssetId`, which `buildImageJobRequest` discards for any non-edit mode. So the run ignored the reference entirely. And because `IMAGE_DEFAULTS` hardcoded `supportsImg2img: false`, no `advanced.strength` was emitted either. The worst failure mode for a control: it looked like it worked.

## The routing

The single armed reference is now routed by **mode**, in `simpleJobs.js`:

| mode | channel | strength |
|---|---|---|
| `edit_image` | `sourceAssetId` — the image being edited | none; the model's edit pipeline owns the conditioning |
| `text_to_image` | `img2imgReferenceAssetId` + `supportsImg2img` | `advanced.strength` |

Text-mode reference-guided generation is a per-model capability (`ui.img2img`). **13 catalog entries declare it** — Z-Image ×2, Krea 2 ×2, SD3.5 ×3, SANA ×2, Ideogram 4 ×2, Boogu ×2 — all with `ui.img2imgStrength.default: 0.5`.

Simple shows no strength slider, so `referenceStrengthFor` reads the model's **own declared default** (falling back to 0.5). A catalog entry that tunes it is honored with no code change here.

On a model *without* the capability the tile stays visible — keeping the design's 2-up tile grid — but renders **disabled and names the reason**, and an armed-but-unusable reference drops its `SET` badge and thumbnail so it can't read as active.

## Why #1791's tests missed this

Worth recording, because both gaps were mine:

- the byte-identical payload test only covered the **styleless, referenceless** path, so it never exercised the reference at all
- the browser pass ran against a **scratch data dir with no assets**, so there was nothing to attach and the tile was never armed

Added: a shell-level test that drives the **real tile through the real picker sheet** and asserts the resulting payload, plus payload tests for both modes, the unsupported-model drop, the "nothing armed" case, and the strength resolution.

**Verified by mutation** — reintroducing the old routing reds 3 of them, including the shell test, so they genuinely discriminate rather than asserting a default.

## Verification

- **2415 web tests pass** (172 files, 9 new), **0 lint errors**; re-run after merging `main` (3 commits behind at push time)
- **Confirmed on the wire** against a live API with a real uploaded asset, intercepting the actual `POST /image/jobs` body:

```json
{ "model": "z_image", "mode": "text_to_image",
  "referenceAssetId": "asset_8e45f0cd…", "sourceAssetId": null,
  "advanced": { "resolution": "1024x1024", "mlxQuantize": 8, "strength": 0.5 } }
```

- Also confirmed in-browser: selecting Anima 2B (no `ui.img2img`) disables the tile with "Anima 2B can't use a reference image — switch model to use one."

### DIFF
--- a/apps/web/src/simple/SimpleAudioStudio.jsx
+++ b/apps/web/src/simple/SimpleAudioStudio.jsx
@@ -9,9 +9,8 @@ import { useSimpleUi } from "./SimpleUiContext.js";
 import {
   Chips,
   DownloadButton,
-  JobFailureNotice,
   SheetSelect,
-  jobFailure,
+  StudioRunStatus,
   jobIsRunning,
   newestLocalJob,
 } from "./studioParts.jsx";
@@ -210,17 +209,13 @@ export function SimpleAudioStudio() {
         {busy ? "Generating…" : "Generate audio"}
       </button>
 
-      {jobFailure(latestJob) && !resultAssets.length ? (
-        <JobFailureNotice failure={jobFailure(latestJob)} />
-      ) : resultAssets.length ? (
+      {/* Live run strip: progress + Cancel + the outcome, right under Generate. */}
+      <StudioRunStatus job={latestJob} />
+
+      {resultAssets.length ? (
         <div className="su-audio-result">
           <AssetMedia asset={resultAssets[0]} />
           <DownloadButton asset={resultAssets[0]} className="su-icon-btn" />
-        </div>
-      ) : busy ? (
-        <div className="su-audio-result">
-          <span aria-hidden="true" className="su-spinner" />
-          <span className="su-card-note">Rendering audio…</span>
         </div>
       ) : null}
     </div>
--- a/apps/web/src/simple/SimpleAudioStudio.jsx
+++ b/apps/web/src/simple/SimpleAudioStudio.jsx
@@ -6,7 +6,15 @@ import { audioModelServesMode } from "../modelEligibility.js";
 import { resolveJobResultAssets } from "../jobResultAssets.js";
 import { buildSimpleAudioRequest } from "./simpleJobs.js";
 import { useSimpleUi } from "./SimpleUiContext.js";
-import { Chips, DownloadButton, SheetSelect, jobIsRunning, newestLocalJob } from "./studioParts.jsx";
+import {
+  Chips,
+  DownloadButton,
+  JobFailureNotice,
+  SheetSelect,
+  jobFailure,
+  jobIsRunning,
+  newestLocalJob,
+} from "./studioParts.jsx";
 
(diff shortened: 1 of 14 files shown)


michaeltrefry:
Extended this PR with two more Simple-UI misses you found — same class of defect (a control that renders but does not reach the worker, and a failure that never reaches the user).

## 1. Krea 2 edit runs always failed

`IMAGE_DEFAULTS` sent `loras: []` unconditionally. Krea 2's edit lane **rejects** a run carrying no `image_edit`-role LoRA — *"without it the source-image conditioning is inert. Select it in the LoRA picker."* Simple has no LoRA picker, so the run was unfixable from inside the UI.

The advanced studio already **manages** this LoRA rather than exposing it in the picker (epic 10871, sc-11069) — which is exactly the Simple behaviour; it just had to be applied. Simple now resolves it through the same shared `findModelEditLora` / `loraIsInstalled` helpers and serializes with `serializeLora` so `conditioningRole` round-trips (the worker reads it straight off the payload entry). `null` for edit models that need none (Qwen-Image-Edit, FLUX.2), so they are untouched.

When the LoRA is not downloaded, the run is **blocked in the UI with a one-click download**, instead of being submitted to fail in the worker.

## 2. A failed run showed nothing at all

The results zone returned `null` whenever a job produced no assets — precisely what a failed job does — so the reason was only reachable via Logs in the Advanced workspace. That is why the first bug was invisible.

The studios now render the worker's message **verbatim** (it is frequently actionable, as this case shows), and the Queue rows carry it too rather than a bare badge. A canceled job is also no longer badged "Failed" — the three error statuses share one bucket for colour/placement, not one label.

## Verification

- **2423 web tests pass** (172 files), 0 lint errors; re-run after merging `main` twice during this work
- **Mutation-verified**: reverting the LoRA routing reds 2 tests; restoring the results-zone guard reds the failure test. None is asserting a default.
- **Confirmed on the wire** against a live API — a Krea 2 Turbo edit now posts:

```json
"loras": [{ "id": "krea2_identity_edit", "conditioningRole": "image_edit",
            "installedPath": "…/krea2_identity_edit_v1_2.safetensors" }]
```

- and cancelling a run renders **"Canceled. Canceled before a worker started."** in the studio and on the Queue row.

Both of these were mine to catch and I did not. The common root cause with the reference bug is the same: I verified that controls *rendered* without verifying that they *reached the worker*. The Video and Audio studios have the same submit shape, so I would like to audit their payloads against their advanced counterparts field-by-field as a follow-up rather than wait for the next report.

PR #1859, opened 2026-07-26
perf(assets): remove row-scaled clean-list I/O (sc-14799)

michaeltrefry:
## Summary

- materialize generation-set hydration in the indexed asset envelope so clean lists perform no per-row generation-set reads
- store video poster bytes and SHA-256 digests atomically in schema-v8 `project.db` asset rows and serve them through a content-addressed API endpoint
- remove managed poster-directory copies and all clean-list poster enumeration, stats, and reads
- reject symlink/reparse-point, escaping, unreadable, malformed, and oversized poster sources; bounded poster payloads must fully decode as JPEG before indexing
- recompute the served blob digest and constant-time compare it with both the stored and requested digests, failing closed on row inconsistency
- reconcile poster lifecycle across overwrite, trash/purge, orphan repair, duplicate sidecar IDs, dirty-marker recovery, and ID/path reuse
- derive indexed URLs and returned `projectId` only from the authoritative route/store/reindex project id, never mutable imported sidecar metadata
- reuse one ready `project.db` connection for readiness, legacy-empty check, and the asset query; retain rollback journals for mounted-network-volume compatibility

## Acceptance criteria

- Clean-list inventory: 1-row and 128-row video listings have equal `total()`, with `sidecar_reads=0`, `generation_set_reads=0`, `poster_stats=0`, `poster_reads=0`, `directory_scans=0`, and `db_opens=1`.
- Enumeration guard: the regression statically rejects `read_dir`/`file_type` in the clean-list implementation so a row-scaled scan cannot be hidden behind one logical counter.
- Poster ingestion: indexing safely reads at most 16 MiB, fully decodes JPEG under dimension/allocation limits, and commits verified bytes, digest, and `asset_json` in one SQLite transaction. Filename alone never determines MIME.
- Poster serving: the content-addressed endpoint reads the exact authoritative asset row, recomputes SHA-256, and constant-time compares recomputed/stored/requested digests. Wrong digest, wrong project, wrong asset, and tampered row all fail closed.
- Freshness: adding, replacing, or deleting only a source sibling poster requires explicit reindex. Until then the indexed URL remains resolvable from the DB. Reindex clears missing/rejected sources atomically.
- Lifecycle: regressions cover real JPEG fixtures, malformed and oversized payloads, symlink sources and legacy destinations, same-ID clear/replace, purge and orphan reuse, duplicate sidecar IDs, and dirty-marker repair after a DB failure.
- Project identity: normal writes, character-reference writes, explicit reindex, legacy auto-reindex, and dirty repair pass an authoritative project id into hydration. A mismatched imported sidecar regression proves the current project owns the returned URL/envelope.
- SQLite tradeoff: one poster blob per video increases DB and rollback-journal write volume, bounded to 16 MiB each. Listing selects `asset_json` without materializing blobs. WAL remains unsuitable for mounted network filesystems SceneWorks supports.
- Benchmark: Windows local SSD release run with real JPEG fixtures, 500 videos / 500 distinct generation sets / 500 posters: first 11.716 ms (`fs_total=10`), steady 11.533 ms (`fs_total=7`); both have zero directory scans, generation reads, poster stats, and poster reads, with one DB open. Network/RunPod evidence remains assigned to SC-14789.

## Codex evidence

- `cargo fmt --all -- --check`
- `cargo clippy -p sceneworks-core -p sceneworks-rust-api --all-targets -- -D warnings`
- `cargo test -p sceneworks-core` — main suite 611 passed, 1 ignored; integration and doc suites passed
- all 8 poster regressions passed, including malformed/oversized validation and tampered-row integrity; constant-inventory and API project/asset scoping regressions passed
- `cargo test -p sceneworks-rust-api` — 381 passed, 2 ignored; one known unrelated Windows process-liveness test fails in this sandbox (`freshly spawned parent reads as dead`)
- `cargo build -p sceneworks-core -p sceneworks-rust-api`
- `cargo run --release -p sceneworks-core --example assets_list_benchmark -- synthetic 500 10`
- `git diff --check`

Shortcut: https://app.shortcut.com/trefry/story/14799

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6244,6 +6244,7 @@ version = "0.8.0"
 dependencies = [
  "directories",
  "getrandom 0.3.4",
+ "image",
  "imagesize",
  "libc",
  "mime_guess",
@@ -6252,6 +6253,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sha2",
+ "subtle",
  "tempfile",
  "tracing",
  "tracing-subscriber",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6245,6 +6245,7 @@ dependencies = [
  "directories",
  "getrandom 0.3.4",
  "imagesize",
+ "libc",
  "mime_guess",
  "parking_lot",
  "rusqlite",
--- a/apps/rust-api/src/assets.rs
+++ b/apps/rust-api/src/assets.rs
@@ -32,6 +32,7 @@ pub(crate) async fn list_assets(
         timeline_reads = filesystem_operations.timeline_reads,
         character_reads = filesystem_operations.character_reads,
         poster_stats = filesystem_operations.poster_stats,
+        poster_reads = filesystem_operations.poster_reads,
         index_marker_reads = filesystem_operations.index_marker_reads,
         index_marker_writes = filesystem_operations.index_marker_writes,
         index_marker_removes = filesystem_operations.index_marker_removes,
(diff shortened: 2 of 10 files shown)


michaeltrefry:
Codex: exceptional correction cycle complete at 50555da6. Poster ingestion now fully decodes bounded JPEG payloads; malformed/oversized inputs clear poster state. Serving recomputes SHA-256 and constant-time compares recomputed, stored, and requested digests; tampered row plus wrong project/asset regressions fail closed. Full core: 611 passed/1 ignored plus integration/docs. Full API: 381 passed/2 ignored with only the known unrelated Windows parent-liveness failure. Clippy/build/fmt green. Real-JPEG benchmark: 11.716 ms first, 11.533 ms steady; directory_scans=0, poster_stats=0, poster_reads=0, db_opens=1.

Issue #1858, opened 2026-07-25
FLUX.2 models remain “Incomplete” after repeated downloads — scheduler config missing


0770hk-hub:
### What happened?

Three FLUX.2 models remain marked **Incomplete** even after using **Fix** and downloading them multiple times:

- FLUX.2 [dev] — Q8
- FLUX.2 [klein] 9B — bf16
- FLUX.2 [klein] 9B-KV — bf16 and Q4

SceneWorks reports that the cached files are incomplete and identifies the missing item as:

Cached files are incomplete:
q4/scheduler/<config>
bf16/scheduler/<config>.
q4/scheduler/<config>.

### Steps to reproduce

1. Open SceneWorks.
2. Open the model manager/library.
3. Select one of the following:
   - FLUX.2 [dev]
   - FLUX.2 [klein] 9B
   - FLUX.2 [klein] 9B-KV
4. Select a supported quantization tier, such as Q4, Q8, or bf16.
5. Download the model and wait for the download to complete.
6. Observe that the model is marked **Incomplete**.
7. Click **Fix** and wait for the files to download again.
8. Reopen the model page or attempt to generate an image.
9. Observe that the model is still marked **Incomplete**, with a missing `scheduler/<config>` message.

The issue is reproducible after multiple download/Fix attempts.

### Platform

macOS (Apple Silicon / MLX) MacBook Pro M5 Max 128GB

### SceneWorks version

0.80

### Model / LoRA involved (if any)

Models involved:
- FLUX.2 [dev] — Q8
- FLUX.2 [klein] 9B — bf16
- FLUX.2 [klein] 9B-KV — bf16 and Q4

### Logs / screenshots

```shell

```

0770hk-hub:
Additional FLUX.2 dependency-loading failure:

The following models work correctly on the same machine:

- FLUX.1 [dev] — bf16
- FLUX.1 [schnell] — bf16

However, **FLUX.2 [klein] 9B True V2** fails immediately at 0 seconds with:

flux2_klein_9b load failed: tokenizer: No such file or directory (os error 2)

PR #1887, opened 2026-07-26
fix(models): repin the Mage q4 tiers after the precision-floor rebuild (sc-15071)

michaeltrefry:
App-side half of [sc-15071](https://app.shortcut.com/trefry/story/15071). Epic 14034. Pairs with **SceneWorks/inference#270**.

## Why

Mage-Flow's **q4** tier — `mlx.quantize: 4`, the shipped default on all six rows — rendered a repeating blue tiled texture instead of the prompt, so a user installing Mage out of the box got the broken tier. Root cause and fix are in inference#270: an 8-bit precision floor on the DiT's `norm_out.linear` (the output head's adaLN modulation) and on the Qwen3-VL LM's 36 decoder layers. q4 luma correlation against the bf16 render goes 0.329 → 0.877.

A **packed artifact skips load-time quantization**, so that fix does not reach anyone until the published q4 artifacts are rebuilt. This PR is the repin after that rebuild.

## What happened to the artifacts

All six variant mirrors and `SceneWorks/Mage-Flow-Components-mlx` had their `q4/` subdir regenerated with the fixed converter and re-uploaded. `q8/` and `bf16/` were not touched and are byte-identical inside the new commits.

Verified end to end, not just uploaded: the published q4 bytes were downloaded fresh into a clean dir and run through `a_packed_tier_renders_identically_to_load_time_quantization`, which reports **max_abs 0** against the fixed load-time path — and that path is the one measured rendering the cube correctly.

## The change

Revisions repinned on all six variant repos plus the shared components repo, and the two sizes that actually moved:

| row | before | after | why |
|---|---|---|---|
| variant q4 DiT | 2,316,856,860 | 2,326,294,167 | `norm_out.linear` at 8 bits (+9.4 MB) |
| shared q4 text encoder | 2,514,418,914 | 4,331,077,508 | q4 now shares q8's LM decoder layers |

`estimatedSizeBytes` for the variant DiT uses the largest of the six mirrors so the estimate never under-reports; that matches the existing convention of one representative number across the six rows. q4 VAE is unchanged (dense in every tier by design).

**`mlx.quantize: 4` stays the default on all six rows.** q4 is correct now, and it is still a genuine low-memory tier — 7.00 GB installed against q8's 9.45 GB, peaking at 7.87 GB of unified memory against q8's 10.11 GB. Changing the default away from q4 was the fallback the story allowed only if q4 could not be made correct; it can, so it wasn't taken.

The manifest comment block now records why the q4 text encoder is nearly the size of the q8 one, so the next reader does not "fix" the apparent anomaly.

## Tests

`pytest tests/ -m "not e2e and not parity"` (66 passed) — including `test_builtin_manifest_audit.py`, whose pinned `COMPONENTS_REVISION` / `DIT_BYTES` / `TE_BYTES` are updated here. `cargo test -p sceneworks-core --lib builtin_manifests` (18 passed), which covers `mage_flow_family_ships_six_variants_as_prequantized_per_tier_artifacts` and `corequisite_downloads_pin_a_full_sha_revision`. `cargo fmt --check`. No Rust source changed.

### DIFF
--- a/config/manifests/builtin.models.jsonc
+++ b/config/manifests/builtin.models.jsonc
@@ -40,20 +40,30 @@
       // The VAE is DENSE in every tier by design: its quantizable weights are the adaLN projections
       // the one-step decoder folds away at load, so packing them on disk would corrupt the decode.
       // It costs 0.345 GB per tier and is quantized at load exactly as before.
+      //
+      // sc-15071 — why the q4 text encoder (4.331 GB) is nearly the size of the q8 one (4.731 GB):
+      // packed uniformly at 4 bits the q4 tier did not render the prompt at all, it produced a
+      // repeating tiled texture. Two projections have an 8-bit floor — the DiT's `norm_out.linear`
+      // and the Qwen3-VL LM's 36 decoder layers — so q4 now shares q8's decoder layers and differs
+      // only in the token embedding and the vision tower. The q4 DiT grew 2.317 -> 2.326 GB for the
+      // same reason. `mlx.quantize: 4` stays the default because q4 is now correct; the full q4
+      // install is 7.00 GB against q8's 9.45 GB, and peaks at 7.87 GB of unified memory against
+      // 10.11 GB. All six q4 artifacts were regenerated and re-uploaded, which is what moved every
+      // `revision` below — the q8 and bf16 payloads in those commits are byte-identical to before.
         {
           "provider": "huggingface",
           "repo": "SceneWorks/Mage-Flow-Edit-Base",
-          "revision": "199b4cae841403e8a1cca3c585a09f10fc70d3f1",
+          "revision": "6c119cdac7ce7cf8c1ab4990d9c8ca18641f2c5d",
           "variant": "q4",
           "default": true,
           "files": ["q4/*"],
-          "estimatedSizeBytes": 2316856860,
-          "footprint": { "diskSizeBytes": 2316856860, "residentMemoryBytes": null, "peakMemoryBytes": null }
+          "estimatedSizeBytes": 2326294167,
+          "footprint": { "diskSizeBytes": 2326294167, "residentMemoryBytes": null, "peakMemoryBytes": null }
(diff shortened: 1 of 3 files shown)


michaeltrefry:
Two things surfaced while doing this, both filed rather than folded in:

- **sc-15111** — the Qwen3-VL **vision tower** still packs at the tier width with no rendering evidence at q4. It is Edit-variants-only and the new sc-15071 gate is a text-to-image probe that never loads it. Its MLP is the same `linear_fc1`/`linear_fc2` SwiGLU shape as the LM MLP that turned out to be the offender, so the prior is not neutral — but a floor without a measurement would be a guess, which is what got us here.

- **sc-15114** — ⚠️ **this repin makes NEW installs correct; it may not invalidate an EXISTING one.** The completeness probe is deliberately revision-agnostic (`min_by_key` over cached revisions, ~models.rs:4765 — the GH #1858 lesson), and `snapshot_matches_receipt` compares the receipt's own revision to what is on disk, not to the manifest. A user who already installed Mage q4 has a complete old snapshot that keeps reading as installed, and both old and new q4 snapshots are complete so the fold ties between them. Medium-high confidence from reading the code, not reproduced. Real exposure here is small — the per-tier Mage artifacts only reached main hours before the fix — but the mechanism is general and worth an explicit "this revision is superseded because the old bytes are defective" marker.

Also in this branch: `mage_flow_q8_mlx_smoke.rs` opened with "Mage's q4 tier does not render correctly", which was true when written and is now misleading. Corrected, with the real reason the smoke stays on q8 (it gates the split layout, which is tier-identical) and a pointer to the gate that does own tier correctness.

The `candle-worker` failure on the first push was Windows-runner infra — `rustc ... (never executed) / The directory name is invalid. (os error 267)` while compiling `candle-audio` from the pinned inference dep, on a diff that touched only a JSONC manifest and a Python test. candle-worker is green on recent main.

PR #1895, opened 2026-07-26
fix(worker): bump the inference pin and count staged components in the MLX fit gate (sc-15154)

michaeltrefry:
Companion to **SceneWorks/inference#279**. Both halves of **sc-15154**, plus the two side findings from the epic-14034 acceptance run.

⚠️ **Pinned at the inference PR head `078acac3`, not a commit on inference `main`.** Re-pin to the merge commit once #279 lands.

## 1. Inference pin `5491c85e` → `078acac3`

The app was pinned behind sc-15071's `afff48ce`, so it still carried `calibration_peak_gb(Q4) = 5.940` — the anchor measured on the tier that rendered a tiled texture — instead of the re-derived **7.868 GB**. It also predates the sc-15154 vision-tower fix, without which Mage EDIT cannot render at q4 or q8 at all.

All four pin sites + the lockfile, via `scripts/bump-inference.mjs`. `check-gen-core-skew.sh` and the mlx-rs alignment check both pass.

**Licensing re-audit (fails-closed).** The provenance population grew 540 → 547: inference sc-15138/#277 made `mlx-gen-krea-realtime`'s already-honest port descriptions match the scanner's marker vocabulary, *specifically* so the crate could not ship UNCLASSIFIED across a pin bump. Classified with the disposition that change states — `architecture-reimplementation-existing-terms` under the Krea AI terms already mapped for the sibling `mlx-gen-krea` area, no per-file exception. SceneWorks ships no Krea Realtime catalog row yet (epic 8431 is gated), so no weights are downloaded at this pin; the About component lands with the model row.

## 2. The MLX fit gate under-counted every caller-staged component

`spec_component_bytes` summed `.safetensors` under `spec.weights` only — but `LoadSpec::components` (epic 13657) are staged from a **different** snapshot. Mage-Flow's per-tier dir holds the DiT alone; its text encoder and VAE are bit-identical across the six variants and hosted once in a shared mirror.

| tier | gate saw | really installs |
|---|---|---|
| q4 | 2.17 GiB (DiT) | **6.52 GiB** (DiT 2.17 + TE 4.03 + VAE 0.32) |
| bf16 | 7.67 GiB (DiT) | **16.26 GiB** |

Two observed consequences, both from the acceptance run:

**(a) the refusal quoted a figure unrelated to the tier.** At a 3 GB cap the typed refusal correctly failed the job — no crash, worker idle, no `exit(-1)` — but named tier **q4** while quoting **"~20 GB"**. That is `2.17 GiB + HEADROOM_GB (18)`, i.e. a third of the tier plus a flat constant; it happened to land near bf16's 17.46 GB install, which is what made it read as the wrong tier's number. Fixed by the byte accounting, and the message now names the rejected tier's own weight bytes next to the peak:

> `mage_flow_edit needs ~25 GB of unified memory even at the smallest installed tier it can run (q4 — ~7 GB of weights plus headroom for activations and the OS) but this machine has ~3 GB.`

**(b) the admit boundary sat below the real q4 peak.** The permissive weights-fit floor (sc-12179) compares `max(te, rest)` against `budget − OS_RESERVE_GB`. With the DiT-only reading that was 2.17 GiB, so emulated caps of **both 6 and 7 GB admitted and rendered**. Corrected it is 4.03 GiB (the text encoder), so **6 GB now refuses and 7 GB admits sequentially** — the boundary straddles inference's measured 7.868 GB q4 peak instead of sitting under it.

The two are different instruments and the PR keeps them distinct: this gate bounds *wired weights* under a staged schedule, while `mlx_gen_mage::memory::ensure_generation_fits` bounds the *complete peak* against MLX's live limit and is the one anchored at 7.868 (`gate_fails_closed_for_unknown_or_insufficient_budget`: 6.0 → Err, 8.0 → Ok, at the newly pinned revision). They must not disagree about direction, and now don't.

The downtier probe builds its spec through a new `tier_probe_spec`, so it stages the same co-requisites the load does and cannot disagree with `apply_residency_policy`. Staging is best-effort — an unresolvable co-requisite leaves the bare spec and the real load fails the job with its own actionable error rather than this probe guessing. A component that resolves *inside* the weights dir is skipped (the scan already counted it).

The provider-side half — `component_footprint` resolving through `resolve_component_dirs` so the text-encoder split follows the staged dir — is in #279 and arrives with the pin.

## Tests

- `a_staged_component_counts_toward_the_tier_that_loads_it` — bare vs staged vs staged-inside-root, and asserts `te == 5_000`, which only holds if the pinned provider footprint also follows the staged dirs. **Mutation-proven**: with the staged sum removed it fails `left: 3000, right: 9000`.
- `the_mage_q4_admit_boundary_moves_with_the_real_install_size` — the manifest's published q4 byte sizes; 6 GB rejects, 7 GB is `Sequential`, and the pre-fix DiT-only reading is asserted to clear 6 GB so the numbers are visibly about the footprint, not the thresholds.

`cargo fmt --check` · `clippy --all-targets -D warnings` · `cargo test` (all lanes green) · `pytest` 66 unit + 14 parity · `npm run check` · `check-license-coverage` + `--self-test` · `scan-inference-provenance --compare` · `bump-inference --self-test`.


michaeltrefry:
[claude] **Merged `main` (`af3f64e5`) — sc-8442/#1894 landed a pin bump to `981c8050` with the same krea-realtime provenance classification this branch had reached independently.**

Resolved by taking **main's** audit wholesale — its evidence line is the better one, because it records *why* the crate falls under terms already mapped (Krea Realtime 14B is Wan 2.1 T2V 14B weight-for-weight) rather than deferring to the sibling area by analogy. Then re-pointed the pin, the scanner `REVISION` and the audit digest at this branch's inference head `c7c14ad4`, which **contains** `981c8050` plus the sc-15154 fixes.

So the licensing part of this PR is now a no-op delta over main: same 547 candidates, same disposition, only the revision and the per-file blob SHAs move (this branch edits `mlx-gen-{boogu,krea,mage}` sources, which changes their blobs). `scan-inference-provenance --compare` and `check-license-coverage` (+ `--self-test`) are green against the merged tree.

Re-verified after the merge: `cargo fmt --check`, `clippy --all-targets -D warnings`, full `cargo test` (24 suites, 0 failures), pytest 66 unit.

michaeltrefry:
[claude] ⚠️ **Merge order matters — do not merge this before SceneWorks/inference#279.**

The pin here is `c7c14ad4`, a commit on **#279's branch**, not on inference `main`. If #279 is squash-merged and its branch deleted, `c7c14ad4` becomes unreachable and this pin stops resolving — a red build on `main`, not a subtle one.

Sequence:
1. merge **inference#279**;
2. re-pin here to the resulting **`main`** commit — `node scripts/bump-inference.mjs --sha <merge-sha>`, then update `inferenceRevision` + `provenanceScan.revision` + `auditDigest` in `config/inference-third-party-source.json` and `REVISION` in `scripts/scan-inference-provenance.mjs`, re-run `scan-inference-provenance --write` and `check-license-coverage` (fails-closed, so it will tell you if you miss one);
3. merge this.

Since I opened the PR, #279 gained two commits — a rustfmt fix and a test-only assertion — so the pinned `c7c14ad4` is slightly behind its head. That delta is confined to `tests/` and doc formatting: no production `.rs` change, so the provenance inventory is byte-identical either way and nothing in this PR's behaviour depends on it. It is superseded by step 2 regardless, which is why I have not spent another CI cycle re-pinning to a commit that will not survive the merge.

michaeltrefry:
[claude] **CI green — all five lanes pass** on `7e642b59`: `web`, `build-windows`, `nax-worker`, `parity`, `candle-worker`.

The earlier `parity` + `candle-worker` red was mine and real, not a flake: the new test asserted the provider's text-encoder footprint split unconditionally, but the MLX media registry is compiled in on **macOS only** — off-macOS `footprint()` returns `None` and `resolve_text_encoder_bytes` correctly falls back to the diffusers subdir scan, which is 0 for a DiT-only tier dir. That assertion is now macOS-gated; the `total` assertions the test exists for stay unconditional on every platform.

Not merged (never auto-merge). See the merge-order note above — inference#279 first.

michaeltrefry:
[claude] ✅ **Re-pinned to merged inference `main` — the branch-commit hazard flagged above is now gone. CI green on all five lanes.**

`5491c85e` → **`1d80161ad9c4a86445725e4dab69ba7b460f4101`** (the squash-merge of #279 on `main`). All four pin sites + lockfile via `bump-inference.mjs`, plus the three fields the fails-closed licensing audit keys off (`inferenceRevision`, `provenanceScan.revision`, `auditDigest`) and the scanner's own `REVISION`. Also merged current `main` (`29c14800`, sc-8443 krea-realtime routing) first.

Population unchanged: **547 candidates, same dispositions** — only the per-file blob SHAs move, because the squash rewrote the commits touching `mlx-gen-{boogu,krea,mage}`.

## The pinned revision carries BOTH halves — verified in its tree, not assumed

The squash means my commits are no longer ancestors, so I read the merged tree directly rather than trusting the merge:

| what | where at `1d80161a` | present |
|---|---|---|
| `VisionTower::from_weights(w, cfg, prefix, group_size)` + `quantize` at `self.group_size` | `mlx-gen-boogu/src/vision/mod.rs` | ✅ |
| boogu / krea call sites passing `convert::QUANT_GROUP_SIZE` | their `loader.rs` | ✅ |
| mage passing `QUANT_GROUP_SIZE` (64) | `mlx-gen-mage/src/text_encoder/load.rs:182` | ✅ |
| `component_footprint` → `resolve_component_dirs` (staged components) | `mlx-gen-mage/src/model.rs:594` | ✅ |
| `packed_bits` naming the assumed group size | `mlx-gen/src/quant.rs:163` | ✅ |
| both new gates | `tests/edit_quant_real_weights.rs`, `convert.rs:557`, `model.rs:656` | ✅ |

## The boundary, re-asserted against the re-pinned build

The two sc-15154 tests now **print** their measurements (matching how the mage-side gate reports its scores), so a future failure is diagnosable from CI output alone. Observed at this pin:

```
q4 @ 6 GB cap: fixed=Reject { needed_gb: 24.52, available_gb: 6.0, staged_gb: 22.03 }   pre-fix(DiT-only)=Sequential
q4 @ 7 GB cap: fixed=Sequential                                                          pre-fix(DiT-only)=Sequential
mage_flow_edit split tier: bare total=3000  staged total=9000  staged te=5000
```

**6 GB refuses, 7 GB admits** — the acceptance-run symptom (both admitting) is gone, and the pre-fix DiT-only reading still clears 6 GB on the same line, so the change is visibly about the footprint rather than a moved threshold.

`staged te=5000` is the load-bearing one: it can only be non-zero if the **pinned** provider footprint follows the staged component dirs. That is direct evidence the bump picked up the inference-side accounting fix, not just the vision-tower one.

## Checks at the new pin

`cargo fmt --check` · `clippy --all-targets -D warnings` · `cargo test` **24 suites / 2567 tests, 0 failures** · pytest **66 unit + 14 parity** · web **lint + 2742 tests + build** · `check-license-coverage` + `--self-test` · `scan-inference-provenance --compare` · `bump-inference --self-test`.

CI on `c19c1f10`: `web`, `build-windows`, `nax-worker`, `parity`, `candle-worker` — **all pass**.

Not merged.

michaeltrefry:
[claude] **One thing found while checking for a third accounting gap before the app-level acceptance — filed as sc-15217, and it is NOT in this PR's blast radius.**

`MageFlow::generate` (inference `model.rs:450`) calls `ensure_generation_fits(...)` for **every** variant and only then branches on `is_edit()`, but `calibration_peak_gb`'s anchors are — per their own doc comment — **generation** measurements. The edit path additionally holds the Qwen3-VL vision tower and the VAE **encoder** (`VaePart::Both` vs `Decode`). Measured, not reasoned:

| tier | real edit peak @512² | generation anchor | shortfall |
|---|---|---|---|
| q4 | 8.013 GB | 7.868 GB | −0.145 GB |
| q8 | 10.422 GB | 10.012 GB | −0.410 GB |

That is the **inference-side complete-peak gate**, a different instrument from the `mlx_fit_gate` this PR corrects (wired weights under a staged schedule). Pre-existing since the Edit variants shipped; neither caused nor touched by sc-15154. I did not re-anchor it here — two data points at one geometry cannot re-anchor a gate whose value is that its numbers are measured, so sc-15217 scopes the tier/geometry sweep including the non-CFG Turbo variant and multi-reference edits.

**Not a blocker for merging this**, and not a blocker for the app-level q4/q8 edit re-run: the margin is 0.15–0.41 GB against a budget of `0.85 × machine RAM`, so it only surfaces if a run deliberately sits on the admit/reject boundary.

PR #1921, opened 2026-07-27
test(identity): add the klein angle-set A/B driver sc-8253 and sc-8278 need

ghost:
## Why

sc-8253 and sc-8278 both have merged, unit-tested implementations, and both are stuck In Review on the same thing: **neither acceptance is a code claim, it's a measured one.**

- sc-8253: *"klein identity from a non-square reference matches the square-crop baseline (≈0.59, not 0.47)"*
- sc-8278: *"angle-set ArcFace materially above the 0.38 baseline, approaching the spike's 0.73"*

Nothing in the tree could produce either number. `lora_eval_harness` scores images but nothing generated the klein angle set outside the full app, so the measurement was a manual UI session nobody had run in a month.

This adds `sc_8253_8278_identity_angle_ab` — an `#[ignore]` real-weight driver modelled on the sc-6702 pose A/B — so it's a repeatable command.

## Design: three arms, sharing a cell

| arm | reference | image_guidance | serves |
|---|---|---|---|
| `squish_off` | stretched to square (pre-sc-8253) | off | sc-8253 ⌐ |
| `square_off` | production `fit_engine_image` crop | off | sc-8253 ⌐ / sc-8278 ⌐ |
| `square_on` | same square crop | 1.5 (production default) | sc-8278 ⌐ |

One run, both stories. 3 arms × 11 angles × 2 seeds = 66, plus a spread arm = 88 generations.

**The squish arm is hand-built on purpose.** `fit_edit_references` is merged and default-on, so *square is production now*. This driver constructs `Conditioning` directly and never routes through that helper — so without manually stretching the reference, both sc-8253 arms would be byte-identical and the run would confidently report "no effect" for a fix that works. Conversely the square arm calls the real `fit_engine_image` rather than reimplementing the crop, so it A/Bs shipped behaviour rather than a copy of it.

**The spread arm is sc-6541's lesson made procedural.** That study's control swung 0.106 in identity across two seeds — larger than the effects it was measuring — and had to retract a marginal-significance claim as pseudoreplicated. Here extra seeds on one condition give a measured generation-noise floor, so deltas get compared against it rather than assumed to clear it.

## Validated end to end

One angle, one seed, through both generation and scoring (~56s/generation):

| arm | identity_cosine |
|---|---|
| `squish_off` | 0.255 |
| `square_off` | 0.476 |
| `square_on` | 0.646 |

**n=1, noise-dominated — this is not a result and must not be quoted as one.** What it establishes: the squish arm genuinely defeats the merged fix, and both effects point the predicted direction. The 11-angle sweep is the measurement.

## Scope honesty

Absolute numbers will **not** reproduce sc-8253's 0.469/0.597 or sc-8278's 0.38 — those used the sc-6599 harness reference, which isn't present locally. This measures **within-run deltas** on one reference at fixed seeds, which is what "does the fix recover identity" actually needs. Stated in the doc comment so a future reader doesn't misread a delta as a failed reproduction.

## Notes

- Weights resolve `AB_KLEIN_WEIGHTS` → app-managed model dir → HF cache. klein lives in the app dir locally, which the existing `hf_snapshot` helper alone doesn't find.
- Re-running skips existing PNGs, so an interrupted sweep resumes.
- Emits `<AB_OUT>/<arm>/<angle>_s<seed>.png`, a fitted `ref/reference.png`, and `prompts.json` in the `{stem: prompt}` shape `lora_eval_harness` already reads — scoring needs no glue.
- Test-only; no production code touched. fmt + clippy clean.

```
cargo test -p sceneworks-worker --lib -- --ignored --nocapture sc_8253_8278_identity_angle_ab
REF_DIR=$AB_OUT/ref GEN_DIR=$AB_OUT/<arm> PROMPTS_JSON=$AB_OUT/prompts.json EVAL_LABEL=<arm> \
  cargo test -p sceneworks-worker --lib -- --ignored --nocapture eval_lora_outputs
```

https://claude.ai/code/session_017GEwY6cdn9YRPhytYjmqYF

### DIFF
--- a/crates/sceneworks-worker/src/image_jobs/tests.rs
+++ b/crates/sceneworks-worker/src/image_jobs/tests.rs
@@ -12928,7 +12928,10 @@ fn the_pinned_runtime_exposes_the_identity_guard_free_finetuned_entrypoint() {
 /// unavailable reference subject. This measures within-run deltas on one reference at fixed seeds.
 ///
 /// Writes `<AB_OUT>/<arm>/<angle>_s<seed>.png`, `<AB_OUT>/ref/reference.png`, and a
-/// `<AB_OUT>/prompts.json` (`{stem: prompt}`) ready for `lora_eval_harness`:
+/// `<AB_OUT>/prompts.json` ready for `lora_eval_harness`. That file is keyed by **angle**, not by
+/// image stem: the harness groups outputs by `prompt_id` to compute `same_prompt_spread`, so a
+/// per-image key would give every output its own group of one and silently report `null` — no
+/// seed-variance measure, which is the whole point of the spread arm.
 ///
 /// ```text
 /// cargo test -p sceneworks-worker --lib -- --ignored --nocapture sc_8253_8278_identity_angle_ab
@@ -13078,7 +13081,7 @@ fn sc_8253_8278_identity_angle_ab() {
                 done += 1;
                 if path.exists() {
                     eprintln!("[sc-8253/8278] {done}/{total} skip {arm}/{stem} (exists)");
-                    prompts.insert(stem, prompt.clone());
+                    prompts.insert(angle.clone(), prompt.clone());
                     continue;
                 }
                 let cancel = gen_core::CancelFlag::new();
@@ -13104,7 +13107,7 @@ fn sc_8253_8278_identity_angle_ab() {
                     .unwrap()
                     .save(&path)
                     .unwrap();
-                prompts.insert(stem, prompt.clone());
+                prompts.insert(angle.clone(), prompt.clone());
                 eprintln!("[sc-8253/8278] {done}/{total} {arm}/{}", path.display());
             }
         }
--- a/crates/sceneworks-worker/src/image_jobs/tests.rs
+++ b/crates/sceneworks-worker/src/image_jobs/tests.rs
@@ -12905,3 +12905,220 @@ fn the_pinned_runtime_exposes_the_identity_guard_free_finetuned_entrypoint() {
          identity tensor, so that path can never serve one; got: {error}"
     );
 }
+
+/// sc-8253 + sc-8278 identity A/B over the FLUX.2-klein angle set — the measurement both
+/// stories' acceptances hang on, and the only thing keeping either In Review.
+///
+/// Three arms over one reference and one seed set, so the two stories share a cell:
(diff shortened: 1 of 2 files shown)


ghost:
## Folded in: the prompt_id fix — plus a latent harness bug it exposed

**The driver fix.** `prompts.json` was keyed by image stem (`{angle}_s{seed}`), so every output got a unique `prompt_id`, every `by_prompt` group had one member, and `mean_pairwise_cosine` needs two — `same_prompt_spread` returned `null` on every run. That is the harness's native seed-variance measure and the whole reason the spread arm exists. Now keyed by **angle**, so an angle's seeds share a `prompt_id`.

Verified against the completed sweep, no regeneration: `same_prompt_spread` **null → 0.0700**.

## The bug that surfaced

Keying by angle puts both `up` and `up_left` in the prompt map — and `up_left_s8003` is a legitimate `_`-delimited prefix match for **both**. The matcher took the *first* match over BTreeMap-sorted keys, so `up` won:

> an up-left render scored against the up prompt, and pooled into the up seed cohort

Nothing in the output would reveal it. Wrong prompt, entirely plausible number.

`match_prompt_key` now takes the **longest** match, and I extracted it out of the `#[ignore]`-gated `run_eval` into the testable section so the behaviour is pinned:

```rust
fn match_prompt_key<'a>(stem: &str, keys: impl Iterator<Item = &'a String>) -> Option<String> {
    keys.filter(|k| stem == k.as_str() || stem.starts_with(&format!("{k}_")))
        .max_by_key(|k| k.len())
        .cloned()
}
```

Test covers: nested keys resolve to the specific one (`up_left_s8003` → `up_left`), a short key still wins when it's the only real match (`up_s8003` → `up`), exact stems, non-matches, and that a key must be a *delimited* prefix so `up` cannot swallow `upward`.

This was **latent, not active** — sc-6541's prompt keys didn't nest, so nothing had tripped it. Any hierarchical prompt naming would have.

## Cross-check that the matching is actually right

Rescoring the same 44 images stem-keyed vs angle-keyed gives **byte-identical** `prompt_adherence_mean` (0.18769090650222722). Exact-match and longest-prefix-match resolve to the same prompt text — which is what should happen, and would not if the nested keys were being confused.

`lora_eval_harness` 6 tests pass; fmt + clippy clean.

PR #1949, opened 2026-07-28
SC-15508: harden Mac wired ceiling provenance

michaeltrefry:
## Summary
- build on merged #1948 with the actual configured `iogpu.wired_limit_mb` policy before legacy/default fallbacks
- fail closed on malformed values and checked MiB-to-byte overflow
- record the selected wired-ceiling source in `hardware.probe`
- add seven resolver unit tests and preserve/update the platform contract

## Runtime failure addressed
Authoritative Mac run 30364511777 passed MLX build/tests/Clippy, exact Qwen resolution, and exact inference checkout, then failed at adapter probe because this host has `iogpu.wired_limit_mb=0` and no `kern.memorystatus_wired_mem_limit`. The allocator-derived fallback now succeeds on the target M5 Max and reports `wired=mlx_default_memory_limit/1.5` with `wiredLimitBytes=87044670532`.

## Validation
- `cargo test --locked -p sceneworks-image-memory-adapter --features mlx --bin image-memory-mlx-adapter` (7/7)
- targeted adapter Clippy with `-D warnings`
- `node --test scripts/platform-review-contracts.test.mjs` (11/11)
- `npm run check` (46/46)
- `cargo fmt --all -- --check`
- `git diff --check`
- live adapter probe on Apple M5 Max / macOS 26.5.2
- independent adversarial review: PASS, high confidence, no findings

Final real-weight calibration rerun remains required after merge before SC-15508 closes.

### DIFF
--- a/crates/sceneworks-image-memory-adapter/src/bin/mlx.rs
+++ b/crates/sceneworks-image-memory-adapter/src/bin/mlx.rs
@@ -43,6 +43,67 @@ fn integer(value: &str, label: &str) -> Result<u64, String> {
         .map_err(|error| format!("parse {label}={value:?}: {error}"))
 }
 
+#[derive(Debug, PartialEq, Eq)]
+struct WiredLimit {
+    bytes: u64,
+    source: &'static str,
+}
+
+fn positive_integer(value: Option<&str>, label: &str) -> Result<Option<u64>, String> {
+    value
+        .map(|value| integer(value, label))
+        .transpose()
+        .map(|value| value.filter(|value| *value > 0))
+}
+
+fn resolve_wired_limit(
+    override_bytes: Option<&str>,
+    iogpu_limit_mb: Option<&str>,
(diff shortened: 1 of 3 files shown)


michaeltrefry:
Independent exact-head review of bfd003a12bda1e050f0fc06a360edd555d647d57: CHANGES_REQUIRED. (1) Capture get_memory_limit() once, pass that exact sample to resolve_wired_limit, and emit the same captured value as mlxMemoryLimitBytes; current code reads twice, so derived and recorded evidence are not structurally tied. (2) An explicitly present SCENEWORKS_MLX_WIRED_LIMIT_BYTES=0 must fail closed; only zero-valued host sysctls should mean unavailable/fall through. Add an explicit-zero regression test. IOGPU conversion/overflow, legacy fallback, checked /3*2, source provenance, DCO, and patch hygiene otherwise pass. Please push a normal signed-off correction and request fresh exact-head review.

PR #1955, opened 2026-07-29
chore(memory-strategy): rename the contract to the lane-neutral vocabulary (sc-15804)

michaeltrefry:
Implements [sc-15804](https://app.shortcut.com/trefry/story/15804) (epic [15448](https://app.shortcut.com/trefry/epic/15448)). **Paired with inference PR https://github.com/SceneWorks/inference/pull/324, now merged** — this branch pins the merged-`main` revision; see "Pin bump" at the bottom.

## What changed and why

Nothing in the SC-15448 memory-optimization ladder is image-specific. Rung 1 sheds a conditioning component, rung 2 bounds decoder scratch, rung 3 bounds attention, rung 4 bounds transformer residency — video and audio have all four. The `Image` prefix recorded which lane adopted the contract first, not what the contract covers. This is step 0 of the epic because it invalidates prior calibration evidence, and prior calibration evidence is currently empty.

**Epic scope does not change** — it remains the 53 image catalog entries. Only the vocabulary generalizes.

### Renamed

| kind | before → after |
| --- | --- |
| crate | `sceneworks-image-memory-adapter` → `sceneworks-memory-adapter` (dir, `Cargo.toml`, `Cargo.lock`, workspace `members`, `docker/rust.Dockerfile` ×8, both `[[bin]]` targets: `image-memory-{mlx,candle}-adapter` → `memory-{mlx,candle}-adapter`) |
| worker module | `sceneworks_worker::image_memory` → `sceneworks_worker::memory_strategy` |
| schemas | `packages/schemas/image-memory-{matrix,calibration}.schema.json` → `memory-{matrix,calibration}.schema.json` |
| generated | `docs/generated/image-memory-{matrix.json,matrix.md,calibration-evidence.json}` → `memory-*` (regenerated, not hand-edited) |
| config | `config/image-memory-calibration-plan.json` → `config/memory-calibration-plan.json` |
| scripts | `generate-image-memory-matrix.mjs`, `image-memory-calibration-harness.mjs`, both `.test.mjs`, and the four `scripts/fixtures/image-memory-provider-*.mjs` → `memory-*`; plus path/identifier citations in `bump-inference.mjs`, `platform-review-contracts.test.mjs`, `split-memory-strategy-stories.mjs` |
| npm | `generate:image-memory-matrix`, `check:image-memory-matrix`, `check:image-memory-calibration` → `generate:memory-matrix`, `check:memory-matrix`, `check:memory-calibration` (and the `check` aggregate) |
| CI | `macos-mlx.yml` dispatch input `run_image_memory_calibration` → `run_memory_calibration`, env `RUN_IMAGE_MEMORY_CALIBRATION` → `RUN_MEMORY_CALIBRATION`, artifact `image-memory-mlx-evidence-*` → `memory-mlx-evidence-*` |
| python | `tests/test_image_memory_matrix.py` → `tests/test_memory_matrix.py` |
| docs | `image-memory-strategy-contract.md` → `memory-strategy-contract.md`; `image-memory-calibration-harness.md` → `memory-calibration-harness.md`; §13 of `memory-strategy-pipeline-target-design.md` rewritten from a plan into a record |

The worker module is `memory_strategy`, not bare `memory` — the reason (the `mlx_gen::memory` collision) is recorded in the inference-side module doc comment; see the paired PR.

### The two content edits

1. Schema `$id` URLs → `https://sceneworks.ai/schemas/memory-{matrix,calibration}.schema.json`.
2. `harnessVersion` const → `"sceneworks-memory-v3"`, in **both** places the calibration schema asserts it *and* in `HARNESS_VERSION` in the harness.

### Stale-evidence rejection — asserted, not assumed

The `harnessVersion` const is the epic's stale-evidence gate, so this PR proves the bump actually invalidates prior-vocabulary records rather than assuming it. `scripts/memory-calibration-harness.test.mjs` adds `prior-vocabulary evidence is rejected as stale by the harnessVersion gate`, which:

- builds a record that is well-formed and **self-consistent in every other respect** — including its deterministic `recordId` recomputed over the old version, exactly as the v2 harness would have emitted it — so identity is not what fails;
- asserts `validateRecord` rejects it with `/invalid harnessVersion/` (naming the version, not some incidental field);
- asserts the schema const rejects both the envelope and the record (`$.harnessVersion` and `$.records[0].harnessVersion` both `value does not equal const`);
- asserts a stale record cannot be smuggled in under a current envelope;
- and includes a **control**: the identical record at the current version passes both gates.

Guarded against a false green by mutation: reverting `HARNESS_VERSION` and the schema const back to `sceneworks-image-memory-v2` turns this test (and the companion "promoted evidence bundle carries the current harnessVersion" test) red — verified locally, then restored.

### No manifest key changed — confirmed by diff, not assumed

A manifest key change without a matching `schema.json` update turns the parity lane red, so this was checked directly. The entire `config/manifests/builtin.models.jsonc` diff is **one `//` comment line**:

```diff
-          // sc-15449 shared image-memory selector provenance. The worker accepts these measured curves
+          // sc-15449 shared memory-strategy selector provenance. The worker accepts these measured curves
```

Zero keys, zero values. `packages/schemas/model-manifest.schema.json` changes one **description** string (`"Image-memory calibration ABI…"` → `"Memory-strategy calibration ABI…"`); its key names (`calibrationAbi`, `calibrationFingerprint`, `sceneWorksRevision`, …) were already lane-neutral and are untouched.

## Premise verification (against `origin/main` @ `34468b24`, before renaming)

The story makes counted claims. Each SceneWorks-side one was re-measured:

1. **"Zero `imageMemory*` keys in `config/manifests/builtin.models.jsonc` and zero in either schema" — CONFIRMED.** `git grep imageMemory` returns nothing for either. No manifest migration needed.
   **Drift:** exactly one `imageMemory` key existed anywhere in the repo — `generatedFrom.sources.imageMemory` in the **generated** `docs/generated/image-memory-matrix.json`, emitted from the `SOURCES` map in the generator. The matrix schema types `generatedFrom` as an unconstrained `{"type": "object"}`, so it is not a schema-bound key; it regenerates as `memoryStrategy`.
2. **"Exactly two content (non-identifier) edits: the two schema `$id` URLs and the `harnessVersion` const" — SUBSTANTIALLY CONFIRMED** (and the const is indeed asserted twice, as the story says).
   **Drift:** three further non-identifier edits exist that the story did not enumerate — the `model-manifest.schema.json` **description** string above, the one manifest `//` comment above, and the tracing event name `event = "image_memory_request_selected"` in `mlx_fit_gate.rs` (log output, not a serialized payload; nothing greps for it). Also the prose titles `"SceneWorks image memory-ladder matrix"` / `"# Generated image memory-ladder matrix"` and the `generated image memory matrix is stale` error string.
3. **"`sceneworks_worker::image_memory` and the `sceneworks-image-memory-adapter` crate exist" — CONFIRMED**, both at exactly those names.

### Drift the story did not anticipate: a real name collision

The story asserts the contract types can safely become bare `Memory*` because the existing bare `Memory*` names "are SAM2/model-bank concepts **in other crates**". That survey covered the inference repo only. **SceneWorks has its own pre-existing `pub(crate) struct MemoryBudget`** in `crates/sceneworks-worker/src/mlx_fit_gate.rs` — the static whole-machine GB admission floor — in the same module that imports the contract's budget type. `ImageMemoryBudget → MemoryBudget` therefore produced `E0255: the name MemoryBudget is defined multiple times`, and 50-odd downstream `E0560`/`E0609`/`E0382` errors as the local struct's uses re-resolved to the contract type.

Resolved by renaming the **local** one to `MlxMemoryBudget`, with a doc comment stating why the two are different things and why the contract type owns the bare name. Done by re-deriving `mlx_fit_gate.rs` from its pre-rename state — where gen-core's type was still unambiguously `ImageMemoryBudget` — disambiguating the local name there, and only then re-applying the rename, so a blind substitution could never conflate the two. The 11 genuine contract-type call sites in that file were checked against the pre-rename file line by line.

## Adversarial-review fixes (second pass)

1. **`vram_gate.rs:517` `error[E0063]`** — the exhaustive `MemoryStrategyParameters` literal was missing `transformer_window_component`, which inference `c25b016b` (SC-15794) made a required field. This turned `candle-worker` red and was invisible locally (see the cfg correction below).
2. **`sceneworks-memory-adapter/src/bin/candle.rs:561`** — the identical missing field, with **no** CI lane compiling that binary at all, so it would only have surfaced during this epic's own downstream Candle calibration work.
3. **CI gap closed** — `memory-candle-adapter` is now compile-checked on the self-hosted Windows CUDA lane (which already triggers on `crates/**`, so this costs no new trigger) and on the Linux candle lane.
4. **`docs/memory-strategy-pipeline-target-design.md` §14** — the mechanical rename ran over the very section that *argued for* the rename, leaving it in a pre-rename present tense ("The contract is image-named throughout: ~25 public `Memory*` types") alongside stale pending advice ("Rename before the calibration wave, not after"). Converted to past tense consistently with §13. The rename had also left **two** sections numbered `## 13.`; the portability section is renumbered `## 14.` and its one cross-reference updated.

### Why `TransformerComponent::Dit` — and why the encoding is rung-sensitive

`transformer_window_component: None` already means `Dit` via `unwrap_or_default()`, so `..Default::default()` would have silently picked a component scope at a memory-admission gate. The value is stated explicitly at both sites instead. `Dit` is the evidence-backed choice:

- `candle-gen-krea` declares `transformer_window_components` empty, which `validate_selected_parameters` reads as the pre-SC-15794 DiT-only behaviour — making `Dit` the *only* scope a Krea request may name explicitly.
- Every measured `strategyParameters` row in the manifest was collected against DiT block streaming.
- SC-15794's own verdict kept `Dit` as the published default: the text-encoder scope cut z-image conditioning 46.5% but moved the request peak 0.0%.

**The rung-sensitivity is load-bearing, and is why this is not a blanket `Some(Dit)`.** `krea_rung_parameters` serves all four rungs, but `validate_selected_parameters` rejects any `Some(..)` component below `BoundedTransformerResidency` as "irrelevant below its owning strategy rung". A blanket `Some(Dit)` would have failed `validate_selection` on three of the four candidates and silently downgraded their `staticImplementation` / `exactParameters` evidence verdicts — a worse bug than the one being fixed. So the scope is declared only on `StreamedBlocks` (rung 4) and left `None` — identical DiT meaning — on the three cheaper rungs. In `candle.rs` the strategy is unconditionally `BoundedTransformerResidency`, so `Some(Dit)` is stated there to make `validate_selection` actively check the scope against the pinned provider rather than silently default.

### How the two fixes were actually compiled

Neither site compiles on a Mac, so "it builds here" would have been worthless. Both were verified by a compile that genuinely includes the code, and each probe was **mutation-tested** to confirm it reproduces the original `E0063`:

- **`vram_gate.rs`** — temporarily made `mod vram_gate` cfg-neutral and un-gated its two sibling helpers (`payload::json_f64` and the `fit_gate` SVD constants, both already `#[cfg(any(test, …))]`) behind `--cfg sceneworks_compile_probe`, then `cargo check -p sceneworks-worker --lib`. Clean with the fix; deleting the field reproduces `vram_gate.rs:517: error[E0063]` exactly. Probe fully reverted — the authoritative verdict is the `candle-worker` check on this PR.
- **`candle.rs`** — `runtime_cuda::gen_core` and `runtime_macos::gen_core` re-export the *same* `gen_core` crate, so a temporary probe binary reproduced the literal against `runtime_macos::gen_core` and compiled it on macOS. Clean with the fix; removing the field reproduces the same `E0063`. Probe deleted.

## How this was verified

| command | result |
| --- | --- |
| `cargo fmt --all -- --check` | pass |
| `cargo clippy --all-targets -- -D warnings` | pass — **macOS host; see the cfg caveat below** |
| `cargo test` (workspace default members) | pass — **macOS host; see the cfg caveat below** |
| `cargo build` | pass (one pre-existing macOS linker `__eh_frame` note, not a code warning) — **macOS host; see the cfg caveat below** |
| `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --workspace` | **pre-existing failures, see below** |
| `cargo clippy -p sceneworks-desktop --all-targets -- -D warnings` | pass |
| `npm run check` (includes `check:memory-calibration`, `check:memory-matrix`, both renamed harness suites, `platform-review-contracts`) | 50 tests pass |
| `node scripts/generate-memory-matrix.mjs --check` | pass (artifacts regenerated and current) |
| `python -m pytest -q tests/ -m "not e2e and not parity"` | 81 passed, 19 deselected |
| `python -m pytest -q tests/test_memory_matrix.py` | 7 passed |
| `npm --prefix apps/web run test` (vitest) | 212 files, 2929 tests pass |

`apps/web` has **zero** files in this diff, so the web suite is a regression check rather than coverage of the change.

### Correction: what a macOS host does NOT compile

The three `cargo` rows above were first reported as unqualified passes. That overstated their coverage and hid two real `error[E0063]`s. A macOS host **never** compiles:

| excluded on macOS | gate | what actually compiles it |
| --- | --- | --- |
| `sceneworks-worker/src/vram_gate.rs`, `krea_control_fit.rs` | `#[cfg(all(not(target_os = "macos"), feature = "backend-candle"))]` — excluded at **every** feature combination on a Mac | `candle-worker` (windows-candle.yml, self-hosted CUDA) |
| `sceneworks-memory-adapter/src/bin/candle.rs` (`memory-candle-adapter`) | `compile_error!` on macOS + `required-features = ["candle"]` | **nothing**, before this PR |

Read those rows as "pass on the lanes a Mac can reach". The authoritative verdict for the candle-gated code is the `candle-worker` check on this PR.

### Pre-existing rustdoc failures, not from this change

The story's AC lists `rustdoc -D warnings`. This repo has **no rustdoc gate** — no `cargo doc` in any workflow, and no `doc` script in `package.json` — and running it anyway surfaces 41 pre-existing errors, all `private-intra-doc-links` / `broken-intra-doc-links`:

- 40 in `crates/sceneworks-core` (`video_request.rs`, `jobs_store*`, `project_store.rs`, `image_request.rs`, `observability.rs`, …) — **zero** `sceneworks-core` files are in this diff.
- 1 in `crates/sceneworks-worker/src/settings.rs`, also not in this diff. Scoping rustdoc to just the crates this PR changes (`-p sceneworks-worker -p sceneworks-memory-adapter`) reports that single pre-existing `settings.rs` error and nothing else — no renamed file produces a rustdoc diagnostic.

Reported rather than fixed, per "do not fix unrelated code". Worth a story if the AC's rustdoc gate is meant to become real.

## Pin bump — done, on a merged-`main` revision

SceneWorks consumes `gen_core::ImageMemory*` throughout the worker and the adapter, so this could not compile until the inference rename landed. inference PR #324 is **merged**, and the pin now sits on **`00b9a515f2a9ad878f7971a660b17234bcffbb67`** — the squash-merge commit on inference `main`, verified with `git merge-base --is-ancestor`, not the old feature-branch head `55c2e613`. Zero references to `55c2e613` remain anywhere in the tree. The re-pin is provably **content-neutral** — `55c2e613^{tree}` and `00b9a515^{tree}` are the same tree object (`fd8aef9c`), so the squash-merge changed the commit identity and nothing else.

The stale pin was also turning the `parity` lane red, because `check-license-coverage.mjs` fails closed when the audited revision and the Cargo pin disagree. The audit was **re-run against `00b9a515`, not restamped**:

- `config/inference-third-party-source.json` — `inferenceRevision`, `provenanceScan.revision`, `crateCoverage.revision`, `provenanceScan.populationSha256` and `auditDigest` all moved to `00b9a515`.
- **Re-audited rather than assumed:** zero `NOTICE`/`LICENSE-*` changes between the previously audited `b739a183` and `00b9a515`; zero changed lines containing `include_str!`/`include_bytes!` across that range; and all 20 recorded include sites still resolve to an include macro at the pinned revision.
- The provenance-candidate delta is exactly 6 rows, every one *same path, same normalized-marker hash, same disposition* — only the blob sha moved, because the rename edited those files' text. No new ported-source markers; no new crates (94 prefixes, `cratePopulationSha256` unchanged).
- `docs/generated/memory-matrix.{json,md}` regenerated, never hand-edited. Its large diff is entirely derived digests: the per-cell `sceneWorks` source-tree hash, `calibrationFingerprint`, the `vramGate` source `sha256`, and `inferenceRevision` — no structural or semantic cells changed.
- `node scripts/check-license-coverage.mjs` → clean (83/86 models covered; the 3 `UNDETERMINED` upstream notices are pre-existing). `node scripts/generate-memory-matrix.mjs --check` → clean.

Note the bump also moves `INFERENCE_RUNTIME_REVISION` (catalog semantic provenance), which is by design: it marks prior analyses stale so they re-run under the new runtime.

## Shortcut

The HARD AC task text on all **53** calibration stories (SC-15450…SC-15502) that cites `packages/schemas/image-memory-calibration.schema.json` was updated to the new path — only that substring, on that one task, with no change to story state, description, owners, task completion, or the separate BACKEND SCOPE task. Verified by re-fetching three of them.

The pre-existing `candle-gen-wan` byte-identity test failure found while verifying the inference half is filed as [sc-15943](https://app.shortcut.com/trefry/story/15943).


michaeltrefry:
Review correction (coordinator): the "Pin bump" section describes the regenerated `docs/generated/memory-matrix.json` diff as four derived-digest kinds. It is actually seven paths — the three unlisted are `$.cells[].evidenceRevision.inference` (6595), `$.generatedFrom.inferenceRevision` (1), and `$.generatedFrom.sources.cargo.sha256` (1). All three are derived provenance digests that necessarily move with a pin bump.

The verified invariant is the one that matters and it holds: **zero structural changes (no added/removed keys, no array-shape changes) and zero semantic/numeric/verdict cells changed** across all 15,316 leaf changes. Independently confirmed by adversarial re-review, along with the tree-identity of `55c2e613` and `00b9a515` (both `fd8aef9c`) and a full re-run of `scan-inference-provenance.mjs` reproducing both population SHAs.

PR #1962, opened 2026-07-30
chore(deps): bump the inference pin to ea18475f — Krea 2048² edit fixes [sc-16003]

michaeltrefry:
Picks up [inference #326](https://github.com/SceneWorks/inference/pull/326) — the fix for **a Krea 2 Turbo image edit at 2048² never finishing its first denoise step**. sc-16003.

## What was broken

The job sat on `step 1/8` indefinitely (one real desktop job ran **2 h 10 min** without completing step 1), GPU pinned at 100% SM / 300 W, **0 MB/s PCIe**. 1024²/1536² edits completed; 2048² **t2i was unaffected**; **other user LoRAs at 2048² were unaffected**.

`Adapter::residual` applied the plain-LoRA legs with `Tensor::broadcast_matmul`, which candle materializes via `.contiguous()`. `TextFusionTransformer::forward` reshapes the conditioning to `[b·n_tokens, num_layers, d]`, making the two `layerwise_blocks` the one site in the model where the batch dim **is** the token count — and `n_tokens` carries the edit path's vision tokens, `(ref_res/32)²` = 4107 at 2048². The managed rank-256 `krea2_identity_edit` LoRA is the only adapter targeting those blocks (16 of its 256 projections) and is mandatory for edits, so each of its 32 residual legs there expanded a 1.3 MB factor into a **5.4 GB** copy and then ran a 4107-batch `M = 12` GEMM against it — ~170 GB of allocate-and-write per denoise step, with `front_end` running every step.

## GPU-measured (RTX PRO 6000, bf16 tier, real weights + the identity LoRA)

| case | before | after |
|---|---|---|
| 2048² edit, step 1 | never finished | **44.5 s** |
| 2048² edit, no-adapter control | 55.5 s/step | unchanged |
| 1536² edit, per step | ~24 s | **14.1 s** |

An 8-step 2048² edit now lands around 6½ min end-to-end; 1536²/1024² edits get ~1.7× faster as a side effect (same waste, previously survivable).

## The bump

`dad9b0e0` → `2c646a6b` via `scripts/bump-inference.mjs --sha`, which moves all four SHA sites in lockstep: the root `[patch]` candle-kernels rev, `crates/sceneworks-worker/Cargo.toml`, `crates/sceneworks-memory-adapter/Cargo.toml`, plus `INFERENCE_PIN` and `INFERENCE_RUNTIME_REVISION`. The only inference delta in the range is #326 itself.

Verified locally:
- `check-gen-core-skew.sh` — exactly one `sceneworks-gen-core`, one `pmetal-mlx-rs` resolution.
- `cargo metadata --locked` consistent; zero stale `dad9b0e0` references anywhere.
- `candle_kernels_patch_guard` + `image_workspace_dep_guard` green (15 tests).
- `cargo fmt --all --check` and `cargo clippy --all-targets` clean.
- `scripts/check-candle-build.mjs` under CUDA is running; result posted as a comment.

Not covered here: two cancellation gaps this hang exposed (the worker only polls `cancel_requested_peek` inside the `GenEvent::Step` arm, and candle-gen checks the cancel flag only between denoise steps). They're documented on sc-16003 for separate stories — with this fix they cap cancel latency at one step instead of forever.


michaeltrefry:
Verification update:

**CUDA build check passed** at this pin — `scripts/check-candle-build.mjs` → `clippy -p sceneworks-worker --features backend-candle --all-targets -- -D warnings` is clean, test targets included.

**Re-verified the fix against current inference main**, not just the tree it was authored on. My original GPU numbers were measured against a stale local main; main has since gained Krea head-chunked attention and a streamed-blocks rung. Re-measured at this pin: a 2-step 2048² edit with the identity LoRA runs **41.9 s + 44.3 s** per step (98.5 s total incl. decode), so the fix holds — an 8-step 2048² edit lands around 7 min instead of never finishing.

**Heads-up on a follow-up re-bump:** [inference #330](https://github.com/SceneWorks/inference/pull/330) closes the cancel gap this hang exposed (`forward_edit` took no `CancelFlag`, so a mid-step cancel was unread until the step ended — and since `abort()` is inert on a `spawn_blocking` task, the worker abandoned the join and the GPU work kept running). Measured: cancel now lands in 1.2 s instead of ~36 s. Once #330 merges this pin wants a re-bump; happy to fold it in here or land it separately, whichever you prefer.

Correction to something I noted in the PR body: the worker side needed **no** change. It already polls `cancel_requested_peek` every 2 s on its own interval branch (not only on step events) and its teardown is already bounded with an abort fallback. The engine was the only missing link.

michaeltrefry:
**Re-bumped to `ea18475f`** now that [#330](https://github.com/SceneWorks/inference/pull/330) has merged, so this single commit carries **both** halves of sc-16003 rather than landing the LoRA fix and then chasing it with a second bump. Amended rather than stacked (no review comments on the diff yet) — the PR is still one mechanical commit, now at the final SHA.

Range `dad9b0e0..ea18475f` contains exactly the two sc-16003 PRs (#326, #330) and nothing else.

Re-verified at this pin:
- `check-gen-core-skew.sh` — one `sceneworks-gen-core`, one `pmetal-mlx-rs`
- `cargo metadata --locked` consistent; zero stale `2c646a6b` / `dad9b0e0` references
- `candle_kernels_patch_guard` + `image_workspace_dep_guard` — 15/15
- `cargo fmt --all --check` and `cargo clippy --all-targets` clean
- CUDA `check-candle-build.mjs` is running; result to follow

What lands for a user, GPU-measured on the bf16 tier with the identity LoRA:

| | before | after |
|---|---|---|
| 2048² edit, per step | never finished (2 h 10 min observed in-app) | **41.9 s / 44.3 s** |
| 2048² edit, mid-step cancel | ~36 s (rest of the step), then abandoned with GPU work still live | **1.2 s**, typed `Canceled` |
| 1536² edit, per step | ~24 s | **14.1 s** |

michaeltrefry:
CUDA check green at `ea18475f`: `scripts/check-candle-build.mjs` → `clippy -p sceneworks-worker --features backend-candle --all-targets -- -D warnings` typechecks clean, test targets included (1m 22s). That completes the local verification list for this pin.

michaeltrefry:
Fixed — that was my miss, not a flake.

`docs/generated/memory-matrix.{json,md}` is **derived from the inference pin**: the generator stamps `inferenceRevision` on the document and on every cell's `evidenceRevision`, and each `calibrationFingerprint` is a `sha256` over `{sceneWorksRevision, inferencePin, model, route, backend, tier, mode, overlay, rung, parameters}`. So bumping the pin makes the checked-in artifact stale by construction, and `tests/test_memory_matrix.py::test_generated_memory_matrix_is_current_and_schema_valid` failed with `generated memory matrix is stale`.

Regenerated it. The diff is large (~15k lines) but **pure provenance** — I checked that no cell's state or strategy moved, and the semantic summary is unchanged:

```
- Catalog entries: 53
- Cells: 6595
- MLX staged-residency static coverage: 39/53
```

Every non-stamp line in the JSON diff is a `calibrationFingerprint`, which is that sha256 over the pin.

**Also closed the gap that let it slip:** `bump-inference.mjs` now runs the regeneration itself. Leaving it to the caller meant *every* pin bump would fail parity CI this way — and it's the same argument for why the lockfile regen already lives in that script: what is derived from the pin should move with it, in one commit. Its `--self-test` and `--dry-run` both still pass.

Verified locally: `python -m pytest tests/test_memory_matrix.py` → 7 passed, and `npm run check:memory-matrix` clean.

michaeltrefry:
Rebased onto `5307afaa` — conflict resolved and CI re-triggered.

The conflict was only in `docs/generated/memory-matrix.{json,md}`. #1959 landed on main and regenerated the matrix too, and since `sceneWorksRevision` is a **source-tree hash**, any source change rewrites all ~15k lines — so two branches that both regenerate it will always collide. Textually merging a generated file is the wrong move, so I took main's version and re-ran the generator on the merged tree. It now records main's source hash `84d02dc4…` with my pin `ea18475f`, and `--check` agrees.

(Worth noting for future pin bumps: this conflict class is structural, not bad luck. Any PR that regenerates the matrix conflicts with any other, so a pin bump wants rebasing late — regenerate, don't merge.)

Re-verified everything on the rebased tree, not just the pieces that moved:
- `node scripts/generate-memory-matrix.mjs --check` clean; `pytest tests/test_memory_matrix.py` → 7 passed
- `candle_kernels_patch_guard` + `image_workspace_dep_guard` → 7 + 24 passed
- `cargo metadata --locked` consistent; zero stale `dad9b0e0` / `2c646a6b` references
- `cargo fmt --all --check` + `cargo clippy --all-targets` clean
- CUDA `check-candle-build.mjs` clean under `-D warnings` (re-run because main gained Rust changes in #1959)

Head is `f9b3b6a9`. Watching CI; will merge once green, as authorized.

michaeltrefry:
Second pin-derived artifact, now fixed — parity was failing the **license/provenance audit**, not the memory matrix this time:

```
- inference source audit is for dad9b0e0…, but Cargo pins ea18475f…
- ported-source inventory was scanned at dad9b0e0…
- crate-prefix inventory was scanned at dad9b0e0…
```

That guard fails closed when `config/inference-third-party-source.json`'s revisions don't equal the pin. Rescanned with `scan-inference-provenance.mjs` (which derives the revision from the pin itself, per sc-15017 — the label is never hand-maintained).

**Nothing needed classifying**, and I checked rather than assumed. The pin delta `dad9b0e0..ea18475f` is exactly my two PRs: 3 modified Rust files in existing crates, no adds/renames, no LICENSE/NOTICE changes, no new `include_str!`/`include_bytes!`. Accordingly:

- crate-prefix inventory **unchanged** — 94 prefixes, same population hash ⇒ no new crate
- candidate inventory moves for exactly those 3 files, all keeping their existing `architecture:` source area
- `portedSourceAreas`, `includeSites`, `crateDispositions` untouched; only revision labels, the candidate population hash and `auditDigest` move

Guard now passes locally: `OK — 83/86 shipped models covered by 60 components (3 undetermined upstream)` — those 3 UNDETERMINED lines are pre-existing and were present in the failing run too.

**`bump-inference.mjs` now runs the license guard too** (alongside the matrix regen). It *verifies* rather than regenerates, because a bump bringing a new crate needs a human decision and the rescan needs a local inference clone — but it fails closed at bump time with the scanner's own remediation text instead of costing a CI round-trip. Its `--self-test` passes and a full idempotent re-run of the bump exercises both new steps green.

One thing I caught and backed out: re-running the bump to exercise those steps had `cargo update` re-resolve **unrelated** transitive deps (`windows-sys 0.59.0`→`0.52.0`, `proc-macro-crate 1.3.1`→`2.0.2`). Reverted the lockfile to the state that already passed `build-windows`/`nax-worker`; `cargo metadata --locked` still clean and the pin is present 90×.

For the record, everything else was green on the previous run — `build-windows`, `nax-worker`, `web`, and `candle-worker` (its first report on this branch). Head is `3a82007d`.

Issue #1308, opened 2026-07-09
Dropping an image to a wrong place replaces the UI with the image


mhavu:
### What happened?

Dragging and dropping an image to a wrong place replaces the entire main window contents with the image.

### Steps to reproduce

1. Drag and drop an image in any place where a dropped image is not expected. 

### Platform

macOS (Apple Silicon / MLX)

### SceneWorks version

0.7.3

### Logs / screenshots

```shell

```

michaeltrefry:
Thank you! I can always count on other folks to try something I never would have! It will be fixed in the next release

PR #1979, opened 2026-07-30
feat(workflow-share): drop a shared image anywhere → "Workflow found" → prefill Image Studio (sc-15951)

michaeltrefry:
[sc-15951](https://app.shortcut.com/sceneworks/story/15951) — epic 15945.

The ComfyUI moment: someone sends you an image, you drag it onto SceneWorks, you are in Image Studio with their settings loaded.

## Where it hooks in

`useDropNavigationGuard` (issue #1308) already swallows drags no in-app dropzone claimed, and it already distinguishes claimed from unclaimed with one test: `event.defaultPrevented` is still false by the time the event reaches `window`, because every real dropzone calls `preventDefault()` as the event bubbles through it. sc-15951 hangs the inspector off **that same branch, unchanged** — it is inside the existing `if (event.defaultPrevented) return;` guard, so a drop AssetPicker / the Image Editor canvas / DatasetAddDialog / DocumentStudio / PoseLibraryScreen handled can never reach it.

Two further narrowings keep today's behaviour where an offer would be meaningless: only `drop` (never `dragover`), and only a drag carrying **exactly one** file — a text or URL drag carries none, and a multi-file drag is a bulk gesture no single-recipe offer could answer honestly.

`apps/web/src/workflowDropRouting.test.jsx` proves it by dispatching real bubbling drops through the two components the ACs name — the real `ImageEditSourcePickerField` upload dropzone (its `importAsset` still gets the file) and the real `ImageEditor` canvas — and asserting the inspector was never offered either. A third case mounts the same tree and drops on neutral chrome, where it is.

## The client prescreen

`POST /api/v1/workflows/inspect` stages the whole body to disk **before** it looks at the first byte, and a scan miss is O(file). So the client checks two things locally before sending anything:

- the **PNG signature**, read out of the first eight bytes via `Blob.slice` (a range read, not a load) — only a PNG can carry the chunk, so a 400 MB TIFF is refused here rather than uploaded to be told it is not a PNG;
- the endpoint's own **512 MiB cap**, so a body guaranteed to come back 413 is never sent.

`Blob.arrayBuffer` with a `FileReader` fallback, because jsdom and some older WebViews lack the former.

## What does NOT happen

`no_workflow` is the common case — every foreign PNG in the world — and it is byte-for-byte the pre-existing swallow: no panel, no spinner, no error band, nothing optimistic to un-render. Failures that say nothing about the file (a network drop, an axum plain-text multipart rejection with no `code`, a `not_png` that slipped the prescreen) are silent too. Only the closed set of "about your file" codes gets a sentence.

## Prefill goes through the existing seam

`launchImageRecipe` is extracted from `sendAssetRecipeToImage` and is now **the** path into the ImageStudio injection effect, shared by the viewer's "Use this recipe" and by "Use this workflow". `recipeFromWorkflowShare` adapts the envelope onto the recipe shape that seam already accepts (`normalizedSettings` / `rawAdapterSettings`). No second prefill path exists.

## Two things it refuses to do quietly

**Never silently substitute.** Only requirements in state `resolved` are prefilled. `installable` is deliberately not enough: `generationModelsForType` and `compatibleLoras` both drop rows that are not on disk, so seeding an installable id would set a control with no option behind it — silent loss wearing the costume of a successful replay. A model this install cannot resolve is left unset and the panel says, in as many words, that Image Studio keeps the model it is already on. Same rule strips `styleId`/`stylePrompt` when the style axis does not resolve, which makes the studio use the already-composed prompt instead of re-selecting a picker entry this catalog does not have.

**`count` is not replayed.** The envelope's `seed` identifies ONE image while `count` came from the batch the run requested, so a naive replay would reproduce the shared image as the first of an N-image batch — N-1 generations nobody asked for. The recipe always asks for one; the panel reports "this image was one of a batch of N" so the fact is stated rather than lost. (sc-15952's story comment raises the same question.)

`omitted` is load-bearing and rendered as such: a recipe whose LoRAs were dropped over the cap is not a LoRA-free recipe, and it serializes identically to one, so the marker is the only thing between the user and a confidently wrong replay.

## For sc-15952

`components/WorkflowResolutionReport.jsx` is a component of its own precisely so the missing-inputs panel renders the same report the same way rather than growing a second description of it. `workflowShare.js` (the adapter, the resolved-only readers, `workflowUnresolved`) is drop-agnostic too.

## Tests

- `apps/web/src/workflowShare.test.js` — 22 tests: prescreen (PNG / non-PNG / oversize / unreadable), the adapter, the resolved-only rule, count, the style round-trip, no-aliasing.
- `apps/web/src/hooks/useWorkflowDrop.test.jsx` — 15 tests: the workflow / no-workflow / not-a-PNG / disabled branches, the showable-vs-silent failure split, supersession, dismissal, launch, import.
- `apps/web/src/hooks/useDropNavigationGuard.test.jsx` — the 4 pre-existing plus 6 for the hand-off, including "never offers a drop a real dropzone claimed".
- `apps/web/src/workflowDropRouting.test.jsx` — the AssetPicker + ImageEditor regression, through the real components.
- `apps/web/src/components/WorkflowDropPanel.test.jsx` — 10 tests: missing model named, omitted collections not presented as absence, user-supplied inputs, the batch note, the three actions.

Full web suite: **218 files, 3160 tests, all passing.** `eslint src` clean, `npm run web:build` clean, `npm run check` clean. `cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings` clean; `cargo test --workspace --no-fail-fast` has the same 23 pre-existing rustls "No provider set" failures (sc-15337) and no others — the sc-15948 web-app coverage lint and the doc pins are green.

### DIFF
--- a/apps/web/src/App.jsx
+++ b/apps/web/src/App.jsx
@@ -29,6 +29,8 @@ import { usePersonTracks } from "./hooks/usePersonTracks.js";
 import { useTimelines } from "./hooks/useTimelines.js";
 import { useAccessGate } from "./hooks/useAccessGate.js";
 import { useDropNavigationGuard } from "./hooks/useDropNavigationGuard.js";
+import { useWorkflowDrop } from "./hooks/useWorkflowDrop.js";
+import { WorkflowDropPanel } from "./components/WorkflowDropPanel.jsx";
 import { useJobEvents } from "./hooks/useJobEvents.js";
 import { AppStaticContext, AppLiveContext } from "./context/AppContext.js";
 import { ScreenActiveContext } from "./context/ScreenActiveContext.js";
@@ -683,9 +685,9 @@ export function App() {
     saveToken,
     lockRemote,
   } = useAccessGate({ setError, pushNotice, dismissNoticeKind });
-  // Stop a file dropped outside a real dropzone from navigating the webview to
-  // the image and replacing the whole UI (issue #1308).
-  useDropNavigationGuard();
+  // The drop guard that stops a stray file from navigating the webview (issue #1308) is
+  // installed further down, next to `useWorkflowDrop` — it now hands the unclaimed file to
+  // the workflow inspector (sc-15951), which needs the active project and `importAsset`.
   const [theme, setTheme] = useState(readStoredTheme);
   // Apply a theme and persist it through the API. localStorage gives an instant
   // initial paint, but on the desktop shell the UI runs at the API's per-launch
@@ -2393,6 +2395,41 @@ export function App() {
     return asset?.generationSet?.recipe ?? asset?.recipe ?? null;
   }
 
+  // THE prefill path into Image Studio (sc-15951). Every caller that wants the studio's form
+  // populated from a recorded recipe goes through here — the viewer's "Use this recipe" below,
+  // and the dropped-image "Use this workflow" — so there is one launch shape, one hydration
+  // race guard, and one injection effect on the other side (`screens/ImageStudio.jsx`). A second
+  // parallel prefill would drift from this one the first time the recipe shape changed.
+  //
+  // `assetId` / `sourceAssetId` are null for a launch with no asset behind it, which is exactly
+  // the dropped-file case: the studio's edit branch then leaves its source picker empty rather
+  // than pointing at an image this install does not have.
+  async function launchImageRecipe({
+    recipe,
+    replaySeed = null,
+    assetId = null,
+    sourceAssetId = null,
+  }) {
+    if (!recipe) {
+      return false;
+    }
+    const projectId = activeProjectRef.current?.id;
+    setActiveView("Image");
+    const hydration = await stableHydrateLaunchDomains("Image", projectId);
+    if (hydration?.ok !== true || activeProjectRef.current?.id !== projectId) {
+      return false;
+    }
+    setStudioLaunch({
+      id: crypto.randomUUID(),
+      view: "Image",
+      assetId,
+      sourceAssetId,
+      recipe,
+      replaySeed,
+    });
+    return true;
+  }
+
   async function sendAssetRecipeToImage(asset, options = {}) {
     const recipe = recipeForAsset(asset);
     if (!asset || !recipe) {
@@ -2404,21 +2441,13 @@ export function App() {
     // Null → Image Studio leaves the seed random (a close variation), the default.
     const seed = assetSeed(asset);
     const replaySeed = options.keepSeed && seed != null && seed !== "" ? seed : null;
-    const projectId = activeProjectRef.current?.id;
     setSelectedAssetId(asset.id);
     closePreview();
-    setActiveView("Image");
-    const hydration = await stableHydrateLaunchDomains("Image", projectId);
-    if (hydration?.ok !== true || activeProjectRef.current?.id !== projectId) {
-      return;
-    }
-    setStudioLaunch({
-      id: crypto.randomUUID(),
-      view: "Image",
-      assetId: asset.id,
-      sourceAssetId: asset.lineage?.sourceAssetId ?? null,
+    await launchImageRecipe({
       recipe,
       replaySeed,
+      assetId: asset.id,
+      sourceAssetId: asset.lineage?.sourceAssetId ?? null,
     });
   }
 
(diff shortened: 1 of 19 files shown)


michaeltrefry:
## Review fixes — c026370

### The central finding: ten knobs displayed, none applied

`enhancePrompt`, `usePid`, `pidTarget`, `strength`, `textStyleGain`, `faceRestore`, `controlMode`, `controlScale`, `poses`, `phases` all travelled in the envelope, were rendered as ordinary "Settings" rows beside "Steps — 28", and reached no control.

**Eight are now wired** in the existing injection effect (`screens/ImageStudio.jsx`), which also fixes the recorded-recipe replay path. The booleans hydrate from `=== true` (`buildImageJobAdvanced` emits them only when on, so an absent key means the run had them off); the numbers keep the current value for an *absent* key but never for a zero one — `finiteRecipeNumber(null)` is `0`, and a 0.0 control strength is a real setting.

**The ordering trap has its own test.** `ImageStudio.recipeKnobs.test.jsx` replays a control recipe onto a *different* model whose declared `controlScale` default is 0.7, and asserts the recipe's 0.85 survives. The two `skip*` flags the effect already sets are what make it hold — verified, not read.

**`phases` is implemented.** `deserializePhases` is added to `imageMultiPhase.js` as the exact inverse of `serializePhases` (round-trip test), and `workflowPhases` first compacts the envelope's `loras[].index` onto the resolved subset: a 3-LoRA share whose first LoRA is missing would otherwise point every phase one slot right and silently activate the wrong adapter. A reference to a LoRA that did not resolve is dropped, never repointed.

**`poses` is sc-16132**, filed under epic 15945 with the ad-hoc-pose design decision spelled out — and it is no longer silent (below).

### The general fix: one table, both halves

`ADVANCED_PREFILL` (`workflowShare.js`) classifies every allow-listed key as reaching a control, riding in the prompt, or reaching nothing. `recipeFromWorkflowShare` carries exactly the first two; `workflowSettingRows` marks exactly the rest "not restored" with the reason written there. A key the table has never heard of defaults to *not restored* — rendered and marked, never silently dropped.

`workflowShare.test.js` reads the pinned `advanced-shared` table out of `docs/workflow-share-envelope.md` (itself pinned to `ADVANCED_KEY_RULES` in both directions by `workflow_share_doc.rs`) and fails if the two disagree either way. The doc's "adding a new advanced setting" section now says so.

### Successful transport is no longer punished

A pose selection **dropped** over the writer's cap lands in `omitted` and forces `runnable: false`. One that arrived **intact** was silent. `workflowNotRestored` folds both into the panel's verdict, alongside `cnScale` (Detail pass), `angleSet` (Character), `systemMessage` + `imageGuidanceScale` (Document Studio), and a **resolved** `stylePreset` — installed here and still reaching no control. The dead `recipe.stylePreset` write that claimed it is gone.

### "Use this workflow" in the Simple shell

It set the view and dismissed, and `SimpleShell` consumes no `studioLaunch` — so nothing happened and the launch sat queued to fire unprompted the next time the user opened the Advanced studio. `launchImageRecipe` now flips the shell the way `SimpleShell`'s own `openInAdvanced` does, and **refuses with a reason** on a viewport locked to Simple rather than closing over a button that did nothing. Both halves are tested end-to-end through the real App.

### Minors

- The verdict no longer counts input images the user supplies as things that "cannot be reproduced on this install" — mirroring `runnable`, which excludes them for the same reason. The rows are still rendered.
- `inspectWorkflowFile` now gets an `AbortController`; superseding, dismissing or unmounting aborts a 400 MB POST instead of ignoring its answer.
- The drop-routing regression suite covers all five real dropzones (`PoseLibraryScreen` renders `DatasetAddDialog`, so four cases cover five zones).

### Verification

Full web suite **3197 tests / 220 files green**, 37 of them new here. eslint clean; `npm run web:build` and `npm run check` green. `sceneworks-core` `workflow_share` (46) and `workflow_share_doc` (16) green — the doc edit does not disturb a pinned block.

`cargo test --workspace --no-fail-fast`: **23 failures, every one `No provider set`** — the sc-15337 rustls-provider baseline, in `sceneworks-mcp` and `sceneworks-rust-api --lib`, neither of which this PR touches. Not grown.

Drop routing, the prescreen and the no-workflow silence were re-run after every change and are unchanged. CI green: `web`, `parity`, `build-windows`.

PR #1987, opened 2026-07-30
feat(workflow-share): privacy — embed setting, first-run disclosure, and Save As without the workflow (sc-15953)

michaeltrefry:
Closes [sc-15953](https://app.shortcut.com/sceneworks/story/15953).

The user-facing half of epic 15945. sc-15948 reads `embedWorkflowInImages` at the write seam; this story owns the surfaces that write it, the disclosure that says it is on, and the two ways to share an image without the recipe.

## The setting

**Settings → Settings → Sharing** carries the toggle. It writes the preference the worker already re-reads at its PNG write seam on every job, so a flip lands on the next generation with no restart — `the_embed_toggle_and_its_disclosure_flag_persist_where_the_worker_looks` asserts that through `sceneworks_core::app_paths::embed_workflow_in_images` rather than through the route's own echo.

## The copy

The recurring defect on this epic is an artifact claiming a stronger guarantee than the code delivers, and settings copy is the most user-visible place that could land. So:

- The six path-exempt prose fields are **not written out as a sentence**. They are rendered from `EMBEDDED_PROSE_FIELDS` in `apps/web/src/workflowEmbed.js`, and `the_settings_copy_names_exactly_the_path_exempt_prose_fields` pins that list against the `prose-fields` table of `docs/workflow-share-envelope.md` in both directions. A seventh prose field added to the sanitizer fails the settings copy as loudly as it fails the doc.
- The copy names the **two-segment-path hole** (`Clients/Acme` is the shape of a HF repo id and is not treated as a location) rather than claiming every path is caught.
- It lists `producer.version` ("the SceneWorks version that wrote the file") and `count` ("how many images the run asked for") as things that travel.
- It says the switch applies **from the next generation** and that images already on disk keep the block they were written with — with a `not.toMatch` test against the wording that would imply a retroactive clean.
- It links to the contract doc, and that link is pinned too.

## First-run disclosure

**Trigger:** a generation accepted while embedding is on, hung off the `afterCreate` hook of the one image-job creator every studio goes through. Submission rather than completion, so the user is told before the files with their prompt in them exist.

**Flag:** `workflowEmbedNoticeSeen`, a durable server preference. Not localStorage — the desktop shell serves the UI from an origin that changes each launch, so a browser-side flag would re-show the notice forever. The merge is **one-way**: a payload carrying `false` is ignored, so a client PUTting a stale preferences object cannot make it reappear.

**Upgrade case:** absent means "not told". An install that has been generating for months has a silently-on default and has never been told, so the notice fires once for them rather than only for fresh installs.

## Strip: excise, not re-encode

`strip_workflow_chunk` walks the PNG chunk framing and drops our text chunks. IHDR, PLTE, every IDAT and IEND are copied through byte for byte, so the compressed image data is the same bytes it was — no encoder, filter or `image`/`png` version is in the path. Each chunk carries its own CRC over its own type and data, so nothing is recomputed.

`strip_of_an_embedded_png_is_byte_identical_to_the_none_path` proves it the strongest way available: stripping the `Some` output yields the **`None` output byte for byte**, over both DEFLATE regimes and four sizes, cross-checked against the independent hand-rolled stripper the write-path test already uses. No PSNR needed.

A PNG whose framing cannot be walked is an **error**, never the original bytes: "here is your copy without the workflow" must not be answered with a file that may still carry one.

## Egress

- **Desktop:** `save_asset_as` grows `without_workflow`. A source with nothing to strip goes through the same `std::fs::copy` the plain save uses. The source asset is never touched.
- **Browser/LAN:** `?stripWorkflow=true` on the existing file route. A query param on the path the media ticket's GET-only allow-list already covers, so remote/LAN works with **no change to the ticket rules** — a separate endpoint would have needed a new entry in the surface hardest to widen safely. `a_media_ticket_authorizes_the_stripped_download_too` pins it. The variant carries its own ETag (the route caches `immutable`, so a shared tag would let a cache answer a strip request with the unstripped body) and `Accept-Ranges: none`.

## Tests

Web: 228 files / 3312 tests pass, eslint clean, `web:build` clean. Rust: `sceneworks-core` and `sceneworks-rust-api` green, fmt + clippy clean, desktop crate compiles and its 102 tests pass. `cargo test --workspace --no-fail-fast` shows the same 23 pre-existing rustls `No provider set` failures (sc-15337) and no new ones.

### DIFF
--- a/apps/desktop/src/settings.rs
+++ b/apps/desktop/src/settings.rs
@@ -1090,12 +1090,20 @@ pub async fn save_image_export(
 /// user returns to their last-used save location (sc-8737); an empty/missing/non-existent
 /// hint is ignored. Returns `Ok(None)` when the user cancels the dialog and `Ok(Some(dest))`
 /// with the absolute destination path on success.
+///
+/// `without_workflow` (sc-15953) writes the copy with any embedded `sceneworks:workflow` chunk
+/// excised, for a user who wants to share the image but not the recipe that made it. It is a
+/// byte-level removal, not a re-encode — `copy_without_workflow_chunk` copies IHDR, every IDAT and
+/// IEND through unchanged — so the saved file has the same pixels as the same compressed data, and
+/// a source with nothing to strip goes through the same `std::fs::copy` the plain save uses. The
+/// SOURCE asset is never touched: this is one copy written differently, not a cleanup pass.
 #[tauri::command]
 pub async fn save_asset_as(
     app: AppHandle,
     source_path: String,
     suggested_filename: String,
     start_dir: Option<String>,
+    without_workflow: Option<bool>,
 ) -> Result<Option<String>, String> {
     // Guard the SOURCE before opening any dialog: only files inside the SceneWorks data
     // dir / HF cache may be copied out.
@@ -1117,7 +1125,12 @@ pub async fn save_asset_as(
         return Ok(None);
     };
 
-    std::fs::copy(&source, &destination).map_err(|error| error.to_string())?;
+    if without_workflow.unwrap_or(false) {
+        sceneworks_core::workflow_png::copy_without_workflow_chunk(&source, &destination)
+            .map_err(|error| error.to_string())?;
+    } else {
+        std::fs::copy(&source, &destination).map_err(|error| error.to_string())?;
+    }
     Ok(Some(destination.to_string_lossy().into_owned()))
 }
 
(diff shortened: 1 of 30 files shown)


michaeltrefry:
Follow-up in the same PR: the toggle is also in the Simple shell's Settings (`SimpleSettings.jsx`). Simple is the default shell on a phone, so leaving it out of that surface would have made "the setting is discoverable" untrue for most of the people the disclosure is aimed at. Same durable handler and the same `WorkflowEmbedDetails` block, so the two surfaces cannot drift in what they claim.

PR #1995, opened 2026-07-31
fix(mlx-fit-gate): hold the OS/app reserve fixed instead of scaling it with megapixels (sc-16195)

michaeltrefry:
Follows #1990 (sc-16194), now merged.

## What was wrong

`MlxRequestPlan::generic_total_peak_bytes` multiplied its **whole** flat activation allowance by the request's megapixels:

```rust
asset_bytes + activation_headroom_bytes * (w*h / 1024²).max(1.0) * batch
```

`activation_headroom_bytes` is `HEADROOM_GB` — a **1024²-only** calibration (sc-10863) — minus the legacy unified reserve. That constant's own doc comment decomposes it as *the max common-case transient (14.04) + a ~4 GiB macOS/app reserve*. Scaling the reserve is wrong on its face: the OS's working set does not grow because this request renders 2048² instead of 1024². The estimator couldn't tell the two halves apart because the split existed only in prose; this PR gives it a name (`OS_APP_RESERVE_GB`) and holds it fixed.

```text
before:  asset + 16·MP
after:   asset + 2 + 14·MP
```

**Bit-identical at the 1024² anchor for every family**, by construction — including lens-dense, whose 27.88 GiB allowance splits 2 + 25.88 and still totals 27.88 at 1 MP.

## The sweep — and the premise it refuted

The story asked for a resolution sweep and **expected it to justify a sublinear area term**, reasoning from two sc-5567 points (16× area → 3.8× memory, exponent ≈0.48) and calling linear-in-area "the wrong model".

New harness `resolution_sweep.rs` — the sibling of `footprint_measure.rs` with the axis rotated (one tier across resolutions, rather than one resolution across tiers) — measured **7 tiers / 5 families × 5 cells** on real weights via the in-process MLX counters (`ps` RSS cannot see the Metal high-water mark).

Normalised to each tier's own 1024² anchor:

| tier | 0.25 MP | 1.00 MP | 1.50 MP | 2.25 MP | 4.00 MP |
|---|---:|---:|---:|---:|---:|
| illustrious_xl_v1 q8 | 0.305 | 1.000 | 1.497 | 2.245 | 3.922 |
| sdxl bf16 | 0.305 | 1.000 | 1.497 | 2.245 | 3.922 |
| z_image_turbo q4 | 0.297 | 1.000 | 1.497 | 2.244 | 3.921 |
| lens q4 | 0.314 | 1.000 | 1.497 | 2.245 | 3.922 |
| qwen_image q8 | 0.511 | 1.000 | 1.498 | 2.246 | 3.994 |
| krea_2_turbo q8 | 0.511 | 1.000 | 1.498 | 2.245 | 3.992 |
| krea_2_turbo bf16 | 0.511 | 1.000 | 1.498 | 2.245 | 3.992 |
| **proportional-to-area** | *0.250* | *1.000* | *1.500* | *2.250* | *4.000* |

**Above 1024² the transient is proportional to area, max deviation 1.97%** — across an SDXL UNet, three single-stream DiTs, and three tiers whose VAE decode is tiled. Fitting the 0.48 exponent would have predicted illustrious q8 at 2048² as `14.04 × 4^0.48 = 27.3 GiB` against a measured **55.06** — a 28 GiB **under**-prediction on a gate whose permissive-side failure mode is an OS Jetsam SIGKILL.

So the area term is left linear **because it measured linear**, and only the fixed reserve was pulled out. Below 1024² the transient stops falling off proportionally (0.297–0.511 at 0.25× area), so the pre-existing `.max(1.0)` clamp is kept as the conservative reading of the data rather than removed as a leftover.

Two side findings, both load-bearing:
- **Tier-independent.** `sdxl bf16` matches `illustrious q8` to the byte at every cell; `krea bf16` matches `krea q8`. Activations are bf16 however the weights are encoded — so the story's "dense and packed" requirement has one answer per family, not two.
- **Instrument validated.** Three tiers reproduce their sc-10863 numbers exactly through the new harness (illustrious resident 4.74 / transient 14.04; qwen transient 7.66 / peak 41.11) before any new cell is trusted.

## Conservativeness (story requirement 4)

All 35 warm cells, `modeled − measured peak`:

| tier | 1024² | 1024×1536 | 1152×2048 | 2048² |
|---|---:|---:|---:|---:|
| sdxl bf16 | **+1.95** | +1.97 | +1.97 | +2.92 |
| z_image_turbo q4 | +2.07 | +2.09 | +2.09 | +3.04 |
| illustrious_xl_v1 q8 | +2.23 | +2.25 | +2.25 | +3.21 |
| lens q4 | +3.17 | +3.18 | +3.19 | +4.13 |
| krea_2_turbo q8 | +9.41 | +12.59 | +17.36 | +28.47 |
| krea_2_turbo bf16 | +9.50 | +12.68 | +17.45 | +28.56 |
| qwen_image q8 | +10.80 | +13.98 | +18.75 | +29.85 |

**Nothing is under-predicted.** The tightest margin is +1.95 GiB and it falls **at the 1024² anchor, where this change is a no-op** — so the margin the gate actually runs closest to is untouched. Only slack above the anchor shrinks, and it shrinks toward the measurement, not past it.

## What this does NOT fix — stated plainly

The story predicted krea-2-turbo bf16 at 1152×2048 would fit a 64 GiB Mac once corrected. **It does not** — and that exact cell is in the sweep, so this is measured rather than argued:

| krea-2-turbo bf16 @ 1152×2048 | GiB | on a 64 GiB Mac (62 GiB budget) |
|---|---:|---|
| **measured real peak** | **49.27** | would have completed |
| modeled after sc-16194 | 69.22 | rejects |
| modeled after this PR | 66.72 | **still rejects** |
| modeled with krea's *measured* anchor (7.67) | 52.45 | admits |

The over-rejection is real — the gate refuses a request that fits with 12.7 GiB to spare — but the resolution shape was never the dominant error. The dominant error is the **flat per-family anchor**: krea and qwen measure 7.66–7.67 against the generic 14.04, because their shared Qwen-Image VAE decodes tiled (sc-11747).

That is a different axis with a different risk profile. This PR is a pure shape correction that provably cannot lower the modeled peak below any measured point; lowering a family's anchor reduces a SIGKILL-guarding margin and must be argued family by family — including for the families this sweep did not reach, which must keep the conservative 14. Filed as **sc-16209** with these measurements attached, and `docs/sc-16195/README.md` §4 says so explicitly rather than leaving the story looking delivered.

## Verification

- New test `generic_request_peak_scales_the_area_term_but_never_the_os_reserve` pins absolute GiB at four cells rather than re-deriving the formula, so it cannot pass against a different split. **Mutation-checked**: zeroing the fixed reserve fails it (57.22 vs 56.22 at 1.5 MP).
- Full `npm run rust:check` green on current `main` — fmt, clippy, **3091 tests**, build.
- Method, all 42 measured rows, and the derivation in `docs/sc-16195/`.

Closes sc-16195.


michaeltrefry:
## Update — adversarial review found a real defect; fixed in 73801642

An independent review pass caught a correctness bug in the first commit that I had seen and talked myself out of. Recording it here because the reasoning matters more than the diff.

### The defect

`fixed_reserve_bytes` was unconditional, so it held 2 GiB out of **whatever** allowance a family carried. But the two constants filling that slot are different quantities:

| constant | what it is |
|---|---|
| `HEADROOM_GB` (18) | measured transient 14.04 **+ ~4 GiB OS reserve** |
| `LENS_DENSE_HEADROOM_GB` (29.88) | a measured transient **alone** (sc-11924) |

Lens dense has no reserve inside it to hold out, so the split moved 2 GiB out of its **area** term — `27.88·MP` became `2 + 25.88·MP`. That is strictly *less* conservative above 1024², by `2·(MP−1)`:

| MP | old margin | new margin (bug) | Δ |
|---:|---:|---:|---:|
| 1.00 | 0.00 | 0.00 | — |
| 1.50 | −0.91 | −1.91 | −1.00 |
| 2.25 | −2.34 | −4.84 | −2.50 |
| 4.00 | −3.66 | −9.66 | −6.00 |

On a path sc-11924 already records as under-predicting, whose permissive-side failure is a Jetsam SIGKILL. It made the estimator **worse on the one family the story named**, i.e. violated story item 4 — the exact thing this PR claims to honour.

My own doc comment had spotted the semantic mismatch and dismissed it: *"both keep exactly their current 1024² totals; only the >1024² slope changes."* The slope **is** the harm. That reasoning was wrong and is gone.

**Fix:** `spec_component_bytes` now returns `HeadroomAllowance { total_gb, os_reserve_gb }` rather than a bare `f64`, so the reserve travels with the allowance that actually contains one. Lens dense gets `os_reserve_gb: 0.0` and is an exact no-op. The change is now provably non-regressive for *every* family, not just the measured ones.

New test `a_bare_transient_allowance_keeps_its_whole_area_term` pins it at four cells with a generic-allowance control. Mutation-checked: giving `LENS_DENSE` an OS reserve fails it at 70.25 vs 69.25 at 1.5 MP — exactly the predicted `2·(MP−1)`.

### Harness gap

`footprint_measure` carries a `FOOTPRINT_RAN` guard (sc-8925) so an unfiltered run fails loudly instead of emitting corrupt rows. My sweep had seven tier tests and **no guard**, and its doc said "one tier per process is NOT required here" — true per *cell*, false per *tier*, and an invitation to that exact misuse. Added `SWEEP_RAN`; corrected the wording.

It now also asserts `get_cache_memory() <= resident/100` after the drain — the direct proof that `clear_cache()` emptied, which the whole `transient = peak − resident` derivation rests on. Story item 1 named this counter and I hadn't sampled it.

### Documentation corrections

- **Finding 4 was overstated.** Only `krea bf16` vs `krea q8` is a genuine within-model dense/packed pair (byte-identical). `sdxl bf16` vs `illustrious q8` is a cross-**model** comparison differing by up to 677 KB — informative about the architecture, not about tiers. And sc-10863's lens pair (14.04 vs 29.88) is a known 2× counterexample. Restated as *"holds where measured, re-measure per family"*, and sc-16209's tractability argument — which leaned on it — softened to match.
- **The cold row is a lower bound, not a ceiling.** Lazy materialization means weights and activations are never fully co-resident on the first pass; the cold peak measures *below* the warm peak at the same geometry in all 7 tiers.
- **Added the fresh-process cross-check** the in-process sweep needed: z_image_turbo q4 2048² re-measured alone reproduces the 5-cell sweep's row **byte-for-byte** (59,127,860,816).
- §5 said "35 cells" while tabulating 28 — the 7 omitted 512² cells are now stated with their range.
- §1's "the transient is not linear in area" was the *premise*, not a finding — relabelled.

### Also

- Test fixtures sat at exactly the 2 GiB reserve, leaving a **zero** area term after the split — fixtures that could never catch an area-scaling regression. Raised to 6 GiB.
- Regenerated `memory-matrix` / `calibration-cost-model` (the earlier `parity` red): `mlx_fit_gate.rs` is source-fingerprinted by both, and they run in the `parity` lane, not `rust:check`. Diff is **fingerprint-only — zero value changes**, confirming no load-time prediction moved.

`npm run check` (112) and `npm run rust:check` (3092) both green.

PR #2039, opened 2026-08-01
fix(sc-16402): use runner Python for Krea provisioning

michaeltrefry:
Avoid actions/setup-python installation on the self-hosted Windows service account, where the downloaded Python installer fails. Validate the runner-provided python and pip explicitly, then keep the existing pinned huggingface_hub install and exact-revision provisioning path.\n\nValidation:\n- git diff --check\n- actionlint passes after ignoring only the known custom cuda runner-label diagnostic\n- hosted failure isolated to actions/setup-python installation before provisioning

### DIFF
--- a/.github/workflows/windows-candle.yml
+++ b/.github/workflows/windows-candle.yml
@@ -219,10 +219,15 @@ jobs:
         if: ${{ github.event_name == 'workflow_dispatch' && inputs.run_five_rung_reference }}
         with:
           node-version: 22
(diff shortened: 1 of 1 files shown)


michaeltrefry:
Closing as superseded by df36472134498dd863c4e1f05e3b97f59f0dec52, which landed the same runner-Python fix on main with stricter Python >=3.12 validation and test coverage. The observed Windows CI failure was in the Candle sidecar build (os error 267), unrelated to this workflow-only diff.

Issue #1966, opened 2026-07-30
candle krea_2_turbo load failed: backend op failed: DriverError(CUDA_ERROR_SYSTEM_DRIVER_MISMATCH, "system has unsupported display driver / cuda driver combination")


airtonix:
### What happened?

<img width="1482" height="1173" alt="Image" src="https://github.com/user-attachments/assets/4e27c7b9-d09d-4e82-9615-ac62153c6dda" />

```
{
  "error": "candle krea_2_turbo load failed: backend op failed: DriverError(CUDA_ERROR_SYSTEM_DRIVER_MISMATCH, \"system has unsupported display driver / cuda driver combination\")",
  "event": "utility_job_failed",
  "jobId": "job_5e2cce818bd846f5f7f1ddeb5dec6575",
  "level": "error",
  "message": "Image generation failed.",
  "reportedAt": "2026-07-30T11:12:50Z"
}
```

(it's also stupidly annoying to copy this from the log tab)

```
~
❯ neofetch
                                            zenobius@brass
     %%%%%%====%%%%%%%%%%
   %%%%%%%%    %%%%%%%%%%%%%%             󱋩  bazzite-dx-nvidia-gnome:testing 
  %%%%%%%%%    %%%%%%%%%%%%%%%%           󰣛  Bazzite
  %%%%%%%%%    %%%%%%%%%%%%%%%###           Linux 7.1.3-ogc5.1.fc44.x86_64
  %%%%%%%%%    %%%%%%%%%%%%%######        󰅐  2 days, 7 hours, 21 mins
  ==                  =======######
  ==                  =========#####      󰻠  AMD Ryzen 7 5800X3D (16) @ 4.55 GHz
  %%%%%%%%%    %%%%%%%####======#####     󰍛  NVIDIA GeForce RTX 3090 Ti [Discrete]
  %%%%%%%%%    %%%%%#######=====#####       25.85 GiB / 62.69 GiB (41%)
  %%%%%%%%%    %%%#########=====#####       635.52 GiB / 928.93 GiB (68%) - btrfs
  %%%%%%%%%    %%##########=====#####       1.57 TiB / 1.79 TiB (88%) - ext4
  %%%%%%%%%====###########=====######       360.19 GiB / 457.37 GiB (79%) - ext4
   %%%%%%%%====#########======######      󰍹  3440x1440 in 34", 60 Hz [External]
[...]

PR #2080, opened 2026-08-02
fix(gpu): tell an NVML version mismatch to reboot, not reinstall (sc-17052)

michaeltrefry:
Closes the loop on [#1966](https://github.com/SceneWorks/SceneWorks/issues/1966) — the reporter answered both follow-up questions, and auditing the branch their answer *ruled out* turned up a real defect in it.

[sc-17052](https://app.shortcut.com/trefry/story/17052), under epic 2273. Follows [sc-16247](https://app.shortcut.com/trefry/story/16247) / [#2003](https://github.com/SceneWorks/SceneWorks/pull/2003), which is Done — so this couldn't ride on it.

## What the reporter's answer settled

They're on the **AppImage** with `nvidia-smi` **healthy**. That's a clean field confirmation of the NVML-vs-CUDA gap #2003 fixed: on Linux, `linux_cuda_preflight()` blocks startup when `nvidia-smi` fails, so the fact that they reached a generation at all proves NVML was answering normally while `cuInit` was failing with 803. Two libraries, two separate version checks — exactly the premise.

## The defect this surfaced

sc-16247 explicitly flagged the other branch as needing its own story:

> If `nvidia-smi` *also* failed → the existing preflight should have blocked startup […] That's a **second, separate hole** worth its own story.

It doesn't apply to this reporter, but checking it found that both desktop NVML preflights discard `nvidia-smi`'s stderr along with its exit status and collapse *every* failure into their "no usable GPU" text:

- `cuda_preflight()` (Windows) → `CUDA_REQUIREMENT`: "No NVIDIA GPU was detected…"
- `linux_cuda_preflight()` (Linux) → `LINUX_CUDA_REQUIREMENT`: "Install or update the NVIDIA driver…"

One `nvidia-smi` failure is not a missing driver. When NVML itself is version-mismatched it prints `Failed to initialize NVML: Driver/library version mismatch` and exits non-zero — that user's driver is installed and fine, the kernel module and userspace libraries are out of step, and **the remedy is a reboot**. Telling them to reinstall a driver that isn't broken is the wrong answer.

This is the NVML-side twin of `CUDA_ERROR_SYSTEM_DRIVER_MISMATCH` (803). So today the app gives the right answer when CUDA reports the mismatch and the wrong one when NVML does.

**The deeper probe can't save it.** `cuda_device_preflight` (sc-16247) knows the right answer, but the NVML gate is upstream of it in `run_startup` — deliberately, since it must fire before the multi-GB runtime download — and `return`s on error. The wrong message short-circuits the right one.

## Change

- `nvml_driver_mismatch(stderr) -> Option<&'static str>` — pure and unit-testable, matched on the `Driver/library version mismatch` tail alone. That's the stable half: the `Failed to initialize NVML:` prefix varies across driver branches, and newer builds append the detected versions on following lines.
- `NVML_DRIVER_MISMATCH`, worded to match `sceneworks-worker`'s `CUDA_DRIVER_MISMATCH` so the same host state reads the same fix wherever it's caught — including the immutable/atomic-distro caveat (Bazzite/Silverblue/Bluefin stage a driver update into the *next* boot, so uptime can be days). Duplicated rather than shared: the desktop shell doesn't link the worker crate, which is why the device probe is a sidecar spawn in the first place. A dependency edge for one string constant is the worse trade.
- Both platform preflights inspect stderr on a non-zero exit before falling back.
- Every other failure — absent binary, `Unknown Error`, `Insufficient Permissions` — keeps the existing verdict. Those really can mean the driver is missing or broken, and a reboot is not the fix.

## Verification

3 new tests. `cargo test -p sceneworks-desktop --bins` → **107 passed, 0 failed**. `cargo fmt` and `cargo clippy --all-targets` clean.

Tests cover both stderr forms (classic one-line and the newer one with appended version lines), assert the remedy names the reboot and the atomic-distro caveat, assert it does *not* repeat the install-a-driver text, and pin the four unrelated failure modes as unmatched.

## Honest gaps

**Not reproduced on hardware.** Faking a genuine NVML mismatch means breaking this box's driver install, which I won't do — the same limit #2003 recorded for 803. The signature is pinned by unit test against the verbatim string `nvidia-smi` emits.

**Linux compile is covered automatically now.** #2003 raised that nothing compiled the `#[cfg(target_os = "linux")]` desktop branch; that gap was since closed by sc-16269, which added `desktop-linux-check.yml` — a PR-triggered typecheck lane scoped to `apps/desktop/**`. This change sits in that path, so it compiles on this PR with no manual dispatch. Correcting my own note above.


michaeltrefry:
## CI: green except one pre-existing failure inherited from `main`

| lane | result |
|---|---|
| `check-linux` | **pass** (4m23s) — compiles the `#[cfg(target_os = "linux")]` branch this PR touches |
| `build-windows` | **pass** (8m54s) |
| `web` | pass |
| CodeQL (js/ts, python) | pass |
| `parity` | **fail — not from this PR** |

`parity` fails on `mlx_fit_gate::tests::production_control_spec_replaces_raw_checkpoint_with_one_typed_residency` (`crates/sceneworks-worker/src/mlx_fit_gate.rs:5606`). This PR touches only `apps/desktop/src/setup.rs`, and `apps/desktop` is outside the workspace `default-members` that lane builds — it cannot reach that crate.

Confirmed pre-existing rather than assumed: the identical test, file and line fails on `main` at `813d08e7f` — this branch's own base — in run [30769322751](https://github.com/SceneWorks/SceneWorks/actions/runs/30769322751). It also reproduces locally at this branch's HEAD. `main` went red on the merge of #2068.

Briefly, since it looks like a total-accounting bug and isn't: the first two assertions pass, so `predicted_peak_bytes` is correctly 5000. The failing one is `assert_eq!(breakdown.unattributed_bytes, BASE_SOURCE_BYTES)` — 5000 vs an expected 4000. The total is right; the **attribution** regressed, with the ControlBranch component's 1000 bytes falling into `unattributed` instead of a component. Worth fixing on its own merits — an entirely-unattributed peak would quietly degrade the memory breakdown — but it belongs in its own change, not folded in here.

Not merging on top of a red `main`; flagged separately.

michaeltrefry:
## CI green — the `parity` failure is resolved

It was the pre-existing red `main` described above, not this change. `main` has since been fixed by [sc-17037](https://app.shortcut.com/trefry/story/17037) (`845ae16f2`, *"decompose the typed peak from the contract, not the run context"*) — the same attribution bug diagnosed in the previous comment: the typed peak was decomposed against the run context rather than the contract, which is why the ControlBranch bytes landed in `unattributed`.

Merged latest `main` into this branch (`fc673c0d9`, clean — no conflicts). All lanes now pass:

| lane | result |
|---|---|
| `parity` | **pass** (7m40s) — previously the inherited failure |
| `build-windows` | pass (10m7s) |
| `check-linux` | pass (1m41s) |
| `check-macos` | pass (1m50s) |
| `web` | pass (2m33s) |
| CodeQL (js/ts, python) | pass |

Also verified locally on the merged tree: `cargo test -p sceneworks-desktop --bins` → 107 passed; `npm run check` → 202 passed; clippy + fmt clean.

Note that `check-macos` is new since this PR opened — the macOS desktop lane from sc-17026, which compiles `apps/desktop` under `-D warnings`. It's a third platform gate now. This change is already scoped `#[cfg(any(target_os = "windows", target_os = "linux"))]`, so it compiles out on macOS rather than tripping the dead-code error that lane caught in `gpu_runtime_dir`.

Ready for review.

PR #2094, opened 2026-08-03
fix(runpod): make the entrypoint bash-3.2-clean, retire the LINUX_ONLY exemption

michaeltrefry:
Follow-up to #2087. Two small commits.

## 1. The RunPod entrypoint could not be syntax-checked on a Mac

`scripts/check-runpod-supervisor.sh` executes `docker/runpod-entrypoint.sh` directly on the host to assert its fail-closed behavior. That entrypoint used `[[ -v VAR ]]` (bash 4.2), and macOS ships bash 3.2.57 as `/bin/bash`, where that is a **syntax error** — so the entrypoint exited 2 where the assertion expected 1, and the harness reported a misleading failure:

```
runpod supervisor self-test failed: public-v4-missing status was 2, expected 1
```

Not a production bug — the entrypoint only ever runs in the RunPod Linux image, where bash is 5.x. It was a coverage gap: `npm run check:docker:runpod:supervisor` could never pass on half the team's machines, and failed inscrutably rather than saying it was unsupported.

Fixed by making the entrypoint 3.2-clean (`[[ -n "${X+x}" ]]`), which lets the exemption go away entirely rather than papering over it.

**Verified on macOS 26.6 / bash 3.2.57:**

```
SceneWorks RunPod supervisor lifecycle self-test passed.
```

### Why `LINUX_ONLY` is now empty rather than removed

#2087 added `docker/runpod-entrypoint.sh` to that exemption set on the stated grounds that "no Mac ever runs the file." That reasoning was wrong, and the failure above is why: a script can ship in a Linux image and still be *executed on a developer's machine by a check in this repo*. The set is retained but empty, with a comment recording that "it ships in a container" is insufficient grounds for an entry, and that making a script 3.2-clean is preferable — an exempt script is one nobody can syntax-check locally.

The gate now reports `4 scripts; 4 held to bash-3.2 portability, 0 container-only`.

## 2. Pool inventory correction

#2087 recorded `nax-macos-2` as holding Qwen `bf16` only. True when written, not now: `q4` and `q8` have since been seeded at the same revision and verified (no broken symlinks, no `.incomplete` files). Both boxes are now symmetric, so any `qwen_tier` dispatch resolves on either.

The `weights` label was applied to `nax-macos-2` only *after* all three tiers verified, since the label promises a snapshot host and a bf16-only box would accept a `q4`/`q8` dispatch and fail it at resolve.

## Type

Bug Fix + Docs

## Testing

- [x] `bash scripts/check-runpod-supervisor.sh` passes on macOS 26.6 (bash 3.2.57) — previously impossible
- [x] `check:shell` + self-test (8 positive, 16 negative, 1 mixed), with the legacy `/bin/bash` 3.2 parse pass active
- [x] `check-scaffold`, `check-source-control-bytes` (1380 files)
- [ ] Reviewer check: confirm the entrypoint's `${X+x}` rewrite is semantically identical to `[[ -v ]]` for the blank-vs-unset distinction the surrounding code relies on

## Note

Branched fresh off `main` rather than continuing the #2087 branch — that branch predates the squash merge, so a PR from it would have proposed reverting the newer work now on `main`.

### DIFF
--- a/docker/runpod-entrypoint.sh
+++ b/docker/runpod-entrypoint.sh
@@ -69,7 +69,10 @@ ensure_writable_dir() {
 
 effective_api_host() {
   local host
-  if [[ -v SCENEWORKS_API_HOST ]]; then
+  # `${X+x}` rather than `[[ -v X ]]`: identical semantics for a scalar, but parses
+  # under bash 3.2, so scripts/check-runpod-supervisor.sh can exercise this
+  # entrypoint directly on a developer's Mac and not only inside the Linux image.
+  if [[ -n "${SCENEWORKS_API_HOST+x}" ]]; then
     host="$(trim_whitespace "${SCENEWORKS_API_HOST}")"
     # Match the API's env_string fallback for an explicitly blank value.
     [[ -n "${host}" ]] || host="127.0.0.1"
--- a/docs/rust-mlx-build.md
+++ b/docs/rust-mlx-build.md
@@ -94,11 +94,14 @@ in the pool can therefore hold different subsets. Keep track of which holds what
 | Box | Qwen tiers | Z-Image |
 | --- | --- | --- |
 | `nax-macos` | bf16, q4, q8 | q4 (plus the inference `real-weights` set) |
-| `nax-macos-2` | bf16 | q4 |
+| `nax-macos-2` | bf16, q4, q8 | q4 |
 
-A `qwen_tier: q4`/`q8` dispatch that lands on a box holding only `bf16` needs
-`provision_qwen_snapshot: true`, or pre-seed the missing tiers first (`q4` 26.4 GiB,
-`q8` 35.9 GiB).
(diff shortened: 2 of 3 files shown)


michaeltrefry:
Closing as a duplicate of #2093, which is the canonical fix — that PR came from the spawned task that actually did the work, and its two files are **byte-identical** to this one's:

```
IDENTICAL  docker/runpod-entrypoint.sh
IDENTICAL  scripts/check-shell-scripts.mjs
```

My error, and worth naming the cause: the task's commit was sitting in my worktree (it had committed onto the #2087 branch), and I re-packaged it onto a fresh branch off `main` without first checking whether it had already opened its own PR. It had, 12 minutes earlier. Reviewing #2093 instead.

The one thing here that is *not* in #2093 is an unrelated docs correction — the pool inventory in `docs/rust-mlx-build.md` still lists `nax-macos-2` as holding Qwen `bf16` only, which was accurate when written and is not now. Re-filed separately rather than bolted onto someone else's fix.

PR #2095, opened 2026-08-03
docs(runners): nax-macos-2 now holds every Qwen tier, not just the bf16 default

michaeltrefry:
## Summary

Docs-only. The nax pool's snapshot inventory in `docs/rust-mlx-build.md` (added in #2087) records `nax-macos-2` as holding Qwen `bf16` only. That was accurate when written and is not now — `q4` and `q8` have since been seeded at the same revision.

## Why it matters beyond bookkeeping

The `weights` label promises a *snapshot host*, and the resolve step matches an exact `/snapshots/<40-hex>/<tier>` path. A box holding only `bf16` will accept a `qwen_tier: q4`/`q8` dispatch and then fail it at resolve — after paying the full build. So the inventory is what tells an operator whether a dispatch will land somewhere that can serve it.

`weights` was applied to `nax-macos-2` only once all three tiers verified, for exactly that reason.

## Verified on the box

```
bf16:  54G  broken=0
q4:    26G  broken=0
q8:    36G  broken=0
incomplete files: 0
```

Both boxes are now symmetric, so any `qwen_tier` dispatch resolves on either. Reworded so that reads as a property of the current pool rather than a guarantee, and added per-tier sizes (`bf16` 53.8, `q4` 26.4, `q8` 35.9, Z-Image `q4` 5.5 GiB) so seeding a third box doesn't need a trip to the Hub API.

## Type

Docs

## Testing

- [x] `check-scaffold`, `check-source-control-bytes`

## Note

Split out of a now-closed #2094, which duplicated #2093's fix. This docs change was the only part of it not already covered there.

### DIFF
--- a/docs/rust-mlx-build.md
+++ b/docs/rust-mlx-build.md
@@ -94,11 +94,14 @@ in the pool can therefore hold different subsets. Keep track of which holds what
 | Box | Qwen tiers | Z-Image |
 | --- | --- | --- |
 | `nax-macos` | bf16, q4, q8 | q4 (plus the inference `real-weights` set) |
-| `nax-macos-2` | bf16 | q4 |
+| `nax-macos-2` | bf16, q4, q8 | q4 |
 
-A `qwen_tier: q4`/`q8` dispatch that lands on a box holding only `bf16` needs
-`provision_qwen_snapshot: true`, or pre-seed the missing tiers first (`q4` 26.4 GiB,
-`q8` 35.9 GiB).
+Both boxes currently hold every tier at revision `8080a417…` (Qwen) / `bb2bc989…`
+(Z-Image), so any `qwen_tier` dispatch resolves on either. That symmetry is a property of
+the current pool, not a guarantee — a box seeded with only the `bf16` default would fail a
+`q4`/`q8` dispatch at resolve unless it passed `provision_qwen_snapshot: true`. Sizes if
+you need to seed a new box: `bf16` 53.8 GiB, `q4` 26.4 GiB, `q8` 35.9 GiB, Z-Image `q4`
+5.5 GiB.
 
 This is why `release.yml` pins on `signing` rather than plain `nax`: it is tag-triggered
 with `cancel-in-progress: false`, so scheduling it onto an unsigned box means a broken


michaeltrefry:
## Scope grew — added the `rw-*` split and a `CARGO_TARGET_DIR` warning (`5364b85e`)

Since opening this, all 25 macOS `real-weights` jobs were moved to `nax-macos-2`, which made two more parts of this same doc stale. Folding them in here rather than opening a third docs PR.

**Label table.** `real-weights` no longer describes how the macOS weekly is routed — that side is now six per-family labels (`rw-audio` 8, `rw-sa3` 7, `rw-mage` 3, `rw-chroma` 3, `rw-krea` 2, `rw-llm` 2), and bare `real-weights` matches only the 10 Windows/CUDA jobs, so on a Mac it is **vestigial**. Also records two host requirements that a label cannot carry: `rw-mage` needs an operator-provisioned oracle bundle that exists on no Hub, and every `stabilityai/*` model is gated so the runner needs its own HF auth.

**Placement, and the real reason for it.** Documented as isolation rather than parallelism — the weekly is still serial on one box. What changed is that it is now serial on a box where nobody runs manual builds.

**The warning worth reviewing.** The "compile MLX once" section recommends:

```sh
export CARGO_TARGET_DIR=~/.cache/sceneworks-target
```

Good advice on a developer box; harmful on a machine that also hosts an Actions runner, and the doc recommended it with no caveat. Cargo takes an **exclusive lock per target directory**, so a manual build and a CI job pointed at the same one do not compete for CPU — they serialize. The CI job parks at `Blocking waiting for file lock on build directory` for as long as the manual build runs, which in the Actions UI reads as an inexplicably slow lane rather than as contention.

That is the mechanism behind manual agent runs blocking CI on `nax-macos`. The doc now names the safe combinations (share among your own worktrees and leave the runner on its default `_work/.../target`, or give the runner a distinct dir) rather than just prohibiting it.

### Verification

`check-scaffold`, `check-source-control-bytes` (1380 files). Docs-only, no behaviour.

### Not included

Whether to also move `nax` off `nax-macos` — that would leave it signing-only and stop agent builds contending with the MLX PR lane, at the cost of making `nax-macos-2` the sole `nax` runner. Open question, not a docs change.

Also unaddressed: `nax-macos` still carries the now-vestigial `real-weights` label. Harmless, but worth removing so the label list stops implying macOS jobs use it.

PR #2106, opened 2026-08-04
Closeout: bump the inference pin to the epic-16948 head and re-dump the capabilities facts (sc-17393)

michaeltrefry:
Closes the epic-16948 closeout ([sc-17393](https://app.shortcut.com/trefry/story/17393)). Moves every inference pin `277f4238` → **`35251a88`** (sc-16961's no-go carry-over record, inference #473 — the epic head), re-runs the full generated-artifact cascade, and closes two real holes in the bespoke-preview source guard.

## The anticipated compile break did not exist — and that was checked, not assumed

The story predicted a hard break across fourteen bespoke request literals. Main had already absorbed sc-16954…sc-16960 through intermediate bumps: `CandleStrictControl::generate_one` already takes `preview: &gen_core::PreviewSink`, and all fourteen call sites already bind a live sink. Each was audited individually rather than inferred from a green build — `preview: Default::default()` compiles, ships, and emits nothing, which is exactly what sc-16962 found already shipped at `krea_control_candle.rs`.

Two `_preview` discards were adjudicated rather than "fixed":

- `instantid.rs` — correct. `InstantIdRequest` has no `preview` field at this revision and no instantid route is advertised in the dump.
- `flux2_dev_gpu_smoke.rs`, `sdxl_edit_pid_gpu_smoke.rs` — correct. `#[ignore]`d hand-run harnesses, the same category as the already-exempt `tests.rs`; no user sees a card for them.

## The guard was covering 3 of 14 lanes

`WIRED_LANES` still listed only the three lanes from when sc-16962 wrote it. The generic sweeps are not a substitute: a sweep proves the absence of a bad *spelling*, not the presence of the field. **Mutation-proved** — deleting `preview: preview.clone()` from `flux2_edit_candle.rs` left both sweeps green and failed only the positive guard. Before this change that mutation was invisible.

The discard check had its own hole: it tested for `_preview,` (a closure parameter), but every strict-control lane receives the sink as a *typed* `generate_one` parameter, so discarding it reads `_preview:` and ends at a colon. Six of the fifteen listed files take the sink only that way and were guarded by a check that could not fail for them. **Mutation-proved** on `zimage_control.rs`. Both mutations reverted, all five guards green again.

## Cascade — each confirmed

| Artifact | How |
|---|---|
| Licence re-scan | Scanner re-run against a clone at `35251a88`: 623 candidates / `f119081d`, 94 crate prefixes / `0f90a97d` — both byte-identical, so the TSV and prefix list are unchanged. Range hand-audited for NOTICE / `LICENSE-*` / new non-test `include_str!`: none. `auditDigest` recomputed. |
| Lockfile + skew | `cargo update` via the script; exactly one `sceneworks-gen-core`, one `pmetal-mlx-rs`. |
| Memory matrix, calibration cost model | Regenerated by the script. |
| **Stage-1 candle capabilities** | **Genuinely re-dumped** on this CUDA box. **29** `supports_preview` route ids → **39** served model ids via the `engines.rs` join. |

### One honest exception: `capabilities.mlx.json` is restamped, not re-dumped

The MLX dumper needs macOS and the nax-macOS runners are **offline**, so no lane could produce it. The restamp is checkable in one command — `git diff --name-only 277f4238 35251a88` in inference returns exactly `candle-gen-catalog/src/lib.rs`, `candle-gen/src/preview.rs`, and one doc. No `mlx-gen` crate, no `gen-core`, no contracts file, so the MLX descriptors cannot differ between the two revisions. The diff here is the single revision line; all 334 lines of content are byte-identical. Re-dump on macOS when a runner returns — sc-17119 owns that lane.

## sc-17097 workaround not needed

Only `krea_2_turbo` stamps `calibrationAbi: 1` against `MEMORY_CALIBRATION_ABI = 3`, so that blocker is scoped to one model and nothing here depends on it. No local stamp edit was made.

## Tests

`npm run check` 243 ✓ · `cargo fmt --all --check` ✓ · `cargo clippy --all-targets -D warnings` ✓ · `cargo test` **2632 passed / 0 failed** ✓ · `cargo test -p sceneworks-worker --features backend-candle` **1256 passed / 0 failed** ✓ · candle clippy `-D warnings` ✓ · `cargo check -p sceneworks-rust-api --features backend-candle` ✓ · web vitest `previewSupportCatalog` 27 ✓


michaeltrefry:
## Windows/CUDA end-to-end evidence

Run on **2× RTX PRO 6000 Blackwell**, `CUDA_COMPUTE_CAP=120`, MSVC 14.44 BuildTools vcvars, `HF_HUB_CACHE=E:\huggingface\hub`, API on `:8010`. `expandable_segments` deliberately unset.

Both binaries **rebuilt from this branch head** (`4c6fa86f5`), not a stale release:
`cargo build --release -p sceneworks-rust-api` and `cargo build --release -p sceneworks-rust-worker --features sceneworks-worker/backend-candle`. Worker came up on GPU 0, GPU 1 and a cpu lane; API gated on health `status: "ok"`, not just HTTP 200.

Every sample reads **`job.result.previewFrame.dataUrl`** — the exact field `WorkerProgressCard.jsx:253` consumes. A harness reading `previewFrame.image` reports zero frames and is indistinguishable from an inert lane; that produced two false negatives in sc-16962. Frames are deduped on the **full** dataUrl.

| Coverage | Model → engine | Frames on `previewFrame.dataUrl` | Job |
|---|---|---|---|
| **Registry-driven** (sink threaded by the registry via `dyn Generator`) | `sana_1600m` → `sana_1600m` | **40** distinct, one per step, `1..40 of 40` | completed, 1 asset |
| **Rewired bespoke** (invoked by name; `SdxlEditRequest`, newly added to `WIRED_LANES`) | `realvisxl` + `edit_image` + `sourceAssetId` → **`sdxl_edit`** | **18** distinct, one per step, `1..18 of 18` | completed, 1 asset |
| **Unwired** (`supports_preview: false`) | `boogu_image_turbo` → `boogu_image_turbo` | **0**, renders cleanly, **no error** | completed, 1 asset |

The bespoke row is the load-bearing one: the worker log confirms `"engine":"sdxl_edit"`, i.e. the by-name lane the registry *cannot* thread a sink for — the exact class where `preview: Default::default()` would have looked green and emitted nothing.

JPEG payload sizes track the denoise as expected rather than repeating a constant frame — SDXL edit falls monotonically `16831B → 12563B` across its 18 steps; Sana runs `1491B` (noise) down to `1427B` mid-denoise and back up to `1711B` as detail resolves. 18 effective steps from 30 requested is correct img2img behaviour (strength × steps).

`previewFrame` is **absent** on all three completed jobs, which is the documented contract — the finished asset supersedes its interim preview (`base.rs:8767`).

### The served flag, end to end

`GET /api/v1/models` on the running stack returned the sc-16965 surface built from the newly dumped facts:

```
sana_1600m         preview.byBackend = {"candle":true,"mlx":true}
realvisxl          preview.byBackend = {"candle":true,"mlx":true}
boogu_image_turbo  preview.byBackend = {"candle":false,"mlx":false}
```

### sc-17097 not exercised

No `calibrationAbi` override was needed or made. Only `krea_2_turbo` stamps `calibrationAbi: 1` against `MEMORY_CALIBRATION_ABI = 3`, so the blocker is scoped to that single model and none of the three routes above depends on it. The worktree is clean.

PR #2109, opened 2026-08-04
chore(inference): bump the pin to 67bcbeb4 for the candle SDXL CFG-off fix (sc-17419)

michaeltrefry:
Closes [sc-17419](https://app.shortcut.com/trefry/story/17419). Delivers [sc-14195](https://app.shortcut.com/trefry/story/14195) / [inference#479](https://github.com/SceneWorks/inference/pull/479) to the app.

## What this unblocks

Until this lands, a candle SDXL image job with `advanced.guidanceScale = 1.0` still fails in the app with `generation failed: shape mismatch in matmul, lhs: [10, 4096, 64], rhs: [20, 64, 77]`. The engine fix is merged in inference, but it does not reach the worker until the SHA pin moves.

`277f4238 → 67bcbeb4`, which is **exactly** the merge commit of #479 — inference `main`'s head is that same SHA, so this bump carries the fix and nothing unrelated beyond what was already on main.

**No SceneWorks code change is needed.** `resolve_guidance` ([image_jobs/base.rs](crates/sceneworks-worker/src/image_jobs/base.rs)) forwards `advanced.guidanceScale` into `GenerationRequest::guidance` with no clamping, so `1.0` already arrived at the engine as `Some(1.0)`. The pin is the entire delivery.

## How it was done

Via `node scripts/bump-inference.mjs`, so the worker + memory-adapter manifests, the root `candle-kernels` `[patch]`, both `INFERENCE_PIN` / `INFERENCE_RUNTIME_REVISION` Rust constants, and `Cargo.lock` all move in lockstep. Gen-core and mlx-rs skew checks pass with one resolution each.

Pin-derived artifacts were **regenerated, not hand-stamped**:

- Provenance + crate-prefix inventories re-scanned at the new revision. 623 → 624 candidates; the new entry is `mlx-gen-chroma/src/convert.rs`, which arrived with upstream main and falls under an already-classified area. Crate population unchanged at 94.
- NOTICE / `LICENSE-*` files and every `include_str!`/`include_bytes!` call site diffed between the two revisions: **no changes**.
- Licence coverage passes (83/86 shipped models; the 3 `UNDETERMINED` are the pre-existing upstream ones).
- `capabilities.candle.json` re-dumped on this box — content unchanged apart from the revision stamp, which is the expected result: the fix changed behaviour, not the advertised capability surface.
- `builtin.preview-support.jsonc` + `previewSupport.json` regenerated (the latter came out byte-identical).

## Two things worth reading before approving

**1. The re-audit found two pre-existing stale line numbers.** `includeSites` cited `candle-gen-flux/src/pipeline.rs:982` (really **1475**, at both revisions) and `mlx-gen-chroma/src/loader.rs:22` (really 26 then, **25** now). Both were already wrong *before* this bump. Corrected here. `check-license-coverage.mjs` validates these entries structurally but never checks the line actually holds an include, so they drift silently — filed as [sc-17502](https://app.shortcut.com/trefry/story/17502). Nothing was mis-licensed: the `included` path and `disposition`, which are what drive the About→Licenses page, were correct in both cases.

**2. The generated matrix downgrades some statuses.** `flux2_dev` candle goes "Runtime verified" → "Implemented/unverified", current calibration runs 5 → 0. This is the designed consequence of moving the pin — verification evidence is revision-scoped, so prior runs no longer describe the pinned engine. The previous bump `653074924` shows the identical shape for `flux_dev`/`flux_schnell`.

## Verification

- `npm run rust:check` green — fmt + clippy + 533/689 tests + build.
- `npm run rust:check:candle` green under `-D warnings` against the new pin (the lane a pin bump is most likely to break; the default build compiles neither backend).
- The engine fix itself was GPU-validated in #479 against `SceneWorks/sdxl-base-mlx` q4 on an RTX PRO 6000, with a negative control reproducing the original error.

## Deliberately not done here

`capabilities.mlx.json` is still stamped at `277f4238`. It must be re-dumped on a Mac (`mlx-gen` does not build on Windows) and I left it alone rather than hand-restamping — editing the revision label without running the dumper is precisely the sc-15017 stale-but-green failure the provenance scanner's own header warns about. Filed as [sc-17501](https://app.shortcut.com/trefry/story/17501). Not a CI gate: the freshness check lives only in the bump script. Precedent: the prior off-Mac bump `f51aa6e3e` refreshed candle only, with the Mac dump following in `653074924`.

I also did not run the full API+worker E2E from the story's acceptance. The guidance passthrough is verified statically (no clamp in `resolve_guidance`) and the render itself is proven at the engine level on real weights in #479; standing up the local stack to re-prove it seemed disproportionate. Say the word if you want it run.


michaeltrefry:
Superseded — closing unmerged.

While this was in review, main advanced to `06e0c5e9` via the sc-17393 closeout and the sc-17439 chroma pin. That revision **already contains the sc-14195 CFG-off fix** (`67bcbeb4` is an ancestor of it, and `let ehs = if use_guide` is present in its `candle-gen-sdxl/src/pipeline.rs`), and both `capabilities.candle.json` and `capabilities.mlx.json` are current at it.

So the delivery this PR existed for has already landed by another route. Rebasing would leave an empty diff.

Two things this PR surfaced that outlived it:

- **sc-15833's compatibility audit used git-object identity over the shared `candle-gen` crate**, so the 42-line doc-only commit `35251a88` (sc-16961) de-authorized the FLUX.2 calibration evidence and failed `parity` + `candle-worker` here. Already fixed independently by sc-17497 ("audit the compiled artifact, not the source tree", #2112).
- **Two `includeSites` line numbers are stale** — `candle-gen-flux/src/pipeline.rs:982` (really 1475) and `mlx-gen-chroma/src/loader.rs:22` (really 25 at 06e0c5e9). They are **still wrong on main today**, having survived main's own re-audit at `06e0c5e9` — which is the evidence for [sc-17502](https://app.shortcut.com/trefry/story/17502): `check-license-coverage.mjs` validates these entries structurally but never opens the file, so they drift invisibly. The corrections in this PR are lost with it; sc-17502 carries them.

PR #2122, opened 2026-08-05
Hash the shipped bundle's resolved features, not just the closure (sc-17606)

michaeltrefry:
Closes [sc-17606](https://app.shortcut.com/trefry/story/17606).

## The gap

The FLUX.2 compatibility audit could not see a feature enabled by a crate **outside** the closure on a dependency FLUX.2 **shares**. The audited binary is built `-p candle-gen-flux2 --features cuda`, which unifies features over a strictly smaller graph than the shipped `-p runtime-cuda` bundle — and **feature flags are not recorded in `Cargo.lock`**, so enabling one that pulls no new optional dependency moves no lockfile entry, no crate tree, and no artifact digest. The record takes the free path and certifies. Realized precedent: `core-llm` unifying `serde_json/preserve_order`, which is on in this very bundle.

## What ships

A **third layer**: a resolved-feature witness over `runtime-cuda`'s resolution restricted to the packages `candle-gen-flux2` links.

```
cargo tree -p runtime-cuda -e normal,build --target x86_64-pc-windows-msvc \
           --prefix none --format "{p}|{f}" --locked
```

- **It runs on every audit, including the free path**, and both validators require it unconditionally. That asymmetry with the artifact layer is the design: the change it looks for moves no closure object, so it can only ever *arrive* on a record whose closure is quiet. A witness gated on a build would have been checked on precisely the runs where the thing it looks for cannot be.
- **Scoped, so it is not the ~50-crate trigger sc-17524 rejected.** The bundle resolves 243 packages; FLUX.2 links 173. A feature change confined to the other 70 moves nothing.
- Record schema is **v4**; v3 is refused rather than re-graded — a record that never looked at the shipped feature resolution cannot be read as evidence about it.
- The frozen expectation carries the **resolution as well as the digest**. A witness over another bundle, another triple, or with dev edges folded in answers a different question with an equally valid sha256.

## Measured, in both directions and end to end

| | witness |
| --- | --- |
| clean | `321625ed…` |
| `candle-llm` gains `serde_json/unbounded_depth` — outside the closure, reaches a shared dep | `bd19739d…` **moves** |
| `candle-audio-kokoro` gains a feature of its own — reaches nothing FLUX.2 links | `321625ed…` **unchanged** |

The first was committed to a real inference revision and driven through the whole tool. All ten closure objects came back byte-identical, so it took the **free path** — the run that would previously have printed *"no build needed"* and exited 0:

```
[audit] all 10 closure objects are byte-identical; no build needed.
[audit] SHIPPED FEATURE SETS DIFFER
  a4f409ae…: sha256:321625ed…
  9d6ba903…: sha256:bd19739d…
exit 1
```

The record is still written on that path, as it is for ARTIFACTS DIFFER — an operator needs the evidence, not just the verdict. The witness also re-derives identically from three different worktree paths (package locations are normalized out).

## The shipped record was re-emitted for real

A cuda-lane run on the RTX box against `5ffd7612 → a4f409ae`. The artifact digest reproduced `sha256:d80844f2…` — a thirteenth and fourteenth build of it — and against the surviving warm target directory the two builds took **59 s and 18 s**, not the ~25 minutes sc-17524 measured cold.

## What it does not close, said plainly

`measurementDelta` records how far the measured binary's resolution sits from the shipped one, computed rather than assumed: today exactly `candle-gen/testkit` and the testkit crate, both test-only. No *shipped* package resolves differently under the measurement build.

The divergence check deliberately does **not** fire when the shipped side moved. The delta is `measured − shipped`, so widening the shipped side widens the delta too — and reporting it there surfaced the primary finding under the wrong name and suppressed the record. That was found by running the tool end to end, not by reasoning about it.

One boundary stays open and is now measured rather than hand-waved: SceneWorks consumes `runtime-cuda` as a git dependency and unifies wider still — `image` (+bmp/gif/tiff/webp), `windows-sys`, `winapi` (+winsock2), `num-complex` (+std), `tracing-core` (+default). Filed as [sc-17639](https://app.shortcut.com/trefry/story/17639).

## Validation

- `npm run check` **298/298**; `npm run rust:check` **exit 0**, zero warnings; `cargo fmt --all -- --check` clean. All re-run after merging `main`.
- **21 hand mutations** (15 JS, 6 Rust) each confirmed RED — including deleting the witness requirement, dropping the empty-resolution guard, hashing the member tree's own features instead of the shipped ones, folding dev edges in, and each validator's individual field checks. One survivor was found and fixed: the Rust captured-digest comparison was unpinned by any test.
- The doc's "what that trade gives up" caveat is one item now, not two.

## Adversarial review pass

A fresh reviewer found five things; all are fixed in the second commit.

**The one that mattered.** `main` was never invoked by a test, and this story put the load-bearing half of the witness layer *inside* it — which revision each witness resolves at, whose digest lands on which side of the record, the exit code. Swapping `compatibleWitness.digest` for `capturedWitness.digest` made the layer permanently green with all 21 existing mutations still red. `main` now takes injectable runners and three tests drive it end to end over a fake that **models the checkout**; six wiring mutations go red. (The build fake originally keyed bytes to a call counter, which let "build the captured revision twice" survive — rewritten after that mutation lived.)

**`scopeFeatures` was ungraded** — declared in the record, hashed into the witness text, checked by neither validator, so a record could say `metal` and authorize. Frozen and checked in both languages now.

**The doc contradicted its own code twice**: it claimed the tool refuses to emit a record when the measurement delta diverges (it emits and exits 1), and claimed no shipped package resolves differently under the measurement build when `candle-gen` demonstrably does. Both corrected, plus two things previously left implicit — `measurementDelta` is descriptive rather than gated, and the 173 linked packages include `image` and `windows-sys`, so the re-capture trigger is narrow but not zero.

**Two smaller ones**: the cross-language pin read the Rust witness by first-match over the whole file (now scoped to the constant's initializer), and `main` creates a worktree on every run since this story, so a Ctrl-C leaves a registration behind routinely — SIGINT/SIGTERM now remove it.

Re-ran the audit for real after the refactor: **the emitted record is byte-identical to the checked-in one**, artifact digest `d80844f2…` again. `npm run check` 301/301, `npm run rust:check` exit 0, fmt clean.


michaeltrefry:
## ⚠️ Schema-version collision with #2123 (sc-17607)

Flagging from the other side: [#2123](https://github.com/SceneWorks/SceneWorks/pull/2123) opened in parallel and **also bumps `schemaVersion` 3 → 4**, with a different `FLUX2_V4_AUDIT_METHOD` string, and it **removes `crates/bundles/runtime-cuda` from the closure** (nine paths) on the grounds that no build of the audited target can adjudicate it. Both PRs touch the same twelve files, the same three frozen tables and the same record.

**They are complementary.** #2123 removes an unadjudicable closure member and replaces the composition question with an identity check against the linked bundle; it accepts one cost explicitly — with `runtime-cuda` out of the closure, `runtime-cuda/Cargo.toml` (where `default = ["media", "audio"]` and `candle-gen-catalog = { features = ["cuda"] }` live) is watched by nothing. **This PR's witness is what closes that**: `cargo tree -p runtime-cuda -e normal,build` sees a manifest feature edit directly. Together they cover composition by identity and feature resolution by witness.

**Proposed resolution** (same text is on #2123): whichever lands first keeps `4`; the second rebases to **`5`** and refuses `1..=4` — each changes the record's *shape*, and one version number describing two shapes is exactly what the version check exists to prevent. The second one through should re-emit the record with the tool rather than hand-merging it; `--workdir D:\repos\inference-audit\sc-17524` reuses the surviving `sc-17524-target` and made #2123's re-run two CUDA release builds in ~14 minutes, no GPU and no weights.

One doc section needs a hand merge either way: both PRs rewrite *"What that trade gives up"* in `docs/inference-artifact-audit-sc-17497.md` in opposite directions.

PR #2120, opened 2026-08-04
SC-15831: Implement FLUX.2 Klein shared image memory ladder

michaeltrefry:
Implements the shared Candle/CUDA image-memory ladder for FLUX.2 Klein, including exact selector admission, typed reference-mode bindings, sibling certification isolation, scheduler/worker routing parity, generated matrix/rung-4 updates, and the inference pin/provenance/capability refresh. Full core, worker, memory-adapter, clippy, generated-doc, license, and 119-test evidence gates pass. Independent final review: PASS (high confidence, no findings).

PR #2123, opened 2026-08-05
Take the two composition crates out of FLUX.2's codegen closure (sc-17607)

michaeltrefry:
Closes [sc-17607](https://app.shortcut.com/trefry/story/17607).

## The asymmetry

FLUX.2's audited closure contained `crates/bundles/runtime-cuda`, justified as *"the worker links it"*. But the bundle does not depend on `candle-gen-flux2` directly — the real edge is

```
runtime-cuda -> candle-gen-catalog -> candle-gen-flux2 -> candle-gen -> gen-core
```

`candle-gen-catalog` is the intermediate node, it is "linked by the worker" in exactly the same sense, and it was in **no** closure list at all: not `AUDIT_CLOSURE_CRATES`, not either frozen table. Across the certified window `5ffd7612 → a4f409ae` the catalog moved **988 lines** (+949/−39) while the audited bundle moved **nothing**. The crate the closure watched was inert; the one it ignored was the one that changed.

## The decision — the story's option 2

Both crates sit *above* the provider, so neither can change FLUX.2's compiled code and the measurement binary links neither. A digest can therefore neither convict nor clear either one. Keeping `runtime-cuda` audited bought no proof and guaranteed a false positive: a bundle edit registering some unrelated provider is unadjudicable by construction, and its only remedy was the ~47.6 GB re-capture this epic exists to stop demanding.

So the closure is now **exactly what the measurement binary compiles**, plus the inputs to that compile — nine paths, `schemaVersion: 4`. Every member is compile-covered, so no move can land without a digest that can answer for it. `COMPOSITION_ONLY_CRATES` names both excluded crates, and a test asserts they are in no closure list, so re-adding one is a deliberate act.

The catalog's original exclusion was not an oversight, which is worth keeping: sc-15833's v1 record dispositioned it explicitly in `changedPathDisposition` (*"Only SenseNova preview registration changed…"*) — a considered judgement in a free-text field no validator reads and that v2+ records do not carry. The reasoning existed and evaporated. This puts it where it fails when someone disagrees with it silently.

## The composition check that replaces it

`crates/sceneworks-worker/src/flux2_composition_audit.rs` asks the question a digest cannot, of the bundle the worker actually links — no weights, no GPU, on the `windows-candle` lane:

- the composed CUDA catalog's `flux2_dev` generator is `candle-gen-flux2`'s **own** registration — same `descriptor`, `load` and `footprint` function pointers as `candle_gen_flux2::register_providers` produces alone;
- so is its **memory-strategy** pair (`contract` + `safety_check`) and its memory-behavior fixture — the admission path SC-15833's rungs were actually measured against;
- negative controls: the comparison must be able to say *no* (`flux2_dev` vs its sibling `flux2_klein_9b` from the same crate), the registry must be the full CUDA bundle, and `providers::flux2` must really resolve to `candle_gen_flux2` (read off `type_name`, which no re-export can rewrite).

Identity rather than a frozen descriptor snapshot: a snapshot re-asserts what the digest already covers and goes red on every innocuous field addition, while identity pins the one fact the digest cannot reach.

## What this genuinely costs

The closure entry was a *tree* object, so it also covered `crates/bundles/runtime-cuda/Cargo.toml` — where the shipped feature resolution is decided (`candle-gen-catalog = { features = ["cuda"] }`, `default = ["media", "audio"]`). That is now watched by nothing: invisible to the digest (the measurement binary never links the bundle), to `Cargo.lock` (feature selections aren't recorded there), to the root virtual manifest, and to the composition check (the same function under a different unification is pointer-identical). Until [sc-17606](https://app.shortcut.com/trefry/story/17606) lands the resolved-feature-set witness, a `runtime-cuda` feature edit is silent.

The lost coverage was accidental and one-directional, which is why it does not change the decision: the path was never in `adjudicates`, so a move in it could only ever produce a *refusal*, never an authorization — and that refusal fired just as readily on a provider being added to the bundle. What is gone is a tripwire that could only say "re-capture", for a question a re-capture cannot answer. This is stated in the doc rather than glossed.

## Validation (RTX PRO 6000, Windows/CUDA)

- **The record was re-emitted by re-running the audit end to end** — `node scripts/inference-artifact-audit.mjs --lane cuda`, two real CUDA release builds — not hand-edited. Exit 0, digest `d80844f2…` reproduced, and the output is byte-identical to the deterministic regeneration in this diff. That is the 13th and 14th build to produce that digest, across four separate runs.
- **Mutation-checked by hand**, three mutations, all RED with the intended message: the catalog drops `candle-gen-flux2` → *"registers no 'flux2_dev' generator at all"*; `flux2_dev` answered by a different registration → *"resolves to a descriptor candle-gen-flux2 did not register"*; the memory-strategy route lost → *"registers no 'flux2_dev' memory strategy"*.
- `cargo test -p sceneworks-worker --features backend-candle --lib flux2_composition_audit` — 3 passed; `cargo clippy -p sceneworks-worker --features backend-candle --all-targets` clean.
- `npm run rust:check` (fmt + clippy `-D warnings` + full workspace `cargo test` + build) and `npm run check` (291 node tests) green.
- `docs/generated/*` regenerated.

> **Update after merging main.** The live inference pin moved to `fbb00d6b` while this was in review, so the five `flux2_dev` q4 cells are now `historical` rather than `Runtime verified` — the audited window `5ffd7612 → a4f409ae` no longer ends at the live pin. That is the fail-closed rule working, it is unrelated to this PR (the same demotion happens on `main` without it), and it does not touch what this PR decides: the closure definition and the composition check are pin-independent. Re-opening the window is a re-certification against `fbb00d6b`, which now runs against the nine-path closure. All four generated-doc conflicts from that merge were resolved by regenerating, never by hand.

## From adversarial review

An independent review found no high-severity defects and five medium ones; all are fixed:

- **The control was fetched through the crate under audit.** `runtime_cuda::providers::flux2` is `candle-gen-catalog`'s own `pub use candle_gen_flux2 as flux2`, so a catalog that re-aliased it to a wrapper would have moved both sides of every comparison together. Closed by asserting the registrar's *defining* path via `type_name_of_val`, which no re-export can rewrite (verified against a minimal repro before relying on it).
- **The guard on the composition module was a text-marker regex** that passed with both tests deleted, and said nothing about the `mod` declaration — the quietest way for this trade to be undone. Now matches the `#[test]` function definitions and the cfg'd `mod` line in `lib.rs`; both halves mutation-checked RED.
- **"Nothing, since sc-17607" was false** — see *What this genuinely costs* above.
- **The frozen `adjudicates` half is now a ceiling at maximum** and can refuse nothing on its own, since the closure and the frozen set are pinned equal. Three comments still sold it as two independent halves; they now say which half is live.
- **The bundle-composition test's stated rationale was impossible** (a lane without `backend-candle` doesn't compile the module at all, so it cannot fail confusingly there). Rewritten to what the test actually buys.

Also fixed from the low findings: the `COMPOSITION_ONLY` pair is now pinned across both languages, the closure ⊆ `adjudicates` assertion became an equality in both directions, a vacuous id assertion was dropped, a redundant count comparison was collapsed, and a stray intra-doc link to the cfg-gated module was removed.

One low finding is **left alone deliberately**: `scripts/generate-memory-matrix.mjs:71` still says "Three closure paths moved" where two did. That is pre-existing and is exactly what open PR #2118 fixes; touching it here would only conflict.


michaeltrefry:
## ⚠️ Schema-version collision with #2122 (sc-17606) — read before merging either

[#2122](https://github.com/SceneWorks/SceneWorks/pull/2122) opened while this was being written and
**also bumps `schemaVersion` 3 → 4**, with a different meaning:

| | this PR (sc-17607) | #2122 (sc-17606) |
| --- | --- | --- |
| closure | **nine** paths — `runtime-cuda` removed | **ten** paths — `runtime-cuda` kept |
| v4 means | the measurement binary's compile closure | the ten-path closure **plus** a resolved-feature witness |
| `FLUX2_V4_AUDIT_METHOD` | *"…the Candle FLUX.2 measurement binary's compile closure and its workspace build inputs"* | *"…the complete Candle FLUX.2 runtime dependency closure … and the shipped bundle's resolved feature set"* |

Both edit the same twelve files, including the same three frozen tables and the same record.

### These are complementary, not competing — both should land

The gap this PR opens is the one #2122 closes. Removing `crates/bundles/runtime-cuda` from the
closure stops watching `runtime-cuda/Cargo.toml`, where the shipped feature resolution lives — which
this PR documents as an accepted cost "until sc-17606 lands the resolved-feature-set witness".
#2122 *is* that witness, and it hashes `cargo tree -p runtime-cuda -e normal,build`, which sees a
manifest feature edit directly. Landed together they cover both halves: composition by identity
(here), feature resolution by witness (there).

### Proposed resolution

1. **Sequential versions, not a merged v4.** Whichever lands first keeps `4`; the second rebases to
   **`5`** and refuses `1..=4`. Each changes the record's *shape*, and the version exists precisely
   so a record produced under one shape is never re-graded under another — one number describing two
   shapes depending on which record you hold is the failure this scheme is built to prevent.
2. **The record must be re-emitted by the tool, not hand-merged.** Both PRs changed
   `inference-compatibility-a4f4.json`. The second one through needs a real re-run — cheap now:
   `--workdir D:\repos\inference-audit\sc-17524` reuses the surviving `sc-17524-target`, and that
   made this PR's re-run two CUDA release builds in ~14 minutes, no GPU and no weights.
3. **`docs/inference-artifact-audit-sc-17497.md` needs a hand merge in one place.** Both PRs rewrite
   the *"What that trade gives up"* list in opposite directions. After both land it should read:
   `runtime-cuda` is out of the closure (here), and the feature-resolution gap that leaves — plus the
   narrower-unification gap #2122 was written for — is covered by the witness. If #2122 lands first,
   this PR's "until sc-17606 lands" wording is stale on arrival and should say "covered by the
   feature witness" instead.

Flagging rather than resolving: which order these merge in is not mine to choose, and rebasing blind
would just move the conflict.

michaeltrefry:
**Ordering decided: #2122 merges first, this PR rebases to v5.** (Michael, 2026-08-05.)

Recording it here so whoever picks this up does not have to re-derive it from the collision comment above.

## What `main` will look like when you rebase

#2122 lands the **ten**-path closure at **`schemaVersion: 4`**, with a third layer — a resolved-feature witness over `runtime-cuda`'s resolution, scoped to the 173 packages `candle-gen-flux2` links. Concretely, on top of what the collision comment already lists you will find:

- `FLUX2_V4_AUDIT_METHOD` = *"compiled artifact identity for changed paths, git object identity for unchanged paths, **and shipped-bundle resolved-feature identity**, across the complete Candle FLUX.2 runtime dependency closure and its workspace build inputs"*
- a **third** frozen constant pair: `FLUX2_COMPATIBILITY_AUDIT.featureWitness` (`scripts/generate-memory-matrix.mjs`) and `FLUX2_FEATURE_WITNESS` (`crates/sceneworks-worker/src/inference_compatibility_audit.rs`), each carrying `digest` **plus the resolution it was taken over** (`shippedPackage`, `scopeRoot`, `scopeFeatures`, `target`, `edges`) — all five are graded by both validators
- `FEATURE_WITNESS_RESOLUTION` in `scripts/inference-artifact-audit.mjs`
- the shipped record at `docs/calibration/sc-15833/inference-compatibility-a4f4.json` carrying a `featureWitness` block

## What this PR then has to do

1. **Bump to `schemaVersion: 5`** and refuse v4 alongside v1/v2/v3 — same argument the existing versions use: a v4 record was graded against a ten-path closure, so it cannot be re-read as evidence about a nine-path one.
2. **Re-emit the record with the tool**, do not hand-edit it. Removing `crates/bundles/runtime-cuda` changes `auditedObjects` and therefore the closure-identity check in both validators.
3. **Re-freeze all three constants** from that run — `compatibleInferenceRevision`, `artifactProof` and `featureWitness.digest` — in both languages. The witness digest will **not** change (it is scoped to what `candle-gen-flux2` links, and `runtime-cuda` is not in that set), but it still has to come from the emitted record rather than be carried over on faith.
4. The witness is what now covers the composition question you are removing from the closure, so the two changes are complementary — worth saying so in the doc where it explains why `runtime-cuda` left.

## Two operational notes

- `scripts/inference-artifact-audit.mjs` now creates a worktree and runs `cargo tree` on **every** run, including the free path. It is resolution, not compilation — no MSVC, no nvcc — but the free path is no longer a pure `git rev-parse`.
- Use `--workdir D:\repos\inference-audit\sc-17524`: the derived `-target` there survives and only that exact path rehydrates it (the remapped paths are baked into every rustc fingerprint). A different path is a cold multi-hour compile.

## Heads-up on the live pin

`main` moved the inference pin to `fbb00d6b` in #2120 **without re-running the audit**, so the five `flux2_dev` q4 cells are already demoted on `main` today. Four closure paths moved across `a4f409ae → fbb00d6b`, including `candle-gen-flux2` itself, so extending that window is a real two-revision build that could return either a repin or a re-capture verdict. That is a separate question from this PR — do not let it expand your scope — but do not be surprised when your regenerated matrix shows zero `Runtime verified` FLUX.2 cells. It is not something you broke.

[author: claude]

michaeltrefry:
## Rebased to v5, anticipating #2122 (sc-17606) merging first

Done per the ordering above. #2122's branch is merged into this one, so when it lands through the queue this PR is already reconciled rather than newly conflicted.

**`v5` = sc-17607's nine-path closure carrying sc-17606's resolved-feature witness.**

Against the four numbered items:

1. **`schemaVersion: 5`, refusing v1–v4.** v1/v2/v3 stay refused for auditing too *little*; v4 is refused for auditing one path *more* than this schema asks about — accepting one means dropping an entry out of someone else's record, the same unearned re-reading by subtraction. `FLUX2_V5_AUDIT_METHOD` names both layers: *"…and shipped-bundle resolved-feature identity, across the Candle FLUX.2 **measurement binary's compile closure** and its workspace build inputs"*.
2. **Record re-emitted by the tool**, not hand-edited — `--lane cuda`, two real CUDA builds off the warm `sc-17524` target, exit 0:
   ```
   schemaVersion : 5      objects : 9   (runtime-cuda absent)
   changed       : Cargo.lock, crates/media/candle-gen/candle-gen
   artifact      : sha256:d80844f2…   captured == compatible
   witness       : sha256:321625ed…   captured == compatible, 173 packages
   ```
3. **All three constants verified against that record** in both languages. The witness digest is unchanged, exactly as predicted — it is scoped to what `candle-gen-flux2` links, which never included the bundle — but it came from the emitted record rather than being carried over.
4. **The doc needed more than a note.** Both stories rewrote *"What that trade gives up"* from opposite ends, and this branch's *"open until sc-17606 lands"* wording was false the moment that story landed. It now reads: the witness covers the manifest gap this change opens, and covers it **better** than the tree object did — a feature edit that reaches a package FLUX.2 links moves the witness digest, one that does not moves nothing, and the tree object could only tell those apart by demanding an unadjudicable re-capture.

**Rust validator** was rebuilt on the witness-threaded base with this story's changes re-applied (9 audited objects, `COMPOSITION_ONLY`, the closure↔adjudicates equality test, the v5 fixture's "moved must name a real closure path" guard). The `runtime-cuda` negative cases became frozen-set-withdrawal cases, since no audited path is unlinked any more — the guard stays exercised for the day a future closure entry is not compile-covered.

Verified locally: `cargo fmt`, `clippy --all-targets -D warnings`, 7 validator tests, 3 composition tests (candle lane, real GPU box), 100 JS tests, derived docs regenerated.

[author: claude]

PR #2140, opened 2026-08-05
Land the Anima, SDXL and Kolors MLX memory ladders (sc-15524, sc-15525, sc-15521)

michaeltrefry:
The SceneWorks half of three epic-15448 paired deliveries whose inference half is already merged. All three stories are sitting in **In Review** waiting only on this.

| Story | Family | inference PR |
|---|---|---|
| sc-15524 | Anima 2B (Base / Aesthetic / Turbo) | #477 |
| sc-15525 | SDXL + 4 derivatives | #480, corrected by #494 |
| sc-15521 | Kolors | #493 |

## Why one PR and not three

Epic 15448 requires the pin to move to the merged inference SHA before the SceneWorks side can regenerate anything at that revision. A pin bump has documented repo-wide blast radius — the provenance/license audit, the entire generated matrix, the cost model, and every currentness verdict in the calibration corpus. Doing it three times triples that churn and produces no extra signal, so the wave is batched.

## Pin

`fbb00d6b` → **`574f0d99f44740622cacb51735474a210c678a96`**, verified as the first `origin/main` tip containing the merges of #477 (`e91c3267`), #480 (`59a2bd00`), #493 (`82e49a36`) and #494 (`862c1a65`). (#477 was already inside the old pin; #480/#493/#494 were not.)

All three build configs move together via `scripts/bump-inference.mjs` — the root `Cargo.toml` `candle-kernels` `[patch]`, `crates/sceneworks-worker/Cargo.toml` and `crates/sceneworks-memory-adapter/Cargo.toml` — plus `INFERENCE_RUNTIME_REVISION` and `INFERENCE_PIN`. Gen-core and mlx-rs skew checks pass: exactly one `sceneworks-gen-core` and one `pmetal-mlx-rs` in the graph.

Provenance/license audit re-scanned against a local inference clone at the new revision: 624 candidates (population hash moved; 11 files' content changed, no files added or removed), 94 crate prefixes unchanged, `include_str!`/`include_bytes!` production sites byte-identical at both revisions (53, same set), audit digest recomputed.

## What each story lands

**sc-15524 Anima** — the catalog's first *complete* ladder. Rungs 1/2/3/4 Implemented on all three entries. Rung 2 publishes the admitted tile domain `[768, 640, 512, 384, 320, 256, 192]` at overlap 64, default 640 — the cheapest edge still holding max Δ 1/255 (768 costs 4.649 GiB for the same quality; edges 160/128/96 are measured and *rejected* on drift). Rung 4's cadence domain is genuinely `[1]`: windows 2/4/14/28 are refused, and window 1 differs from the next admitted plan by a real 26% (1.110 vs 1.497 GiB of denoise). Scope `Dit` only, and that is a measured decision — rung 1 already sheds the Qwen3-0.6B tower and the 6-block conditioner is ~1% of the DiT, so a `TextEncoder` scope's only honest value here is 0.0%. Fingerprint `anima-mlx-cosmos-shared-ladder-v1`.

**sc-15525 SDXL + derivatives** — rung 4 is declared **per entry per tier**, never family-wide, because SC-16355's AC is a request peak measured per tier:

| entry | tier | staged | rung 4 | buys | verdict |
|---|---|---:|---:|---:|---|
| `sdxl` | bf16 | 19.005 GiB | 14.938 | −21.40% | Implemented |
| `realvisxl` | bf16 | 19.003 | 14.936 | −21.40% | Implemented |
| `realvisxl` | q8 | 17.693 | 15.516 | −12.31% | Implemented |
| `realvisxl` | q4 | 16.654 | 15.493 | −6.97% | Implemented |
| `realvisxl_lightning` | q4 | 16.654 | 15.493 | −6.97% | Implemented |
| `illustrious_xl_v1` | q8 | 17.693 | — | — | **Missing** |
| `illustrious_xl_v2` | q8 | 17.693 | — | — | **Missing** |

The two Illustrious entries are Missing *outright* — q8 is their only advertised tier, and its `unet/config.json` ships without the `quantization` marker, so `packed_quant_bits` reads nothing and `streamable` refuses. The rung fails closed, which is correct; the repair is a republished snapshot (inference sc-17522).

Cadence domain `[1, 2, 5, 10]` with default **10** — PR #494's correction. The 512² row that once justified 1 was positional allocator noise, not a cadence effect, and the peak column is flat wherever the instrument can resolve it, so the widest cadence costs no memory and renders 1262 vs 3524 ms/step at 1024². Fingerprint bumped to **`sdxl-mlx-unet-shared-ladder-v3`** so v2-era evidence cannot bind.

Rungs 2 and 3 stay `Missing` **with their measurements recorded rather than omitted** — rung 2 withheld on quality (48/255 inherited bar vs 84/255 drift on the *production* latent, where the synthetic sweep read 38), rung 3 at 0.00% on both the request peak and the U-Net seam and not bit-exact on its query-row axis. The matrix forbids a non-`unmeasured` characterization on a non-implemented state, so those numbers live where every other measurement in this manifest lives: in the declaration's own prose, next to the rung it withholds.

**sc-15521 Kolors** — rungs 0/1/4 Implemented on all three tiers, rungs 2/3 Missing with numbers. Per-tier rung-4 saving at 1024²: q4 −7.21%, q8 −12.72%, bf16 −21.37%.

This is the ladder's **first genuinely three-valued rung-4 scope**: `[Dit, TextEncoder, Both]`, reading 11.3644 / 8.8396 / 4.5436 GiB at bf16/512², where ChatGLM3-6B carries the peak and bounding the U-Net alone moves nothing at all.

> **The scope axis needed no schema change.** The matrix already carried it as `transformerWindowComponent` inside `strategyParameters` (and `transformerWindowComponents` in `publishedRanges`), the survey and manifest already spelled it PascalCase, and `mlx_fit_gate.rs` already parses `dit` / `text_encoder` / `both`. What is new is that the domain is load-bearing rather than degenerate — every prior family published a single value.

Fingerprint `kolors-mlx-chatglm3-sdxl-unet-ladder-v1`, deliberately **not** shared with SDXL's even though `crate::unet` re-exports SDXL's `UNet2DConditionModel` verbatim: the mechanism transfers, the numbers do not.

## Survey correction

`sc-15525`'s survey text said SDXL has **four** 2-deep `Transformer2D` sub-stacks. It is **five** — the up path carries `layers_per_block + 1 = 3` attentions. Forward-order depths are `[2, 2, 10, 10, 10, 10, 10, 10, 2, 2, 2]`: eleven stacks, six at depth 10, five at depth 2, **70** windowable blocks. Corrected in the survey summary, the block-stack inventory, an explicit finding, and the manifest contract comment.

## Worker

**No change was needed, and that is the finding.** The `SEQUENTIAL_CAPABLE_ENGINES` allowlist the inference doc comments still cite no longer exists in this repo — `engine_supports_sequential` derives from the descriptor's `supports_sequential_offload` bit, all three families advertise it at the new pin, the registry sweep test names all three, `spec_component_bytes` already calls `media().footprint(engine_id, spec)` generically, and `image_jobs/base.rs` reads the predicate on the live request path. Anima's staged-residency cells were already `Implemented/unverified`. Filed **sc-17729** for the stale inference comments (bernini and boogu carry the same claim).

`jobs_store::routing::mlx` and the memory adapter needed nothing for the scope axis — it was already plumbed end to end.

## Closure-currency reds: resolved, and how

Five tests were red on this branch purely from the closure move, all asserting the same invariant. **All are green now**, and none were deleted.

`fbb00d6b → 574f0d99` moves three files inside every MLX provider's compile closure (`mlx-gen/src/asset_facts.rs`, `mlx-gen-pid/src/engine.rs`, and a comment-only `quant.rs` that the semantic hash absolves), so `mlx:qwen_image` and `mlx:krea_2_turbo_control` genuinely re-date. That is the mechanism working, not a bug.

What was wrong was that the tests folded two questions into one assertion:

- **Agreement** — do the shipped manifest opt-in and the packaged evidence bundle describe the same measurements? A fact about two checked-in files; no pin or closure move can falsify it.
- **Currency** — is that capture still authoritative at the live closure? A comparison against a table that moves whenever a shared `mlx-gen` crate does.

Graded together, the move did more than fail the currency claim — it made the *other* claims untestable. `packaged_krea_1024_refuses_before_render_and_names_only_a_fitting_current_cell` picks its fixture by digest and needs two exact cells to choose between, so the moved closure emptied the fixture and the refusal-naming behaviour went silently unexercised.

Each test now grades agreement at the closure the artefacts were captured under, and **derives** the currency verdict from the digest pair rather than hardcoding it either way:

| digests | required outcome |
|---|---|
| `declared == live` | calibrated admission must be reached (`AdmissionPath::Evidence`) |
| `declared != live` | admission must fall back to `Legacy` / `StaleIdentity` — asserted, not skipped |

Both branches are live code, so a re-capture restores the `Evidence` requirement with no edit to these tests, and a demotion for the *wrong* reason (missing record, unparseable opt-in) still fails. Digests are read from the shipped opt-in via `shipped_mlx_declared_closure_digest`, never restated as hex literals, and that helper refuses a split opt-in — one stale binding cannot hide inside a current ladder.

**Production behaviour is untouched: every hunk is inside `mod tests`.** `MlxCalibrationBinding` still fails closed on a superseded closure, which is exactly what makes the demotion branch assertable.

### Stated plainly: this is a relaxation

These tests no longer *require* the shipped MLX evidence to be current — they require it to be current **or honestly demoted**. On the Python side `measured_at_live_pin` is now a two-way equality against the empty set rather than a non-emptiness assertion. sc-17726's original acceptance said these should go green without weakening any of them; that AC is superseded by its own comment 18106 (2026-08-07), which hands the posture to **epic 18093** — stale lanes keep serving measured numbers with a widened margin instead of triggering corpus-wide recapture. If that reading is wrong, this is the change to revisit.

The re-measurement itself is still owed and still tracked on sc-17726.

## Follow-ups filed

- **sc-17726** (bug) — re-measure the MLX calibration corpus at the new pin. No longer blocks this PR: the five tests now grade the demotion honestly instead of requiring currency. Superseded by epic 18093; see comment 18114 for the full disposition.
- **sc-17727** (chore) — backfill `footprint.residentMemoryBytes` / `peakMemoryBytes` for the 11 hosted tier variants still null (sdxl q4/bf16, realvisxl ×3, realvisxl_lightning ×3, kolors ×3), plus the Anima structural question.
- **sc-17728** (bug) — `parse_evidence_parameters` requires rung-2/3 parameters on *every* rung-4 calibration binding, which structurally blocks SDXL and Kolors evidence since both implement rung 4 with rungs 2/3 Missing.
- **sc-17729** (chore) — stale `SEQUENTIAL_CAPABLE_ENGINES` doc comments in inference.

## Footprint (SC-15674)

Anima's six downloads gained `footprint` blocks — they had none at all. `residentMemoryBytes` / `peakMemoryBytes` stay null on the 11 unmeasured hosted tiers, which is the schema's *declared-unmeasured* value rather than an omission. They were **not** filled from the inference evidence on purpose: those figures are ladder request peaks under a named composition, whereas these fields are `footprint_measure.rs` quantities (resident = `get_active_memory()` after a generation *and* a `clear_cache()`; peak = `get_peak_memory()` across load + one generation on the resident hold-all path). Substituting one for the other would corrupt the sc-8509 RAM→tier suggestion, this field's only consumer. Recorded in the manifest and tracked on sc-17727.

## Verification

Locally on Apple Silicon at pin `574f0d99`:

- `cargo test -- --skip nax_16bit_sdpa_is_correct` — **31 suites ok, 0 failed**
- `cargo fmt --all -- --check` — **pass**
- `cargo clippy --all-targets -- -D warnings` — **pass**
- `cargo clippy -p sceneworks-memory-adapter --features mlx --all-targets -- -D warnings` — **pass**
- `npm run check` — **pass** (368 node:test assertions, 0 fail)
- `npm run check:compose`, `npm run check:docker:rust-api:self-test` — **pass**
- `pytest -q tests/ -m "not e2e and not parity"` — **87 passed**

In CI at `5b8374e8`, all seven required contexts pass — `build-windows`, `candle`, `check-linux`, `check-macos`, `macOS build, lint and workspace tests (hosted)`, `parity`, `web` — plus non-required `candle-worker` and `nax-worker`.

Note for anyone touching `mlx_fit_gate.rs`: it is one of the fifteen `SOURCE_PATHS` the memory matrix is fingerprinted on, so even a test-only edit requires `npm run generate:memory-matrix`, which in turn rotates `calibration-cost-model.json`'s `matrixRevision` and its sha256 over the matrix and requires `npm run generate:calibration-cost-model`. Both hops are in this branch.


michaeltrefry:
## Update — stage-1 facts re-dumped (d1cc0f62)

`nax-worker` went red on **`Verify capabilities.mlx.json is a real dump, not a restamp`**, and `web` went red on four `previewSupportCatalog` assertions. Both are the documented pin-bump blast radius: the engine-capability facts are stamped with the revision they were dumped at.

Two of the three are now fixed from a **real dump on this Mac**, not a restamp:

- `config/engine-capabilities/capabilities.mlx.json`
- `config/engine-capabilities/audio/capabilities.candle.json` (lane-neutral; the same run writes it)

Both came out **byte-identical to the checked-in files apart from the `inferenceRevision` line** — so the MLX registry is measurably codegen-inert across `fbb00d6b..574f0d99`, which is exactly the claim that guard exists to demand and which the two prior restamps (`35251a88`, `06e0c5e9`) asserted without evidence.

### The candle half needs the CUDA box — tracked on sc-17733

`config/engine-capabilities/capabilities.candle.json` is **not** updated and cannot be from here. `dump-engine-capabilities` has to *run*; running links; linking the candle lane needs a real `libcuda`. `scripts/check-candle-build.mjs` documents that boundary in its own header — a stub toolkit gets you a typecheck and nothing more. `windows-candle.yml` only *diffs* its fresh dump rather than uploading it, so it cannot be harvested from a CI run either.

Command for whoever has the box:

```
cargo run -p sceneworks-worker --bin dump-engine-capabilities \
  --no-default-features --features backend-candle -- <scratch>
```

then copy `<scratch>/capabilities.candle.json` into place and run `npm run gen:preview-support` from `apps/web`.

### Deliberate intermediate state

`config/engine-capabilities/` is temporarily **split**: mlx + audio at `574f0d99`, candle media at `fbb00d6b`. `derivePreviewSupport` fails closed on precisely that split, so `gen:preview-support` refuses to run and `previewSupportCatalog.test.js` throws at suite level until the candle dump lands.

That is a deliberate trade, not an oversight. The alternative was to discard the Mac dump and leave all three files stale. Rejected: the Mac dump is the scarce artifact here — this repo's own history is two restamps caused by no Mac being available — and banking it cuts the remaining work to one machine-specific command instead of two. The failure is loud and its error message names the exact fix.

### CI at `d53dbf45`, for the record

| check | result |
|---|---|
| `check-linux` | pass |
| `check-macos` | pass |
| `Analyze (javascript-typescript)` / `Analyze (python)` | pass |
| `nax-worker` | fail — stale `capabilities.mlx.json`; **fixed by d1cc0f62** |
| `web` | fail — stale facts stamps; **candle half remains, sc-17733** |
| `parity` | fail — the two currentness tests only (`2 failed, 85 passed`), **sc-17726** |

The `parity` failure is byte-for-byte the local result: `test_calibration_evidence_is_schema_valid_and_matrix_ingested` and `test_historical_records_remain_unverified_after_the_z_image_pin_advance`, both purely the pin move invalidating the 17 MLX records captured at `fbb00d6b`. Nothing in that job is related to the three ladders.

michaeltrefry:
## CI at `d1cc0f62` (head)

| check | result |
|---|---|
| `nax-worker` | **pass** — the Mac re-dump fixed the "real dump, not a restamp" step |
| `check-linux` | pass |
| `check-macos` | pass |
| `build-windows` | pass |
| `Analyze (javascript-typescript)` / `Analyze (python)` | pass |
| `package-windows` | skipped |
| `web` | fail — `previewSupportCatalog.test.js` only (**1 failed | 235 passed**), the candle facts split, sc-17733 |
| `parity` | fail — **2 failed, 85 passed**, both the currentness tests, sc-17726 |
| `candle-worker` | fail — **1243 passed, 4 failed**, all `mlx_fit_gate` `StaleIdentity`, sc-17726 |
| `macOS build, lint and workspace tests (hosted)` | still running at time of writing |

Every failure is one of the two tracked pin-bump consequences. Nothing red is attributable to the three ladders themselves.

### The candle-worker failures are the more serious half of sc-17726

```
shipped_qwen_manifest_and_packaged_evidence_select_the_calibrated_path
  bf16: shipped manifest + shipped evidence must reach calibrated admission;
  got fallback Some(StaleIdentity)
shipped_krea_manifest_and_packaged_evidence_select_the_calibrated_path
shipped_qwen_manifest_carries_every_tier_ladder_at_the_running_pin
packaged_krea_1024_refuses_before_render_and_names_only_a_fitting_current_cell
```

The JS/Python failures are about what the generated matrix *reports*. `StaleIdentity` is the **live admission path actually falling back to the legacy estimator** — the same thing, observed directly. The `mlx.calibrations[]` bindings for qwen_image and krea_2_turbo_control carry `inferenceRevision: fbb00d6b`, and `MlxCalibrationBinding::from_manifest` rejects them against the new runtime revision.

Deliberately **not** fixed by editing those `inferenceRevision` lines: that field is capture provenance, and rewriting it would claim a measurement was taken at a revision it was not — the manifest-side equivalent of the restamp the `nax-worker` guard exists to catch. The sanctioned field is `compatibleInferenceRevision`, which the schema admits "only after a checked-in provider dependency-closure audit proves the captured implementation unchanged". One audit (or one re-measurement campaign) closes all nine failures together. Full inventory is on sc-17726.

Also worth noting for whoever picks that up: these four did **not** surface on `check-macos`, which passed at the same SHA. The Mac lane appears blind to a live-admission regression the Windows lane catches — recorded on the story.

Issue #2074, opened 2026-08-02
SCAIL-2 Animate character returns HTTP 400 "Unsupported video mode" on macOS v0.8.1


sahilkashyap64:
<img width="3456" height="2234" alt="Image" src="https://github.com/user-attachments/assets/d72669d6-a1b0-4f59-8720-2e700f9a72ff" />

### What happened?

SceneWorks:
v0.8.1

System:
MacBook Pro M5 Max
128 GB unified memory
macOS 26.2+
Apple MLX worker detected

<img width="1080" height="1920" alt="Image" src="https://github.com/user-attachments/assets/00c925df-c7fe-47bf-bc34-620fc298315d" />

### Steps to reproduce

Model:
SCAIL-2 Q4, installed successfully

Mode:
Animate character

Inputs:
Driving video: H.264 MP4, 832x480, yuv420p, 25 fps, 2.56 seconds
Reference image: PNG, 1080x1920

Settings:
Resolution 832x480
Static motion

https://github.com/user-attachments/assets/54ed05cf-40d2-4736-a558-bd48d52955a4

<img width="1080" height="1920" alt="Image" src="https://github.com/user-attachments/assets/8e7dbc30-1754-430b-ad76-48b43e10badd" />

### Platform

macOS (Apple Silicon / MLX)

### SceneWorks version

0.8.1

### Model / LoRA involved (if any)

SCAIL-2 Q4

### Logs / screenshots

```shell

```

PR #2145, opened 2026-08-05
ci: give the merge queue a backend verdict, on hosted runners

michaeltrefry:
## What does this PR do?

The merge queue gates on `web` + `parity` only. Both live in `check.yml`, which already has `merge_group:` — so the queue *does* validate the speculative merge, but only for what a hosted Linux default-feature build compiles. Nothing that compiles a **backend cfg** saw the queue at all:

| configuration | only CI compiler | saw the merge group? |
|---|---|---|
| macOS / mlx | `macos-mlx.yml` | ❌ no `merge_group:` |
| Linux, candle OFF | `check.yml` `parity` | ✅ |
| Linux, candle ON | `windows-candle.yml` | ❌ no `merge_group:` |

So two individually-green PRs whose merge doesn't compile under either backend cfg would merge clean.

**One correction to the premise this started from:** today's break ([e71b932f](https://github.com/SceneWorks/SceneWorks/commit/e71b932f)) was *not* an instance of this. It landed in `crates/sceneworks-core`, a workspace default member, so `parity`'s bare `cargo test` already covers it and the queue **would** have caught it. The gap is real, but it's the cfg-gated class — not that one.

### Why not just add `merge_group:` to both lanes

Two facts, measured before writing any of this:

1. **`merge_group` supports no `paths:` filter** ([community #45899](https://github.com/orgs/community/discussions/45899)). Every other trigger on these lanes is path-gated, and those filters guard a two-Mac pool and the fleet's single `cuda` box. A bare trigger wakes both for a README-only queue entry.
2. **`windows-candle.yml` cannot absorb the queue.** Five consecutive `candle-worker` runs on 2026-08-05: **27.6 / 27.5 / 29.1 / 30.6 / 24.2 min**, on top of 4–24 min queue waits. The in-file cost comment still claims *"~4.2m median (p90 4.5m)"* and *"~12% utilized"* — **~7× stale**. With `max_entries_to_build: 5` and ~30 merges/day, queued entries would pass `check_response_timeout_minutes: 60` and be **evicted** — the exact failure `check.yml`'s own comment warns about.

So the queue gets **hosted** coverage, and the scarce boxes stay out of it.

### The change

- **`check.yml`** — new `candle` job running `npm run rust:check:candle`. Typechecks `all(not(target_os = "macos"), feature = "backend-candle")` on a GPU-less runner via the stub `nvcc` from sc-12356 (cargo check/clippy never link, so the typecheck is real). **That script has existed since sc-12356 but ran nowhere in CI** — only the pre-push hook, with `--allow-skip`. This is its first gate.
- **`macos-mlx.yml`** — `merge_group:` trigger, plus a `changes` gate job supplying the path filter the event can't carry, by **parsing this lane's own `mlx_paths` anchor rather than copying it**. A copy rots in the silent direction: a newly watched path the gate doesn't know about reads as "not relevant" and skips the lane on a green-looking merge group. Hosted `macos-checks` carries the queue verdict; self-hosted `nax-worker` is excluded — its unique contribution is `nax_16bit_sdpa_is_correct`, a property of the M5 matrix-unit kernels rather than of how two PRs compose, and that pool already produced six jetsam kills on 2026-08-04.

## Related issue

No GitHub issue. Follow-on work tracked in **sc-17773**. (The worktree's `github-issue-2074` name is stale — #2074 is an unrelated closed SCAIL-2 bug.)

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that changes existing behavior)
- [ ] Documentation only
- [x] Refactor / chore (no user-facing behavior change)

## How was this tested?

**The gate was mutation-tested, not just eyeballed.** Changed `cuda_vram_cap_gb()`'s return type in `vram_gate` (gated *exclusively* on candle):

- `cargo check` → **exit 0**. Both `parity` *and* macOS are blind to it.
- `npm run rust:check:candle` → **exit 101**, `E0308` at three call sites.

Also ran the candle typecheck clean end-to-end (exit 0, 34 candle-gen crates + memory adapter, 24.8s warm, stub `nvcc`), and simulated the relevance gate against a real diff, a docs-only diff, and an empty diff.

**Guards:** 7 unit tests (anchor parser + glob semantics), 3 contract tests pinning which lanes see the queue and which deliberately don't. All 6 mutations caught — dropping either trigger, gutting the candle command, adding the `cuda` box to the queue, an anchor typo, un-excluding `nax-worker`.

- [x] macOS (Apple Silicon / MLX) — `cargo check` + the candle cross-compile
- [ ] Windows (NVIDIA / candle) — not run on Windows hardware; the candle cfg was typechecked via cross-compile
- [ ] Docker server (candle / CPU)
- [x] Not platform-specific (CI configuration)

## Checklist

- [x] I ran `npm run rust:check` — n/a, no Rust changed; ran `cargo check` and `rust:check:candle` for the verification above
- [ ] I ran the web app checks — n/a, web untouched
- [x] I ran `npm run check` — **317/318**
- [x] I updated documentation where relevant — `CONTRIBUTING.md`'s "candle build" section framed `windows-candle.yml` as the only CI compiler of that cfg, which this change makes stale
- [x] All my commits are signed off
- [x] My PR is focused on a single logical change

### ⚠️ Two things a reviewer should know

**1. This makes the lanes *reachable*, not *enforcing*.** Both now run on the speculative merge, but neither blocks it — required checks are a ruleset setting (`17708030`), which is Michael's to change. Tracked in **sc-17773**:
- `candle` **can** be added to the required set safely — `check.yml` has no path filters, so it always reports.
- `macos-checks` **cannot**, without restructuring. A workflow skipped by *path filter* stays **Pending** and blocks; a job skipped by `if:` reports **Success** ([docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks)). Requiring it as-is would wedge every docs-only PR.

**2. One pre-existing test failure.** `npm run check` is 317/318; `inference-artifact-audit`'s *"main puts each BUILD's digest on its own side"* fails. **Confirmed it fails on clean `HEAD` with these changes stashed** — pre-existing, untouched here, and not fixed in this PR.

### Known limitation, stated plainly

`rust:check:candle` **typechecks only** — `cargo test` links and needs a real libcuda, so candle-gated `#[test]` *bodies* still first execute on `windows-candle.yml`, post-merge. What the queue now gates is the compile class (E0599 / E0277 / E0308, cfg-gated symbol gone, dead code under `-D warnings`) — which is the class a speculative merge introduces.


michaeltrefry:
**Correction to this PR's stated rationale** — filed as #2149.

This PR's body and commit message assert that `windows-candle.yml` runs on *"the fleet's single `cuda` box"* at *"~28 min"*. **The pool claim is wrong, and the duration figure was derived badly.**

Four runners carried the `cuda` label over the measurement window — `cuda-windows`, `-2`, `-3`, `-4` — evenly loaded (23/18/22/22 jobs), at **~29% per box**. Not one saturated machine.

I got there by reading `gh run list`'s `createdAt → updatedAt`, which **folds queue wait into run time** — ~50m for what is actually a 24.1m median job (p90 31.9m, over 85 completed jobs). I never checked `runner_name` on the jobs API, which had the answer the whole time.

**What this does and does not change:**

- ❌ The *reason* given here — "the box cannot absorb the queue, it's saturated" — does not hold. There was more headroom than I claimed.
- ✅ The *decision* to keep this lane out of the merge queue still holds, for a reason I should have led with: a p90 queue wait (18m) plus a p90 run (32m) reaches ~50m against `check_response_timeout_minutes: 60`. Too little margin to gate on.
- ✅ Everything else this PR shipped is unaffected — the `merge_group` triggers, the relevance gate, the hosted `candle` job, and the contract tests were all verified independently of the pool-size claim.

#2149 corrects the prose in all four files that carried it (including two pre-existing instances I inherited rather than wrote) and records the measurement method so the next reader re-measures instead of trusting the number.

PR #2174, opened 2026-08-07
fix(calibration): derive the capture-time closure key from the selected plan entry (sc-17774)

michaeltrefry:
## The bug

`runProviderPlan` stamps a provider's compile-closure digest at capture time. The default lookup keyed `config/inference-provider-closures.json` on `providerName` — the `--provider` flag, which carries a **plan entry name** (`mlx-z-image-q4-fresh-reference-resident`). That file is keyed `<backend>:<provider>` (`mlx:z_image_turbo`). The lookup was `undefined` for every possible input, and the capture aborted via `fail()`.

The two uses of `providerName` were mutually incompatible: passing the closure key as `--provider` instead would have made the plan filter select zero providers and fail there. And the documented `--backend` + `--fixture` invocation passes no `--provider` at all, so it looked up `undefined` and aborted the same way.

**Blast radius:** fails closed, so no shipped evidence carries a wrong digest. All 65 records in the bundle were stamped by `backfill-closure-digests.mjs`. The capture path had never once succeeded — which is why nothing noticed. It blocks every *new* authoritative capture, including the first for any newly declared model (sc-15528 / Bernini).

## The fix

Derive the key as `${backend}:${target.provider}` from the selected plan entries — both halves live there. This required moving `probeRepositories()` below plan selection, which is the first point where a selection exists.

Two things fell out of doing it properly:

- `capturesAuthoritativeEvidence` was computed over **all** `config.providers` rather than the selected ones. Now the digest is stamped exactly when the selection actually contains authoritative cases.
- A run writes one `repositories` receipt into every record, so it can stamp one digest. A selection spanning several closures is now refused, naming them. This doesn't constrain real use: all 43 fixtures and all 170 plan names each map to exactly one `backend:provider` pair, so it only trips on an unnarrowed backend-wide run — which could not have produced a correct digest under any key.

The key string is load-bearing beyond the lookup: `digestsAtRevision` passes the map key into `closureText`'s `# provider:` line, so a digest derived under any other string could never equal the declared one.

## Tests

Every existing `runProviderPlan` call injects `closureDigestFor: stubClosureDigest`, so the default path had **zero** coverage. Three tests now drive it uninjected:

1. **End-to-end** — builds a real-git inference checkout at the real declared crate path, uses the shipped closures config verbatim and the real `candle-krea-production-current-v1` plan entry, then asserts the stamped digest equals an independent `providerClosureDigest` call *and* that `evidenceSemantics` reports `current` against it. Also asserts the digest derived under the plan-entry name differs.
2. **Undeclared lane** — declaring under the plan entry name (the shape the broken lookup expected) must not satisfy the derivation.
3. **Unnarrowed run** — refused rather than stamped under one of several lanes.

Mutation-checked: with the fix reverted, all three fail (39 pass / 3 fail).

## Also in this PR

Two test files were referenced by no runner — absent from the hand-maintained `node --test` list in `package.json` and from every workflow:

- `scripts/inference-closure-digest.test.mjs` (26 tests) — covers the closure walk: `{ workspace = true }` resolution, the inert `[package.metadata.legacy-workspace-dependencies]` decoy, sibling-crate absorption, the stripper quote guard. All fail *green*, and CI's `--check` can't catch them because it re-derives through the same walk.
- `scripts/sc-16170-certification.test.mjs` (2 tests)

Both already passed. **291 → 319 tests.**

`docs/generated/calibration-cost-model.json` is regenerated (two hashes) because the harness is one of its 11 fingerprint sources.

## Known gap, not addressed here

`candle:z_image` has 90 authoritative plan entries, no closure declaration, and no records. This fix makes it fail closed at capture time rather than silently; nothing in `npm run check` flags it today. Pre-existing, filed separately.


michaeltrefry:
Superseded by #2171, which landed the same fix with a better design: it derives one closure digest per authoritative lane (`laneOf`) and stamps it per-record, so it supports multi-lane captures instead of refusing them, and it keeps `--resume` working by comparing on repository identity.

#2171 also covers everything this PR's tests covered, plus the `[patch]` closure dimension (sc-17935).

The one item here not carried by #2171 — gating `scripts/sc-16170-certification.test.mjs`, which is still referenced by no runner on `main` — is split out into a separate one-line PR.

Closing unmerged; nothing is lost.

PR #2175, opened 2026-08-07
fix(worker): key MLX calibration currency on the provider closure, not the pin (sc-17726)

michaeltrefry:
## The defect

sc-17774 (#2166) replaced pin-identity calibration invalidation with a per-provider compile-closure digest. [`docs/calibration-invalidation-sc-17774.md`](docs/calibration-invalidation-sc-17774.md) states the rule: *"`inferenceRevision` survives on every record and binding as capture provenance. It says where a measurement came from. It is never a currency term."*

Two live filters in `mlx_fit_gate.rs` survived that sweep and still compared exactly that:

- the `identity_matches` filter in `evidence_admission_route`
- the same conjunct in `verified_lower_alternative`

Both read `binding.query.inference_revision == catalog_semantic_jobs::INFERENCE_RUNTIME_REVISION` — the raw Cargo pin. **Every MLX calibration binding was excluded the moment the pin moved**, whether or not that provider's closure had. That is the repo-wide pin-keyed invalidation sc-17774 exists to remove, still in force for the entire MLX lane.

Reproduced in isolation — bumping the pin alone, with `config/inference-provider-closures.json` untouched so no closure moves:

| | main | this branch |
|---|---|---|
| `cargo test -p sceneworks-worker --lib mlx_fit_gate` | **4 failed**, 80 passed | **0 failed, 84 passed** |

## Narrowed, not deleted

Both conjuncts become the closure comparison. Deleting them is **not** equivalent, for different reasons at each site:

- **`verified_lower_alternative`** has no downstream currency term at all. Its result never becomes a `Candidate`, so `memory_strategy::candidate_exclusion` never grades it. Unfiltered, it would name a smaller geometry that the very next request refuses for the identical staleness.
- **`identity_matches`** sits *ahead of* the decision to enter `AdmissionPath::Evidence`, which withholds the resident baseline on purpose. A stale binding admitted there has nothing to fall back to and kills the request instead of demoting it. `z_image_turbo` ships with a genuinely expired ladder today (measured at `d4802320`; `mlx:z_image_turbo` has moved since) and renders **only** because this filter demotes it — deleting the conjunct breaks that lane now and every MLX lane whose closure later moves.

An expired calibration is epistemically the same as no calibration and must reach the legacy estimator. That is already what the candle lane does, for a different structural reason (its resident candidate is always in the pool). `candidate_exclusion` keeps the same comparison as the fail-closed backstop for anything reaching the selector another way.

The live digest is resolved once, hoisted above the admission call, and threaded to both seams so the injected test resolver reaches them — the synthetic fixture lanes are deliberately absent from the shipped closure config.

## Tests

Retargeted from the pin to the closure digest:

- `shipped_qwen_manifest_carries_every_tier_ladder_at_the_running_pin` → `..._at_the_current_closure`
- `shipped_z_image_manifest_rejects_historical_exact_mlx_ladder_rungs` (now grades the closure; the revision literal stays as explicit capture provenance)
- `packaged_krea_plan` / `packaged_krea_generator` record filters
- `a_moved_provider_closure_demotes_the_calibrated_ladder` — asserts the demotion at the admission seam, that no stale lower alternative is named, and end-to-end via `process_limit_bytes`, which only a verified cell sets

Two inert pin restatements removed from the fixtures: `fixture_bundle` was rewriting each record's revision to the live pin — dead under the closure unit, and a restatement of the pin as a currency term in the file every gate test builds its evidence from.

### Mutation checks

| mutation | caught by |
|---|---|
| drop conjunct in `identity_matches` | moved-closure test **+ the real shipped z_image test** |
| restore compare-against-self (sc-17774's defect 1 shape) | same two |
| drop conjunct in `verified_lower_alternative` | survived initially — added a case to `verified_lower_geometry_requires_exact_current_identity_and_geometry`; now caught |
| grade qwen / krea against the wrong lane's digest | qwen ladder test / krea 1024 test |
| corrupt z_image's historical digest literal | z_image test |

## Scope note

**This does not make #2140 green.** I fetched its regenerated closure config: all three MLX providers moved across that pin bump (`qwen_image` d754→aa8adb88, `krea_2_turbo_control` 65f9→b66a416e, `z_image_turbo` 36e7→25f6eb3a). Since mlx has exactly three entries, "only three MLX closures move" is all of them — those calibrations genuinely expired, and `currentCalibrationRuns` 11 → 0 is correct. The defect fixed here is general; #2140 simply was not a case where it was load-bearing.

## Verification

- `cargo test -p sceneworks-worker --lib` — 1423 passed, 0 failed
- `npm run rust:check` — clean
- `npm run check` — 288/288
- derived docs regenerated (`mlx_fit_gate.rs` is a fingerprinted matrix source); source-tree hash rotation only, no cell changed

Related: sc-17726, sc-17774.


michaeltrefry:
## Merge-queue conflict, and how it was resolved

**Symptom:** the merge queue reported 2175 as `UNMERGEABLE` while the PR page showed no conflict with `main`. Both were correct — they compute against different bases. `main` had not moved from `1e50841a`, but the queue was evaluating against `main` + the entries ahead of it:

| pos | PR | state at the time |
|---|---|---|
| 1 | #2171 — closure digest false greens (sc-17935) | awaiting checks |
| 2 | #2174 — capture-time closure key (sc-17774) | `UNMERGEABLE` |
| 3 | #2175 (this one) | `UNMERGEABLE` |

#2174 was unmergeable too, so this was not specific to 2175. The cause is structural: all three touch a **fingerprinted matrix source**, so all three rotate `sceneWorksRevision` in `docs/generated/memory-matrix.json` and the cost model — on the same line. Any two such PRs conflict textually in files that can only be regenerated, never hand-merged.

**Resolved.** #2171 landed as `5e06b777`; `main` merged in at `453dd000`, with the four derived docs regenerated rather than hand-resolved.

## One thing the conflict flag did not surface

A trial merge against #2171 *before* it landed showed the textual conflicts were only the generated docs — but a test also failed. #2171 bumps `digestVersion` v2 → v3 and restamps every digest (same pin, new derivation), and this PR had hardcoded z_image's pre-restamp digest as a hex literal. `--restamp` is explicitly documented as a legitimate no-op re-derivation, so that literal would have turned a no-op into a red test on any future version bump.

`f2b96c88` replaces it: the expected digest is now read back from the packaged bundle. Mutation-testing that rewrite then exposed something worse — every conjunct in the assertion was true-by-construction of the shipped data, so deleting any one of them changed nothing. It documented state rather than grading it.

So `f2b96c88` also adds the case that does grade it: an opt-in rewritten to claim the **live** digest walks straight past the admission filter (which compares exactly that field) and is stopped only by `EvidenceBundle::evidence_for` finding the record still stamped with the digest it was really measured under. The two comparisons look redundant and are not — one asks *"is this a measurement of current code?"*, the other asks *"is this binding telling the truth about which measurement it is?"*. Defeating the second reds the test.

## Verification after the merge

- `cargo test -p sceneworks-worker --lib` — 1423 passed, 0 failed
- `npm run rust:check` — exit 0
- `npm run check` — 334 passed, 0 failed
- derived docs regenerated and `--check` clean against the post-#2171 sources

Digests confirmed consistent under #2171's v3 restamp: `mlx:qwen_image` and `mlx:krea_2_turbo_control` bindings match their live entries (current), and `mlx:z_image_turbo` still does not (correctly historical).

## Note for whoever picks this up

#2174 is likely still blocked on the same generated-doc collision and will need the same treatment: merge `main`, then `npm run generate:memory-matrix` and `npm run generate:calibration-cost-model` — do not hand-resolve those four files.

PR #2178, opened 2026-08-07
chore(ci): gate scripts/sc-16170-certification.test.mjs

michaeltrefry:
The `node --test` list in `npm run check` is hand-maintained, so a test file never added to it runs nowhere.

#2171 closed this for `scripts/inference-closure-digest.test.mjs`. `scripts/sc-16170-certification.test.mjs` was the last remaining one — referenced by no runner in `package.json` and by no workflow.

Its 2 tests already passed; they cover key-order independence and material source-session/record drift detection in the certification bundle comparison. This is wiring only, no source change.

**334 → 336 tests.** After this, no `scripts/**/*.test.mjs` is unreferenced by a runner — verified by enumerating every `*.test.mjs` on disk against `package.json` plus all workflow files.

Split out of #2174, which was closed as superseded by #2171. This was the one item #2171 did not carry.


michaeltrefry:
Closing on reflection — this adds a standing CI obligation to a frozen artifact.

`scripts/sc-16170-certification.mjs` currently promotes nothing: the evidence bundle holds 0 candle z_image records and 0 sc-16170 source sessions. Those captures predate the schema-v4 `loadShape` axis, so per docs/sc-16170-z-image-cuda-certification.md the 90 cases are validated as historical evidence but deliberately not promoted, and re-certification requires new measurements. Nothing in CI invokes the script.

Gating its unit test would mean anyone touching `canonicalJson` or `validateBundle` owes green tests to a comparator serving a script that produces no output. The script and its 97 logs are worth keeping as an archived derivation; its test is not worth a recurring budget line.

If the orphaned-test-file problem is worth solving, the right shape is a check that enumerates `scripts/**/*.test.mjs` and fails on any not referenced by a runner — forcing an explicit gate-it-or-delete-it decision — rather than wiring them up one at a time.

PR #2200, opened 2026-08-08
fix(loras): forward-port the LoRA install-badge and family-gate fixes to main [sc-18196, sc-18200]

michaeltrefry:
Forward-port to `main` of two hotfixes already merged into `release/next`, as required by `RELEASING.md` invariant 5. Both defects are present on `main` today.

Stories: [sc-18196](https://app.shortcut.com/trefry/story/18196) · [sc-18200](https://app.shortcut.com/trefry/story/18200)
Sources: #2193 (`9591b174`) and #2195 (`30430277`). One PR because both touch `apps/rust-api/src/loras.rs` and would conflict as separate branches.

## sc-18196 — stale install badge (clean cherry-pick)

The API probes `installState` live from the HF cache on every request; the client's badge is a snapshot from its last catalog fetch. Anything filling the cache out-of-band flips the server to `installed` while an open Model Manager still reads "Not Installed", so Download 400s with a message contradicting the badge on screen. Tagged with a `lora_already_installed` code; the client keys off the **code** (not the prose) and resyncs.

This matters more than it looks: Model Manager's Download is the **only** install path, because both studio shells filter uninstalled LoRAs out of the picker entirely.

## sc-18200 — cross-architecture LoRAs rejected at job creation

`scail2_lightning` is a lightx2v Wan2.1-I2V LoRA applied to SCAIL-2 on purpose. Detection correctly reports `wan-video`, but the gate compared it against `loraCompatibility.families` alone and never consulted `extra_compatible_lora_families` — the registry that exists for exactly this relation. Also fixes the same latent bug for Krea Realtime.

## ⚠️ This half is NOT a clean cherry-pick — please read

`main` carries **sc-15017**, which `release/next` does not. `base_model_satisfies_gate` reads the **same registry**, so the new entry acquires a second meaning here:

```rust
extra_compatible_lora_families(&family).contains(&"wan-video")
    && is_wan_14b_class_id(model_id)
    && is_wan_14b_class_id(base)
    && !is_wan_i2v_id(base)          // written for Krea Realtime
```

That last axis is **inverted for SCAIL-2**. Its own rationale says why: *"an I2V LoRA targets `cross_attn.k_img`/`v_img`, which a text-to-video backbone does not have."* True of Krea Realtime. SCAIL-2 is Wan2.1-**I2V**-derived and **does** carry those projections — they're adaptable targets, and the lightning adapter patches them.

Ported naively, SCAIL-2 would have:

| Wan base stamped on the LoRA | Naive port | Correct |
|---|---|---|
| `wan_2_2_i2v_14b` | refused | **admit** — exact architectural match |
| `wan_2_2_t2v_14b` | admitted | admit (same block layout, minus the img stack) |
| `wan_2_2` (5B) | refused | refused — different latent geometry |

So the I2V exclusion is now a property of the **host**, not of every extra-compatible family: `extra_compatible_backbone_is_image_conditioned` splits SCAIL-2 (image-conditioned) from Krea Realtime (text-only). Krea Realtime's behavior is unchanged. Both size classes still gate.

Scope note: the base-model gate only fires when a LoRA **records** a `baseModel`. The lightning catalog entry doesn't, so `scail2_lightning` is unaffected either way — this is about user-imported, base-stamped Wan LoRAs on SCAIL-2.

The sc-15017 test that used `scail2` as its example of a family with *no* extra-compatible entry now names `ltx-video`, since scail2 legitimately has one.

## Verification

- `sceneworks-core` 858 tests, `sceneworks-rust-api` 559 tests, web `src/hooks/` 186 tests — all pass.
- `cargo fmt --all --check` clean; `cargo clippy -p sceneworks-core -p sceneworks-rust-api --all-targets` zero warnings.
- Mutation-checked: treating `scail2` as a T2V backbone fails `scail2_base_model_gate_admits_i2v_unlike_the_t2v_backbone` and nothing else.

## Not included

The inference-pin bump (#2198) is release-line-specific — `main` tracks its own inference revision — so it is deliberately not forward-ported. The inference fix itself needs a separate port to inference `main`, tracked under sc-18198.

### DIFF
--- a/apps/rust-api/src/lib.rs
+++ b/apps/rust-api/src/lib.rs
@@ -38,10 +38,10 @@ use sceneworks_core::jobs_store::{
     RetryJob, RouteDecision, StaleSweep, UnsupportedReason, WorkerHeartbeat, JOB_STATUSES,
 };
 use sceneworks_core::lora_family::{
-    apply_adapter_metadata_to_manifest_entry, apply_model_manifest_defaults, canonical_lora_family,
-    detect_lora_family, detect_model_family, first_safetensors_path, read_adapter_metadata,
-    read_safetensors_header, reconcile_detected_family, AdapterFileMetadata,
-    SafetensorsHeaderError,
+    accepted_lora_families, apply_adapter_metadata_to_manifest_entry,
+    apply_model_manifest_defaults, canonical_lora_family, detect_lora_family, detect_model_family,
+    first_safetensors_path, read_adapter_metadata, read_safetensors_header,
+    reconcile_detected_family, AdapterFileMetadata, SafetensorsHeaderError,
 };
 use sceneworks_core::lora_url::{lora_source_url_file_stem, parse_lora_source_url, LoraUrlError};
 use sceneworks_core::project_store::{
--- a/apps/rust-api/src/loras.rs
+++ b/apps/rust-api/src/loras.rs
@@ -42,7 +42,18 @@ pub(crate) async fn create_lora_download_job(
     if lora.get("installState").and_then(Value::as_str) == Some("installed")
         && lora.get("updateAvailable").and_then(Value::as_bool) != Some(true)
     {
-        return Err(ApiError::bad_request("LoRA is already installed"));
+        // `installState` is probed live from the HF cache on every catalog read, but a
+        // client's rendered badge is a snapshot from its last fetch. Anything that fills
+        // the cache out-of-band — the on-demand pull at first generation, another client,
+        // a manual `hf download` — flips the server to "installed" while an open catalog
+        // view still reads "Not Installed". Tag the rejection so the client can tell this
+        // benign disagreement apart from a real failure and resync instead of surfacing a
+        // message that contradicts the badge it is showing.
+        return Err(ApiError {
+            status: StatusCode::BAD_REQUEST,
+            detail: "LoRA is already installed".to_owned(),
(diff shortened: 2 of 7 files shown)


michaeltrefry:
CI fix pushed (`f0768f75`, amended into the sc-18200 commit).

**Cause:** `main` carries a cross-language drift guard that `release/next` does not — `presetUtils.test.jsx` parses the Rust `extra_compatible_lora_families` match arms off disk and asserts they equal the JS mirror `EXTRA_COMPATIBLE_LORA_FAMILIES` in `presetUtils.js`. Adding the `scail2` arm on the Rust side without the mirror made the tables unequal.

**Not cosmetic.** The mirror's stated purpose is keeping the LoRA picker from being *stricter* than the backend — "it would hide a LoRA the backend would happily run". Without the entry the picker would have hidden Wan-family LoRAs on SCAIL-2 that the job-creation gate now accepts, which is the same class of invisible-capability bug this whole thread has been chasing.

This is the second main-only interaction the forward-port had to satisfy (the first being `base_model_satisfies_gate`'s I2V axis). Both trace to sc-15017, which isn't on `release/next` — so nothing latent was shipped to the release line; I verified the mirror is absent there.

**Verification:** ran the full web suite locally rather than bouncing off CI again — 237 files / 3542 tests pass, matching CI's counts. Rust side unchanged by this commit (858 core + 559 API still green).

PR #2208, opened 2026-08-09
chore(check): fail on evidence-corpus shrink without an explicit tombstone (sc-18224)

michaeltrefry:
### Problem
After sc-18100 deleted the one-shot census assertions, every remaining consistency check verifies that calibration artifacts agree **with each other** — bundle vs manifest vs matrix. None verifies the evidence corpus hasn't **shrunk**. A bad merge, an overzealous regeneration, or an agent deleting an offending record and regenerating the matrix to match all left a perfectly self-consistent tree and passed CI.

### Gate design
`scripts/check-evidence-corpus.mjs`:

- **What it compares:** the set of record ids in `docs/generated/memory-calibration-evidence.json` at HEAD (working tree, so uncommitted deletions red locally too) vs the same file at `git merge-base HEAD origin/main`. Any id present at the base but absent at HEAD → **FAIL**. Growth never fails.
- **Escape hatch:** `config/memory-evidence-tombstones.json` — one entry per deleted id with `reason` + authorizing `story` (`sc-NNNNN`). The entry is the reviewable act: a PR deleting evidence must also touch this file. The gate **also fails on a tombstone whose record still exists** at HEAD, so premature/stale entries can't accumulate; tombstones for long-merged deletions remain as history.
- **CI reality:** both lanes running `npm run check` (`check.yml` parity, `publish-runpod.yml`) use actions/checkout's default `fetch-depth: 1` — neither `origin/main` nor HEAD's parents exist locally. The gate fetches `origin main` itself, then deepens **both** sides (`git fetch --deepen origin` widens the checkout's own configured refspec — the PR merge ref / merge-queue ref / tag — and a second fetch widens `origin/main`), escalating to `--unshallow`, and **fails closed** if no merge-base is determinable ("cannot see the base" is exactly the condition a deletion would want). On `main` itself the merge-base **is** HEAD, so the gate is inert post-merge. On a pull_request/merge_group checkout HEAD is the speculative merge (contains main), so a stale branch never false-fails on records main added later. `SCENEWORKS_EVIDENCE_BASE` / `--base` provide an explicit override.

### Wiring
- `npm run check` runs `--self-test` (proves the gate fires, against the real bundle, without a broken checkin) then the live check, ahead of the matrix regeneration checks.
- `scripts/check-evidence-corpus.test.mjs` added to the hand-maintained `node --test` list; a test in the suite asserts this wiring so it can't silently drop out.
- New `check:evidence-corpus` npm script for direct use.

### Tests (all green)
11 tests: unit coverage of id parsing, tombstone validation, and comparison semantics, plus **end-to-end scratch-git-repo tests**: full clone red/green, uncommitted working-tree deletion, post-merge inertness on main, an explicit-override case, and a **shallow `--depth=1 --single-branch` clone replicating the CI checkout shape** — the gate fetches its own base over the wire and reds. Full `npm run check`: exit 0, 354 node tests pass.

### Mutation evidence (real bundle, working tree)
1. Deleted record `imc-033814e110b9c190da59` from the bundle, no tombstone → **exit 1**: `evidence record imc-033814e110b9c190da59 exists at the merge-base but not at HEAD and has no tombstone...`
2. Added a tombstone for it → **exit 0**: `evidence corpus ok: 64 records at HEAD, 65 at base dff56056bf88, 1 tombstone`
3. Restored the record but kept the tombstone → **exit 1**: `...declares imc-033814e110b9c190da59 deleted, but the record still exists...`
4. Restored both → `evidence corpus ok: 65 records at HEAD, 65 at base dff56056bf88, 0 tombstones`


michaeltrefry:
CI verification of the gate in the real shallow checkout (parity job, run 31311555024 — fetch-depth 1, Check workflow green):

```
check-evidence-corpus self-test passed
evidence corpus: origin/main not resolvable locally, fetching it
evidence corpus: histories still disconnected, deepening by 64
evidence corpus: histories still disconnected, deepening by 1024
evidence corpus: unshallowing to find the merge-base
evidence corpus ok: 65 records at HEAD, 65 at base dff56056bf88, 0 tombstones
```

The whole base-resolution sequence took ~20s inside the parity job; the escalation ended at `--unshallow` because actions/checkout's narrow refspec keeps the PR merge ref itself shallow through the deepen steps. Fails closed if no merge-base is ever found.

The earlier `build-windows` failure was a hosted-runner flake (`cl.exe` exit 0xc0000006 STATUS_IN_PAGE_ERROR while compiling vendored sqlite3.c — no Rust touched in this diff); rerun requested.

[author: claude]

michaeltrefry:
Review fixes pushed (17da8a46):

1. Unreadable base blob now fails closed. Absence of the bundle at the base is confirmed via `git ls-tree` on the base tree; a present-but-unreadable blob (corrupt object / partial-clone promisor failure) throws instead of degrading to an empty base set. Reproduced in a new e2e test: `fetch.unpackLimit` explodes the fetched pack to loose objects, the base blob object is deleted, and `runCheck` rejects with "cannot be read... failing closed" — previously baseIds=∅ / errors=0. A second new test covers the legitimate confirmed-absent (pre-bundle base) growth-only case.

2. The wiring test now pins the exact `--self-test && node scripts/check-evidence-corpus.mjs` sequence. Mutation-verified: deleting the live invocation from the check line reds the new regex (the old suffix regex stayed green on that mutation).

Also merged origin/main and PR #2206's branch (sc-18227) and resolved the `check` line union — both PRs' test-list entries present — so the merge-queue conflict between the two package.json edits is pre-resolved from this side. #2206's `test-list-coverage` meta-test now pins `check-evidence-corpus.test.mjs` too and passes. Full `npm run check`: exit 0, 363 node tests.

[author: claude]

PR #2218, opened 2026-08-09
feat(krea): serve pose ControlNet on imported single-file Krea 2 checkpoints

michaeltrefry:
Depends on **SceneWorks/inference#533** (the pin here points at that branch's head, `014134e3`).

## Why

Imported Krea-2 fine-tune checkpoints (a single-file `.safetensors` DiT, e.g. KreaMania) rejected pose
ControlNet conditioning on every backend — the claim gate refused any non-empty pose set, the scheduler
mirrored it, and the UI therefore greyed the pose picker.

That was a **lane-scope restriction, not a structural incompatibility**. The trained pose branch
(`Krea2ControlBranch`, provider `krea_2_turbo_control`) is a `control_scale`-scaled, RMS-clamped
residual folded onto the frozen DiT at load — architecturally independent of the DiT's own weights —
so it composes with a same-shape imported fine-tune exactly as it does with the builtin Turbo base.

## What

**Worker.** `KREA_IMPORTED_SUPPORTS_POSE_CONTROL` mirrors the backend capability;
`krea_imported_available` admits a pose shape on a capable backend; a new
`ImageRoute::KreaImportedControl` (one image per pose, claimed *ahead of* the plain imported arm so a
pose set can never fall into per-image t2i and silently drop the poses) drives
`generate_krea_imported_control_stream`. That stream is the builtin control lane's render loop over a
natively-loaded imported DiT: same DWPose preprocessing, same shared seed across the set, same
control-scale clamp, same identity-likeness scoring, same telemetry fields.

**Scheduler.** `imported_image_request_family_eligible` now takes a named
`ImportedImageBackendCaps { adapters, pose_control }` instead of a positional bool, so a call site
cannot silently transpose capabilities. Pose is admitted for the `krea_2` family on MLX; shapes the
pose loop would silently drop (the plural edit-reference set, a bare `sourceAssetId`, edit mode) stay
rejected on every backend.

**UI affordance.** The imported model's `macSupport.features.pose` is now *derived* by probing the
same claim gate the scheduler runs, rather than a hardcoded flag — so the picker cannot drift from
what actually routes, and an `sdxl`/`mage-flow` import keeps the pose picker dark.

**Candle** keeps rejecting pose (its single-file path threads neither adapters nor control, sc-14135),
and that refusal is **terminal, not a silent never-route**: `candle_supported` classifies it as a
named gap, so the enforce sweep fails such a job with a reason instead of leaving it queued forever. A
new test pins both halves together, so a future change cannot go green by dropping the surface on
*both* backends.

## Two things the guards caught, worth calling out

**A new job-time download site.** Reusing the builtin lane's `ensure_krea_control_weights` would have
added one, which epic 17625 AC9 forbids (`job_time_download_guard` failed the build). The overlay and
the identity face stack are now resolved **cache-only** via `resolve_hf_component_file`: a missing
overlay is an actionable "install it from the Model Manager" error raised before any compute, and a
missing face stack costs the optional likeness scores, never the render.

**The measured envelope is deliberately NOT reused.** Every measured `mlx:krea_2_turbo_control`
record in the corpus was captured at the **q4** base tier (fixture `krea-pose-control-q4-seed16099`).
This composition is **dense** — the native loader materializes the single file bf16 — so the imported
12B DiT is ~24 GiB resident against the ~6 GiB the q4 record measured. Same architecture is not the
same footprint once the tier differs, and this gate's permissive-side failure mode is an OS Jetsam
SIGKILL. The lane therefore passes `None` for the resolved-artifact provenance on purpose and runs
the conservative estimate path, with the estimation spec pointed at the dense `bf16` tier so the byte
estimate matches the real assembly. *Capturing a dense-tier control calibration is the follow-up that
would let this lane serve measured numbers.* Independently, the generator's own measured,
architecture-keyed `control_geometry_fits` check still guards the render exactly as it does the
builtin lane.

Note the pin bump demotes no calibration: `mlx:krea_2_turbo_control` was already stale beforehand
(captured closures `3064f675`/`d355971e` vs live `8717a575`).

## Validation

- `sceneworks-core` 860 lib + 170 `jobs_store` integration tests, `sceneworks-worker` 1461 lib tests —
  all green, including all 22 `job_time_download_guard` tests.
- `npm run rust:check`, `npm run rust:check:candle` (the candle cfg caught and fixed a real
  `E0425` this branch introduced), `npm run check:rust-derived-docs`, `cargo fmt --all --check`,
  `check-license-coverage` — all clean.
- **End-to-end on this M5 Max** (`krea_imported_control_mlx_gpu_smoke`, real 24 GiB imported
  KreaMania checkpoint + the pinned pose overlay): route resolves to `KreaImportedControl`, the
  branch assembles onto the file-loaded DiT, and three 1024² renders land with two negative controls —
  vs its own `control_scale = 0` passthrough (delta **47.0**) and vs a **different** skeleton (delta
  **40.2**), the latter proving the render tracks the pose *input* rather than merely that some
  residual fired. Both renders visibly match their skeletons (arms down at the sides; arms raised with
  a widened stance).


michaeltrefry:
## ⚠️ One check needs a CUDA box — `capabilities.candle.json`

The `web` lane will stay red on two assertions until someone re-dumps the candle **media** capability
facts off-Mac. This is a property of any inference pin bump in this repo, not of this change
specifically — the facts files are pin-keyed and each lane owns its own dump.

**What I could do here** (done, committed): `capabilities.mlx.json` re-dumped by actually running the
dumper on a Mac, and `builtin.preview-support.jsonc` regenerated.

**What needs your Windows/CUDA box** — one command, then commit both files it writes:

```bash
cargo run -p sceneworks-worker --bin dump-engine-capabilities --no-default-features --features backend-candle
```

then, from a Mac or anywhere:

```bash
cd apps/web && npm run gen:preview-support
```

**Why I couldn't do it here:** running that exact command on this Mac writes `capabilities.mlx.json`,
not the candle file — the candle media lanes are `cfg(all(not(target_os = "macos"), feature = "backend-candle"))`,
so nothing candle-side links into a macOS build. I verified this by running it rather than assuming.

**Why the candle *audio* dump is also left at the old revision** (I re-dumped it, then deliberately
reverted): a backend split across two inference revisions makes `derivePreviewSupport` refuse to
derive at all — "backend candle was dumped at two different inference revisions". Leaving both candle
files at the old pin keeps the backend internally consistent and `gen:preview-support` working, and
the off-Mac dumper rewrites both in the single command above.

Remaining red assertions, for reference:
- `capabilities.candle.json was dumped at the revision the worker currently pins`
- `the manifest stamps each backend's dump revision, and it is the current pin`

Everything else on this PR is green locally, including the `parity` audit and all 22
`job_time_download_guard` tests.

michaeltrefry:
## 📋 Shortcut story text (not filed — no Shortcut MCP in this session)

I could not create the story: the Shortcut MCP server is configured in `~/.claude.json` but is not
connected in this session, so no Shortcut tools were available. Handing off the text rather than
reporting the workflow step as done.

**Type:** Feature · **Suggested epic:** 14015 (ALT community checkpoints import — where
sc-14018/14023/14071/14111/14119 live). Cross-reference epic 8459 S5 / sc-8465, which built the pose
control branch this reuses.

**Title:** Serve pose ControlNet on imported single-file Krea 2 checkpoints (MLX)

**Description:**

Imported Krea-2 fine-tune checkpoints rejected pose conditioning on every backend. That was a
lane-scope restriction, not a structural incompatibility: the trained pose branch is a
`control_scale`-scaled RMS-clamped residual folded onto the frozen DiT at load, architecturally
independent of the DiT's weights, so it composes with a same-shape imported fine-tune exactly as with
the builtin base.

Delivered across SceneWorks/inference#533 and SceneWorks/SceneWorks#2218:

1. **inference** — `load_control_from_native_dit_file`, the control sibling of the t2i native
   entrypoint; native control memory contract keeping the `krea_2_turbo_control` provider identity;
   the native-file resident projection hoisted to one shared `native_dit_transformer_bytes`.
2. **worker** — `KREA_IMPORTED_SUPPORTS_POSE_CONTROL`, pose admitted in `krea_imported_available`,
   `ImageRoute::KreaImportedControl` (one image per pose) ahead of the plain imported arm, and
   `generate_krea_imported_control_stream`.
3. **scheduler** — `ImportedImageBackendCaps { adapters, pose_control }` replacing a positional bool;
   pose admitted for `krea_2` on MLX; the UI pose affordance derived by probing the claim gate.
4. **candle** — still rejected (sc-14135), and terminally so via `candle_supported`, not a silent
   never-route.

**Findings worth recording on the story:**

- The premise that measured evidence carries over is **wrong for this corpus**: every measured
  `mlx:krea_2_turbo_control` record was captured at the **q4** tier, while this composition is dense
  (~24 GiB vs the ~6 GiB measured). The lane runs the conservative estimate path on purpose.
  *Worth a follow-up story:* capture a dense-tier control calibration — though note it would also
  need an evidence-reuse policy decision, since an imported DiT has no pinned artifact identity.
- **No calibration was demoted:** `mlx:krea_2_turbo_control` was already stale beforehand.
- **Epic 17625 AC9:** reusing the builtin download helper would have added a new job-time download
  site; the lane resolves the overlay and face stack cache-only.
- **Every pin bump needs an off-Mac `capabilities.candle.json` dump** (see the comment above) — worth
  its own tracked chore if it keeps biting.

**Validation:** inference 160 unit tests + clippy clean; SceneWorks core 860 + worker 1461 lib tests,
`parity` audit, candle typecheck, all 22 download-guard tests; and an end-to-end MLX GPU render on an
M5 Max with a real 24 GiB imported checkpoint (pose-locked vs scale-0 delta 47.0, vs a different
skeleton 40.2, both renders visibly matching their skeletons).

michaeltrefry:
### Correction + the measured trade-off behind leaving the candle audio dump at the old pin

My earlier comment said two assertions stay red. It is **four signals**, all one cause — the candle
facts (media *and* audio) need the off-Mac dump:

- `web` → `capabilities.candle.json was dumped at the revision the worker currently pins`
- `web` → `the manifest stamps each backend's dump revision, and it is the current pin`
- `web` → `audio/capabilities.candle.json is a real dump: … at the current pin`
- `nax-worker` → `Verify capabilities.mlx.json is a real dump, not a restamp` (fails on its *audio*
  half; the mlx half matches)

**There is no locally-reachable green state**, and I measured both candidates rather than reasoning
about them:

| state | `nax-worker` | `web` |
|---|---|---|
| audio at **new** pin | ✅ passes | ❌ **whole file dies** — `derivePreviewSupport` throws on the split-revision candle backend at import, so 0 of 51 tests run |
| audio at **old** pin (shipped) | ❌ one step | ❌ 3 of 51 assertions; **48 still pass** |

I shipped the second: keeping the candle backend internally consistent preserves 48 real assertions
of coverage and keeps `gen:preview-support` runnable, at the cost of one extra red step. Trading 48
live assertions for one green step would be a bad deal, and the "green" would only be hiding the same
missing dump.

The single off-Mac command in the comment above rewrites **both** candle files and clears all four.

michaeltrefry:
### The candle dump will be a content no-op — and I am deliberately not restamping it

`candle-worker` now fails too, on the same cause (`Verify capabilities.candle.json is a real dump`),
so **three lanes** — `nax-worker`, `web`, `candle-worker` — are red for the one missing off-Mac dump.

Useful detail from that run: the CUDA box produced a fresh dump and printed the complete diff against
the checked-in file. It is **exactly one line**:

```
@@ -1,7 +1,7 @@
-    "inferenceRevision": "40fa7583a01974617e2a7275052d6d446688c956"
+    "inferenceRevision": "014134e3035ad7e4eca5c2ed7bded2375dc3c071"
```

So the candle media registry's descriptor content is **unchanged** across this pin — expected, since
inference#533 adds an entrypoint, not a descriptor. Whoever runs the dump should expect a one-line
result, and a larger diff would mean something else moved and deserves a look.

I could have hand-written that line and turned all three lanes green. **I did not**, because both
`macos-mlx.yml` and `windows-candle.yml` say explicitly, twice, *"Do NOT just rewrite the
inferenceRevision line."* That prohibition is clearly aimed at exactly this situation — the authors
anticipated a one-line diff and pre-refused it — and "it's only the revision line **this time**" is
the precise rationalization that would erode the guard for a future pin where descriptors really did
move. Overriding a deliberate repo policy is a maintainer's call, not mine.

If you'd rather I just write the line, say so and I will. Otherwise the one command on the CUDA box
(in the comment above) resolves all three.

michaeltrefry:
BLOCKING cross-PR hazard for SC-18345: this PR current remote head 050b0439795d6b2a9c4449e6d111cf8a0d79fc32 still hand-builds the imported Krea pose MlxRequestInputs with has_reference=false/reference_count=0 while the live request sends one Conditioning::Control. That is the same live refusal fixed by #2220, so merging this head after #2220 would reintroduce the defect for imported checkpoints. Do not merge this head. I have a local corrective change that routes the imported call site through the shared production helper and source-pins both call sites; it will be synchronized onto this PR only after #2220/current main land, then regenerated, fully validated, independently reviewed, and rerun through CI. The current parity failure is also unresolved and independently blocks merge.

michaeltrefry:
[author: codex]\n\nOwnership update: this PR is now tracked by SC-18382 under the imported-checkpoints epic, and I am carrying it through repair, fresh review, CI, merge queue, and exact-main verification.\n\nCurrent live head a2b4d674f3f7f9899179da670ab3f41743f21db6 still contains the blocking imported pose declaration with has_reference=false/reference_count=0 while emitting one Control. Green CI does not make that geometry correct. Do not merge this head.\n\nMerge order is intentional: PR 2220 establishes the reviewed canonical Krea pose helper first. After it lands on main, I will synchronize this exact live branch, replace the imported hand-written declaration with that helper, add a source-pinned cross-call-site regression, regenerate pin-derived artifacts, and publish only after a fresh independent review passes.\n\nTracker: https://app.shortcut.com/trefry/story/18382

michaeltrefry:
[author: codex]

Repaired head `b631e4bcce04847614e98a48a639e5cfad4f43fb` is now published. The previous do-not-merge warning is resolved at this head.

What changed during takeover:

- synchronized the exact reviewed #2220 landing from `main` and resolved the old branch conflict;
- replaced the imported MLX lane's hand-written zero-reference declaration with the shared `krea_control_memory_inputs` geometry (`control:1`, `has_reference=true`, `reference_count=1`);
- added a source-pinned regression requiring both builtin and imported production call sites to use the shared helper;
- corrected the Candle Krea control identity to match pinned inference: `image_to_image` with one reference across evidence selection, runtime scope, production registration, and tests;
- retained inference pin `014134e3035ad7e4eca5c2ed7bded2375dc3c071` from merged inference PR #533 and regenerated dependency closure, capability facts, preview support, and memory-matrix artifacts at that pin;
- corrected stale route documentation found during independent review.

Local validation on the final tree includes full `npm run check` (374/374), strict MLX and macOS worker clippy, the repository's Candle cross-target check, generated-artifact checks, Python memory-matrix tests (9/9), focused imported-Krea/core/geometry tests, formatting, and diff cleanliness. I did not rerun the original 24 GiB real-weight smoke; the existing PR records that earlier run.

Independent review passed the exact final head with high confidence and no findings. Fresh GitHub checks are now running. I will enroll this head in the merge queue only after required checks pass, then verify exact `main` ancestry before closing SC-18382 and SC-18345.

Tracker: https://app.shortcut.com/trefry/story/18382

PR #2230, opened 2026-08-10
test(worker): drive the imported pose smoke through the real memory gate

michaeltrefry:
## Why

The imported pose-control lane (sc-18406 / #2218) shipped with `has_reference: false` / `reference_count: 0` in its `MlxRequestInputs`, mirrored faithfully from the builtin Krea control lane — which had the same values. Both were wrong: gen-core charges `Conditioning::Control` as **one image reference** and re-derives the geometry at `configure_request`, so every real pose render was refused:

```
krea_2_turbo_control: request geometry 1024x1024x1 references=1 does not fit admitted 1024x1024x1 references=0
```

sc-18377 fixed it via the shared `krea_control_memory_inputs` constructor and added a static guard grading the declaration against `image_reference_count()`.

## The gap this closes

**Nothing caught it because nothing ran the composition.** Every piece was individually tested and individually plausible — the claim predicates were unit tested, the declaration was a plain struct literal, and the GPU smoke proved the engine renders. But the smoke called `load_control_from_native_dit_file` and `model.generate` **directly**, bypassing `mlx_fit_gate::evaluate_request` and `memory_strategy::generate_with_scope` — exactly where the declaration is graded.

So it validated the architectural claim (a pose branch folds onto an imported DiT and the output tracks the skeleton) while the user-facing claim (submit a pose set, get pixels) went unexercised.

The smoke now admits through the real fit gate against the **loaded provider** and renders through the same `generate_with_scope` seam `krea_control_generate_one` uses, building the plan and inputs exactly as `generate_krea_imported_control_stream` does. The sc-18377 guard grades the declaration statically; this grades it where `MlxRequestScopeCore` actually runs.

## Verified discriminating by mutation, not assumed

With `KREA_CONTROL_REFERENCE_COUNT` set back to `0`:

| variant | result |
|---|---|
| both new assertions removed | fails in **0.91s** with the exact historical refusal — so the **gate**, not just the test, rejects it |
| assertions present | fails in **0.89s** before any compute — faster, more legible signal |

Restored, all three renders pass through the gate on real weights (M5 Max, 24 GiB imported checkpoint):

- pose-locked vs `control_scale=0` passthrough: **44.574**
- arms-down vs arms-up skeleton: **36.585**

Both far above the noise floor, so the pose-tracking assertions the smoke already made still hold on the gated path.

## Checks

`cargo fmt` clean · `clippy -p sceneworks-worker --all-targets` 0 warnings · `krea_control` unit tests 4/4 · the `#[ignore]` smoke passes on real weights in 241s.

Related: [sc-18406](https://app.shortcut.com/trefry/story/18406), sc-18377.

### DIFF
--- a/crates/sceneworks-worker/src/image_jobs/tests.rs
+++ b/crates/sceneworks-worker/src/image_jobs/tests.rs
@@ -14960,6 +14960,37 @@ fn krea_imported_control_mlx_gpu_smoke() {
         "the imported control assembly keeps the builtin provider identity"
     );
 
+    // The MEMORY GATE the lane really runs, on the real generator — not a unit fixture.
+    //
+    // This is the composition the shipped defect lived in. Every piece was individually plausible:
+    // the lane declared its `MlxRequestInputs`, the engine rendered when driven directly, and the
+    // route/claim predicates were unit-tested. What nothing exercised was generator + gate +
+    // render TOGETHER, so a declaration that disagreed with the live request — `references=0`
+    // against the one reference gen-core charges for `Conditioning::Control` — passed every test
+    // and refused every real render with `request geometry ... references=1 does not fit admitted
+    // ... references=0`. `krea_control_declares_the_reference_count_gen_core_derives_from_its_request`
+    // now grades the declaration statically; this grades it against a LOADED provider, which is the
+    // only place `MlxRequestScopeCore::configure_request` actually runs.
+    //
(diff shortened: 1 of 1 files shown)


michaeltrefry:
[author: codex] Epic coordination: temporarily removed from the merge queue. This is a valuable real-gate regression for the imported Krea pose lane, but it was queued without the independent exact-head review required by the epic workflow. I am taking it into the owned #2218/#2220 lane, will review and synchronize it against current `main`, then re-enroll it if clean.

PR #2332, opened 2026-08-14
feat(catalog): express an exact-value step constraint and enforce it on both lanes (sc-19502)

michaeltrefry:
Pairs with **inference PR [#610](https://github.com/SceneWorks/inference/pull/610)** — merge that one first.

## Problem

sc-19426 added `limits.hardMinSteps` and deliberately left LTX-2.3 at `None`, recording it as *"a floor-shaped hole this key cannot fill"*. Two separate defects:

1. **Unexpressible.** LTX is distilled — its σ waypoints are baked into training, so it renders at **exactly 8** steps. `hardMinSteps: 8` says "at least 8" and wrongly admits `steps: 30`, which the engine refuses.
2. **The lanes disagreed.** candle rejected `steps: 30`; `mlx-gen-ltx` **never read `req.steps` at all** — accepted, knob inert, baked 8-step schedule rendered anyway.

So the Advanced **Steps** control was binding on one backend and silently inert on the other, and off-schedule counts failed *after* dispatch rather than at enqueue.

## AC 1 — the shape: an allowed-SET key, not a per-lane block

`limits.steps: [8]`, membership-tested, following the `limits.fps` precedent. Says "exactly 8" rather than "at least 8", and generalizes to two baked schedules without becoming a range.

Per-lane was the story's other candidate. Reading both engines settled it: **the two lanes bake a byte-identical `STAGE1_SIGMAS` table** (same 9 σ waypoints ⇒ same 8 steps), so the divergence was an mlx **defect**, not a real per-lane difference. #610 moves the constraint onto the shared `Capabilities::supported_steps` seam both lanes' `validate` already calls, so they refuse identically. A per-lane key would have encoded the bug in the schema and made it permanent — `both_ltx_lanes_agree_on_one_legal_step_count` asserts no per-lane form appears.

## AC 2 — declared, enforced, reachable

- **`allowed_steps()`** reads `limits.steps`, mirroring `allowed_fps` including its *unsatisfiable means no menu* tolerance (a typo'd manifest must not brick a model).
- **`steps_limit_error()`** gains the menu arm rather than becoming a second function — so the API (`generation.rs`) and worker (`video_jobs/mod.rs`) gates that **already call it** became readers with **no new wiring**. A parallel function would have been a second thing to remember at two seams, which is exactly how a declared key becomes an inert one. The menu governs when both axes are declared.
- Declared for `ltx_2_3` / `ltx_2_3_eros`, registered in `model-manifest.schema.json` with its `x-sceneworks-contract` classification + anchored readers, mirrored into `constants.js`.
- **Catalog passthrough asserted on the wire**, so the UI pin cannot silently stop seeing the key.

## AC 3 — the lane divergence

Resolved in #610. The distilled schedule genuinely cannot be resampled without going out-of-distribution, so the honest fix is an **explicit refusal on both lanes** — the knob is refused clearly, never silently ignored.

## AC 4 — the control

**Pinned, not hidden.** When a model declares exactly one legal value the Steps input is disabled with the value visible and a title explaining why. This is a *third* case in the screen's existing taxonomy: not Lightning's transient disable, not `supportsGuidance`'s hide of a genuinely-absent axis. The value is real and worth knowing; it is simply not the user's to move. A stale override left by a previously-selected model is suppressed so it can never leak into the payload.

**Menu-shaped when the menu has more than one entry** (review follow-up, `4869693a4`). The first pass stopped at the singleton and let a two-or-more-entry menu fall through to the free-text number box — while `steps_limit_error` refuses off-menu counts by **membership**, not by a floor. That is the same *"UI looser than the gate, offering a knob the backend will refuse"* desync this story exists to remove, in a different shape. Latent today (nothing shipped declares a multi-entry menu), but every other reader on the seam is already set-shaped — `allowed_steps`, `humanized_number_menu`, `stepsMenuFromModel`, the Preset Manager's `checkInMenu`, gen-core's `supported_steps` — so the studio was the single place that would have silently reopened the defect. It now renders a `<select>` over exactly the declared set, following the in-file `fpsOptions` precedent; the tooltip states the legal set in the gate's own phrasing.

The picker's cleared option keeps the Advanced panel's `cleared values → model default` semantics and is deliberately **not** `limits.steps[0]` — that is not a default. It shows `defaults.steps`, which the Rust manifest fixed-point invariant already pins onto the menu, so the cleared state is provably renderable. `stepsOffMenu` is the multi-entry half of the same suppression `stepsPinned` performs: a stale override displays as the cleared entry and never reaches `advanced.steps`. The `length === 1` pinned treatment is untouched.

**Also found and fixed:** the Preset Manager validated `steps` with only a generic 1–200 range. A preset stores *defaults*, so an off-menu value there was not one bad job but **every job it drove** — savable, then guaranteed to 400. Now menu-checked with the same `checkInMenu` fps uses.

## AC 5 — tests

14 added/extended across core, API, worker and web. Every new guard proven **individually** by mutation:

| mutation | red |
|---|---|
| delete the menu check | menu tests + shipped-manifest survey |
| **degrade the menu into a floor** (the sc-19426 shape) | the over-menu half — the exact gap this key fills |
| drop the manifest declaration | survey + cross-lane guard |
| add a per-lane declaration | **only** the cross-lane guard |
| delete the API gate | the enqueue test |
| delete the worker gate | the preflight test |
| un-pin the control | the pin tests |
| drop the payload suppression | **only** the stale-override test |
| gate the preset check on `isVideo` | **only** the image-lane test |
| reader accepts numeric strings | **only** the unusable-declaration test |
| **revert the multi-entry picker to the free-text box** (`stepsChoice = null`) | **only** the two new multi-entry tests — all 105 pre-existing tests in the trio stayed green, which is what makes the singleton-only gap invisible without them |
| drop the multi-entry payload suppression (`stepsOffMenu = false`) | **only** the multi-entry payload test, with `expected 30 to be undefined` — the picker-shape test stayed green, so the two guards are pinned separately |

The stale-override test drives the real path (type on an unpinned model → switch to LTX) rather than typing into an already-disabled box, which would have passed regardless. The multi-entry payload test does the same and additionally asserts an **on**-menu pick *does* reach `advanced.steps`, so a blanket suppression cannot pass it.

**Suites:** `sceneworks-core` 895 ✅ · `sceneworks-worker` 591 ✅ · `sceneworks-rust-api` 1478 ✅ · web 3638 ✅ · pytest 101 ✅ · `npm run check` 424 ✅ · eslint clean · `cargo fmt --all` clean.

*(Rust/pytest counts are from the pre-review run; `4869693a4` is `apps/web`-only, so the re-run after it covered web 3638 ✅ across 245 files, `npm run check` 424 ✅, and eslint clean.)*

`docs/generated/memory-matrix.{json,md}` is a regenerated fingerprint only (the manifest sha moved); no matrix content changed.

**Not run:** real-weight suites — another agent holds the machine's single real-weight slot. Everything here is manifest/gate/UI logic with no weights on the path.

## Follow-up

`Capabilities` still has no way to advertise a step **ceiling** (SVD's `MAX_STEPS = 200` remains undiscoverable weights-free). Out of scope here; noted for a separate story.


michaeltrefry:
Pushed a fix for the red `parity-rust` lane (`parity` is just its aggregate gate, so both go green together).

`clippy::type_complexity` rejected the survey table's three-column tuple:

```
error: very complex type used. Consider factoring parts into `type` definitions
    --> crates/sceneworks-core/src/video_request.rs:2319:24
     |
2319 |     const STEP_FLOORS: &[(&str, Option<u32>, Option<&[u32]>)] = &[
```

Now a named `StepConstraints` alias, which reads better anyway — the two `Option`s are different axes (floor vs menu), not a pair.

My fault for pushing without it: I verified with per-package `cargo test`, which does not run clippy. Re-verified with the repo's own gates this time:
- `npm run rust:fmt` (`cargo fmt --all -- --check`) → clean
- `npm run rust:lint` (`cargo clippy --all-targets -- -D warnings`) → **exit 0** across the whole workspace
- `npm run rust:test` (`cargo test`) → **exit 0**, full workspace
- `npm run check` 414 ✅ · web 3636 ✅ · pytest 101 ✅

Note the `-m parity` snapshot suite errored locally on a 30s API-bind timeout under concurrent build load — a harness startup issue, not a contract mismatch. `check-linux` and `check-macos`, which run that same suite, both passed on the previous push with the manifest change already in place.

[author: claude]

PR #2407, opened 2026-08-17
feat(sc-19707): prefer valid local resolved artifacts across every SceneWorks model runtime

michaeltrefry:
sc-19707 — prefer valid local resolved artifacts across every SceneWorks model runtime.

Base: `feature/sc-19703-resolved-model-hot-cache` (merged in; the PR was CONFLICTING).

## The blocker, and why the fix lives at the guard

The preference overlay is a **directory-level seam**: it answers `(repository, revision)` with a
snapshot path and can never re-ask *"does this bundle hold what THIS caller needs"* once the path is
out. So a per-model decision could not be made safe. A job carrying `owner/matrix` q4 (bundled) and
q8 (not bundled) admitted q4 as LocalReady, installed the overlay for the shared pair, and then
resolved the **q8 load into the q4 bundle root** — which has no `q8/` subtree at all.

`RuntimeSourceGuard::begin` now decides in three passes, in the only layer that can see every model
a job will load:

1. **Pass 1** — resolve every entry with local candidates, keeping the **selected closure** beside
   each resolution. (`ModelResolution::local_ready` deliberately carries `requirements: Vec::new()`,
   so the closure cannot be read back off the resolution — every locally-resolved model would
   otherwise look like it required nothing.)
2. **Pass 2** — build `(repo, revision) -> union of required files` across **all** resolutions,
   local *and* non-local, and select a serving entry only when `entry_covers(entry, union)`. The
   non-local sibling is what poisons a pair. A requirement with `revision: None` makes the whole
   repository unserveable.
3. **Pass 3** — a model stays LocalReady only if **every** pair in its closure was selected; all
   others re-resolve with `&[]`.

Exactly **one** scope is installed, holding exactly the selected entries.

**Deliberate consequence, named in a test** (`a_pair_one_sibling_selection_does_not_cover_is_served_to_neither`):
in a q4+q8 job the **q4 model also falls back to source**. That is the correct conservative answer —
partial service of a pair is exactly what cannot be made safe at a directory-level seam.

## Scope vs each recorded item

| Item | State |
|---|---|
| Base merge (PR was DIRTY) | **done** — one conflict in `resolved_cache.rs` (HEAD's `#[derive(Debug)]` on `ResolvedCacheLease` vs base's `mark_interrupted_on_drop` impl); both sides kept, rebuilt after merging |
| Compile errors (`held.entry.*`, stale `prefer_local_artifacts` callers, deleted `ensure_covers`) | **done** |
| **Blocker 1** — overlay keyed on `(repository, revision)` with no file-coverage check | **done** — three-pass guard selection above |
| **Blocker 2** — fail-closed fallback did not close | **done** — one scope installed only after the whole decision; on refusal nothing is installed, every lease is dropped, and every locally resolved model re-resolves against source |
| **major** `valid_local_artifacts` dropped layout-unsupported entries with a bare `continue` | **done** — returns `LocalArtifactScan { artifacts, rejections }`; the guard emits `resolved_cache_local_tier_unsupported` from the rejections. Coverage/selection failure gets its **own** class, `resolved_cache_local_tier_not_selected` |
| **major** no test asserts `resolved_cache_local_tier_unsupported` / `model_source_tier_selected` | **done** — three new emission tests |
| **minor** blocking `lock_shared` on the job's async block | **done** — `begin` runs on `spawn_blocking`; `acquire_complete`'s lock semantics untouched (shared with sc-19710) |
| **minor** `valid_local_artifacts` swallowed every failure | **done** — every rejection class logs a reason |
| **minor** worker `with_local_cache` tests took no overlay mutex | **done** — `RUST_TEST_THREADS=1` made a green run prove nothing about the process-wide overlay |

## Fixed in passing

`select_local_tier` leases **only the winners**, not every candidate. Leasing a candidate that
coverage then rejects would (a) stamp `last_used_at` on an entry this job never reads, skewing
retention's LRU against the entries that actually served, and (b) block the whole guard on the
blocking `lock_shared` behind an evictor holding a bundle no load here will touch. Selection now runs
against the scanned artifacts, and each winner is re-proved against the **leased** copy before it is
installed.

## Tests

Rewritten (both previously dropped their guard **before** asserting, so neither asked its own
question — with no scope live the assertion passes whether or not the bug is present):

- `a_narrower_serving_bundle_is_never_adopted` — now asserts **with the scope held**, at the **path**
  layer: the serving bundle still answers for the shared pair and the refused one never does.
- `a_wrong_revision_or_wrong_tier_bundle_is_not_used` — now asserts **with the guard held** that the
  q8 load reads the source library, not the q4 bundle root, and that the path it was handed actually
  holds the q8 weights.

Added: `a_pair_one_sibling_selection_does_not_cover_is_served_to_neither` (with an "alone, q4 IS
served" control), `an_unsupported_bundle_shape_is_reported_and_the_model_stays_on_the_source_tier`,
`the_serving_tier_is_reported_for_every_admitted_model`,
`a_coverage_refusal_is_reported_as_its_own_class_not_as_an_unsupported_shape`,
`entry_covers_is_a_strict_superset_test`, `a_shared_snapshot_stays_served_until_the_last_scope_drops`
(refcount asymmetry), plus rejection assertions in
`valid_local_artifacts_offers_only_loadable_published_entries`.

### Mutation evidence (each guard mutated alone, `touch` after restore)

| Mutation | Result |
|---|---|
| union built over LocalReady plans only (drops the non-local sibling) | **kills 2** — coverage + deliberate-consequence tests |
| `entry_covers` → always `true` | **kills 3** — those 2 plus the core superset test |
| pass-3 fallback never fires | **kills 2** |
| scan rejections dropped again (the original defect) | **kills** the unsupported-shape test |
| conflicting claim replaces the serving entry | **kills** the rewritten narrower-bundle test |
| refuses the caller but poisons the overlay anyway (the exact blocker-2 shape) | **kills** the rewritten narrower-bundle test |

**Honest gap:** the pass-2d re-verification (`leased.contains(entry)`) is a defence against the
scan→lease race and **survives mutation** — no test can reach it without a store hook between the
scan and the lease. It is strictly fail-closed: its worst case is behaving as if it were absent.

## Verification (local, macOS)

- `npm run rust:check` — **exit 0** (derived docs, `cargo fmt --all --check`, clippy, workspace
  tests, build). `rust:test`: **3708 passed, 0 failed** across 33 suites.
- `npm run check` — **412 passed, 0 failed**.
- `cargo test -p sceneworks-worker --lib external_library_runtime` — **24 passed**.
- `cargo test -p sceneworks-core --lib` — **1019 passed, 0 failed, 2 ignored**.

No web/TS files changed, so no vitest or contract-snapshot regeneration is implicated. No new file
calls `model_source_library(`, so `model_artifact_inventory.rs` is unchanged.

---

## Update: base absorb (#2406) + adversarial-review follow-up

`feature/sc-19703-resolved-model-hot-cache` advanced with **#2406 (sc-19709)** after the first base
merge, so it was absorbed here (~518 new lines in `external_library.rs`: `probe_binding`
canonical-path+identity aliasing, `realias_unlocked`, `ModelResolution.library_present`, plus the
API `model_library` module and web files). The merge was clean, but a clean merge proves nothing —
verified the shared seam by hand: `ModelResolution::local_ready` still carries
`requirements: Vec::new()` (the assumption `EntryPlan.requirements` exists for) and
`resolve_model_availability`'s signature and local-candidate matching are unchanged. Then rebuilt
and re-ran everything.

**[major] `revision: None` poison survived mutation.** Replacing the `unserveable_repositories`
insert with a bare `continue` left all 24 guard tests and 10 core tests green, because nothing
constructed a legacy receipt without `snapshotRevision`. Added
`a_requirement_with_no_recorded_revision_makes_its_whole_repository_unserveable` — one bundled pinned
selection plus a legacy carrier on the SAME repository whose receipt records no revision *and whose
manifest declares none either* (a declared pin would otherwise override the receipt's `None` before
the guard ever sees it). Asserts, guard held and at the PATH layer, that neither the pinned nor the
unpinned `refs/main` walk reaches the bundle. **Mutant verified killed**: applied the bare-`continue`
mutation → red on exactly that test with the other 24 green → restored, `touch`ed, re-ran green.

**[minor] Same-process API reader.** Giving API read paths a non-preferring library would touch every
`model_source_library` caller across four API modules, each needing a read-vs-load judgement that
belongs to the availability/catalog stories — so, per the review's own fallback, the safety comment
now states the real bound rather than overclaiming "one job at a time": a concurrent same-process
reader may see a transient **false absent** that self-heals at job end, and **never wrong bytes**.
Pinned by `an_out_of_union_reader_sees_absent_never_the_wrong_bytes`, which seeds a source file the
bundle does not hold with distinguishable content and asserts it reads absent during the scope, is
never answered with other bytes, keeps its authoritative path through `redirect_source_library_path`,
and returns after drop.

**Fixed in passing — a real parallel-test race.**
`resolved_cache_retention_is_opt_in_and_never_creates_the_store` read the cache policy from the
process environment **without** the shared `test_env` lock, so a concurrent `with_local_cache` body
could flip the policy to enabled mid-assertion. It failed once in a full workspace run and passed in
isolation. Note this corrects a premise in the original review item: **`.cargo/config.toml` does NOT
set `RUST_TEST_THREADS`** — this suite really does run in parallel, so the race is live rather than
merely latent. The test now pins the variable off through that lock. (Honest caveat: 6 full-suite
runs of the pre-fix code did not reproduce the failure, so the fix is justified structurally — the
test now cannot observe another test's env — not by a reproduced red.)

### Re-verification on the absorbed base

- `npm run rust:check` — **exit 0** (captured directly). **3731 passed, 0 failed** across 33 suites.
- `cargo test -p sceneworks-worker --lib external_library_runtime` — **25 passed**.
- `cargo test -p sceneworks-core --lib model_artifacts::local_preference` — **11 passed**.
- `cargo test -p sceneworks-worker --lib` — **1530 passed**, run 4x consecutively after the race fix.

### DIFF
--- a/apps/rust-api/src/model_sources.rs
+++ b/apps/rust-api/src/model_sources.rs
@@ -316,9 +316,13 @@ pub(crate) fn local_resolved_artifacts(state: &AppState) -> Vec<ResolvedModelArt
     if !state.settings.resolved_cache.enabled {
         return Vec::new();
     }
+    // Rejections are the worker guard's to report: it is the layer that emits the runtime's
+    // model-source observability, and the catalog would otherwise re-announce the same entry on
+    // every listing.
     sceneworks_core::model_artifacts::resolved_cache::ResolvedCacheStore::valid_local_artifacts(
         &state.settings.data_dir,
     )
+    .artifacts
 }
 
 /// The one availability judgement for a manifest entry on this host: exact selected closure
--- a/apps/rust-api/src/model_sources.rs
+++ b/apps/rust-api/src/model_sources.rs
@@ -307,20 +307,18 @@ pub(crate) async fn resolve_model_manifest_entry_by_repo(
 
 /// App-owned resolved-local artifacts for LocalReady detection. Empty when the resolved cache is
 /// disabled or uninitialized; read-only (never creates a session or refreshes usage).
+///
+/// This is the SAME provider the worker's pre-loader guard uses (sc-19707), so catalog/preflight
+/// and the actual load can never disagree about which entries are valid: an entry that is still
+/// materializing, torn, unverifiable, or in a shape the runtime cannot serve is not a local
+/// candidate on either side of the boundary.
 pub(crate) fn local_resolved_artifacts(state: &AppState) -> Vec<ResolvedModelArtifact> {
     if !state.settings.resolved_cache.enabled {
         return Vec::new();
     }
-    sceneworks_core::model_artifacts::resolved_cache::ResolvedCacheStore::enumerate_existing(
+    sceneworks_core::model_artifacts::resolved_cache::ResolvedCacheStore::valid_local_artifacts(
         &state.settings.data_dir,
     )
-    .map(|entries| {
-        entries
-            .into_iter()
-            .filter_map(|entry| entry.metadata.map(|metadata| metadata.artifact))
-            .collect()
-    })
-    .unwrap_or_default()
 }
 
 /// The one availability judgement for a manifest entry on this host: exact selected closure
--- a/crates/sceneworks-core/src/hf_home.rs
+++ b/crates/sceneworks-core/src/hf_home.rs
@@ -97,7 +97,7 @@ pub fn huggingface_hub_cache_dir(data_dir: &Path) -> PathBuf {
 /// Typed authoritative model source library. Runtime and catalog compatibility helpers delegate
 /// to this value instead of reconstructing a configured Hugging Face root independently.
 pub fn model_source_library(data_dir: &Path) -> ArtifactSourceLibrary {
-    ArtifactSourceLibrary::new(huggingface_hub_cache_dir(data_dir))
+    ArtifactSourceLibrary::new_preferring_local(huggingface_hub_cache_dir(data_dir))
         .expect("the resolved Hugging Face source-library root is nonempty")
 }
 
--- a/crates/sceneworks-core/src/model_artifacts.rs
+++ b/crates/sceneworks-core/src/model_artifacts.rs
@@ -9,6 +9,8 @@
 pub mod artifact_selection;
 #[path = "model_artifacts/external_library.rs"]
 pub mod external_library;
+#[path = "model_artifacts/local_preference.rs"]
+pub mod local_preference;
 #[path = "resolved_cache.rs"]
 pub mod resolved_cache;
 
@@ -622,6 +624,12 @@ pub struct PromotionCandidate {
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct ArtifactSourceLibrary {
     root: PathBuf,
+    /// Whether snapshot discovery through this handle may be served from an active local-tier
+    /// preference scope (sc-19707). True only for the CONFIGURED runtime library built by
+    /// [`crate::hf_home::model_source_library`]; a library constructed directly for maintenance,
+    /// confinement, or materialization deliberately keeps reading the authoritative source, so
+    /// promoting a bundle can never read from a bundle.
+    prefer_local: bool,
 }
 
 /// Behavior-preserving runtime resolver. Source discovery and runtime validation are separate:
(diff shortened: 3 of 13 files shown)


michaeltrefry:
Two correctness follow-ups pushed after the description was written:

1. **A leased snapshot is only answered with when it is still on disk** (`3b6c11c92`). The lease is what prevents eviction mid-load, so this is belt-and-braces: an answer the preference seam gives must be loadable, never a path the caller then discovers is gone. Covered by `a_vanished_bundle_directory_is_never_answered_with`.

2. **A second bundle now ADOPTS a shared snapshot already being served** (`3de7d30cf`). Two models legitimately share one component snapshot — a text encoder is materialized into both bundles — and the first version failed the second installation closed, which would have dropped that model back to the source tier and, with the drive disconnected, failed a job whose models were both complete locally. Each overlay entry now carries the snapshot-relative file set it holds, and a second installation adopts the serving bundle only when that bundle holds every file this artifact needs from the pair; a narrower serving bundle still fails closed. Covered by `a_second_bundle_adopts_the_snapshot_already_being_served`, `a_narrower_serving_bundle_is_never_adopted`, and the worker-level `every_model_a_job_carries_is_served_from_its_own_leased_bundle`.

`npm run rust:check` re-run green locally on the final commit.

PR #2435, opened 2026-08-19
fix(sc-20570): read the mage calibration expectation from the provider crate, and degrade instead of refusing

michaeltrefry:
Fixes [sc-20570](https://app.shortcut.com/trefry/story/20570). Found in production 2026-08-19 — unconditional, every `mage_flow*` request:

```
image generate: generation failed: mage_flow_base loaded provider calibration does not match
ABI 3 / fingerprint mage-flow-generation-peak-v1; refusing request admission against an
unpaired estimator
```

## What was wrong

`crates/sceneworks-worker/src/mlx_fit_gate.rs` pinned the expected mage calibration fingerprint as a SceneWorks literal, `mage-flow-generation-peak-v1`. No loaded provider could ever publish that string, so the pairing check at the bottom of `evaluate_request_with_budget_using_bundle` returned `Err` for every mage request.

**Correction to the story's diagnosis.** The story cited `candle-gen-mage`'s parameterized `mage-flow-cuda-shared-ladder-provider-abi-v2-{id}` scheme. That crate is not the provider on this path — the story flagged the oddity itself ("`mage` is a candle crate … yet the guard lives in `mlx_fit_gate.rs`"), and the resolution is that `mage_flow*` routes to `mlx-gen-mage` on the Mac (`engines.rs`, `adapter_label: "mlx_mage"`). At the pinned rev `6f47e670` that crate declares:

```rust
// crates/media/mlx-gen/mlx-gen-mage/src/model.rs:59
pub const MEMORY_CALIBRATION_FINGERPRINT: &str = "mage-flow-mlx-shared-ladder-2026-08-03-v1";
```

So the literal matched neither provider. The diagnosis (pin drift, hardcoded constant) stands; only the identity of the drifted-from provider changes.

## What changed

The story asks for derive-from-contract, and says that if the special case exists for a reason that still holds it should "at minimum degrade to Legacy like every other engine". Both are implemented — they fix independent defects.

**1. The expectation is read from the provider crate.** `request_total_peak_bytes` and `live_request_budget` quote mage's peak and budget by calling `providers::mage::memory::{generation_peak_gb, production_safe_budget_gb}` directly — that provider-specific estimator is the thing the calibration identity authorizes, so the expectation is now the constant that same crate publishes:

```rust
#[cfg(target_os = "macos")]
const MAGE_CALIBRATION_FINGERPRINT: &str =
    runtime_macos::providers::mage::model::MEMORY_CALIBRATION_FINGERPRINT;
```

`mlx-gen-mage` compiles only on macOS, and off-Mac there is no MLX mage provider to admit at all, so the non-macOS build falls back to a documented mirror that exists purely so the cross-platform request-selector tests keep modelling the production shape. macOS is authority — this box is also the `nax-macos` runner, and the Mac build fails if the mirror ever disagrees with the crate.

**2. mage no longer hard-fails on a calibration-identity mismatch.** It was the only engine that did; the generic identity check twenty lines earlier degrades every other engine to `AdmissionPath::Legacy { StaleIdentity }`. An unpaired identity should cost the request its verified claim, not the request. The check moved above `budget_for_admission` so the demoted route is the one the budget and the existing `mlx_memory_admission_path` event are computed from, and it emits a new `mlx_mage_calibration_unpaired` warning naming the expected and loaded identities.

With both changes, a future provider rename costs mage its evidence for one release instead of bricking six routes.

## Tests

| Test | Covers |
| --- | --- |
| `mage_calibration_expectation_is_read_from_the_provider_crate` (macOS) | the expectation tracks the crate; the off-Mac mirror tracks it too |
| `a_renamed_mage_calibration_identity_degrades_instead_of_refusing` | the degrade itself |
| `mage_resident_path_admits_an_unpaired_provider_on_the_legacy_ladder` | replaces the test that asserted the refusal |
| `mage_pairing_requires_both_the_live_abi_and_the_provider_fingerprint` | the predicate's two conjuncts |

The degrade test is built so it cannot pass vacuously: both arms run the same evaluation over the same evidence lane, and in **both** the contract agrees with its own evidence — that is the post-rename world, where the provider renames and re-captures under the new name, so nothing is internally inconsistent and the generic identity check passes either way. Only the mage seam can separate them. The paired arm asserts it reaches Evidence (`process_limit_bytes.is_some()`) as an explicit precondition before the renamed arm asserts it does not.

**Mutation checks** — each guard flipped alone, test failure confirmed, restored, `touch`ed:

| Mutation | Failing test |
| --- | --- |
| degrade condition -> `false` | `a_renamed_mage_calibration_identity_degrades_instead_of_refusing` (renamed arm reaches Evidence) |
| degrade body -> `return Err` | `mage_resident_path_admits_an_unpaired_provider_on_the_legacy_ladder` |
| ABI conjunct dropped | `mage_pairing_requires_both_the_live_abi_and_the_provider_fingerprint` |
| fingerprint conjunct dropped | same, plus the renamed-identity test |
| constant reverted to the stale literal | `mage_calibration_expectation_is_read_from_the_provider_crate` |
| off-Mac mirror drifted | same |

## Generated files

`docs/generated/memory-matrix.{json,md}` carry the source-tree hash rotation that any edit to `mlx_fit_gate.rs` forces — it is one of the matrix's fingerprinted sources. Provenance only: two hash lines, no cell, coverage, or measurement changed. Same shape as the sibling commits on this branch (`e0b0635e3`, `88ac9a183`). No calibration capture, capability dump, or measurement campaign was run.

## Verification

`npm run rust:check` green (exit 0) on macOS — derived docs, fmt, clippy, tests, build.

Local checks cannot reach the real seam: this lane is cpu-only, and the fix's production path needs a loaded `mlx-gen-mage` generator on Metal. Note the cross-lane oddity the story raised — mage is a candle crate name (`candle-gen-mage` also exists), but the guard fires on the **Mac MLX** path against `mlx-gen-mage`, so verification must be on the Mac, not a CUDA runner.

1. On the Mac, run the API + worker natively with real weights.
2. `mage_flow_base`, q4, 1024×1024, count 1 — must generate end-to-end. This is the story's "Done when".
3. Confirm the logs show `mlx_memory_admission_path` for route `mage_flow_base` and **no** `mlx_mage_calibration_unpaired` warning — the loaded provider's identity should equal the crate constant, so the route is paired.
4. Optional drift rehearsal: no code change needed — the unit tests already cover the renamed path. If you want it live, temporarily edit the checked-out `mlx-gen-mage` constant and confirm the request still generates while emitting `mlx_mage_calibration_unpaired`. Revert afterwards; do not commit an inference change.

There is no mage row in the shipped calibration bundle, so today the paired and unpaired routes both land on the legacy estimate ladder in production — the degrade is currently a no-op in effect, and the observable fix is that the request runs at all.


michaeltrefry:
## Device verification results

Ran the exact "Device verification (coordinator-run)" steps from the PR body natively on this Mac (`nax-macos`), after confirming `nax-worker` was `skipped` (not `in_progress`/queued) on every recent run and no SceneWorks worker process was already running.

**Setup:** `cargo build` (via `npm run rust:check`), then `sceneworks-rust-api` + `sceneworks-rust-worker` (`SCENEWORKS_GPU_ID=mlx`) run natively against a scratch data dir, real weights from `~/.cache/huggingface/hub`. Also ran a second short-lived `SCENEWORKS_GPU_ID=cpu` utility worker to fetch the q4 tier (only bf16 was locally cached going in).

**Job:** `mage_flow_base`, `quantTier: q4`, 1024×1024, count 1, `text_to_image` — `job_9ae229deff8ecc54d500ec0a498d7a63`.

### Result: completed end-to-end on Metal

```
image_pipeline_load_start  jobId=job_9ae2... engine=mage_flow_base
image_pipeline_load_complete jobId=job_9ae2... engine=mage_flow_base
image_inference_start jobId=job_9ae2... backend=mlx
image_inference_complete jobId=job_9ae2... backend=mlx   (30s, 30/30 steps)
```
Job status `completed`, real 1024×1024 PNG asset produced (`adapter: mlx_mage`), visually coherent (verified by eye). No `quant_tier_downgraded` event this run — confirms it actually ran at q4, not a bf16 fallback (the first attempt, before the q4 tier was locally cached, *did* fall back to bf16 and is not part of this evidence).

### Admission-path event

```json
{"message":"selected MLX memory-admission path","event":"mlx_memory_admission_path","route":"mage_flow_base","model":"mage_flow_base","path":"Legacy","fallback_reason":"Some(StaleIdentity)","evidence_revision":"none","width":1024,"height":1024,"count":1,"level":"info"}
```

### `mlx_mage_calibration_unpaired` warning

**Absent.** Grepped the full worker log for the run — zero matches. This is the discriminator that actually exercises this PR's fix: it fires only when `mage_estimator_is_paired` returns `false` (fingerprint/ABI mismatch between `MAGE_CALIBRATION_FINGERPRINT` and the loaded `mlx-gen-mage` provider's published constant). Its absence means the two now agree at runtime on Metal — the bug this PR fixes (permanent refusal from a stale literal) is gone, and the pairing check itself passes.

### Note on the `Legacy`/`StaleIdentity` path (matches the PR's own caveat)

The admission path is `Legacy`, not `Evidence` — this is **expected and documented in the PR body itself**: "There is no mage row in the shipped calibration bundle, so today the paired and unpaired routes both land on the legacy estimate ladder in production." Traced the code: this `StaleIdentity` comes from the **generic** empty-evidence fallback (`mlx_fit_gate.rs` ~line 3203-3216, `admission.evidence.is_empty()`), not from the mage-specific pairing block twenty-some lines later (~3273) — that block never fired its warning, i.e. never re-demoted anything, because the pairing check it guards already passed. Two different code paths reuse the same `StaleIdentity` enum variant for two different causes (no evidence at all vs. an identity mismatch); that's pre-existing and out of this PR's scope. Confirmed non-q4 (bf16, before the tier download) runs the same job type also land on `Legacy`, but with `fallback_reason: Some(NoBinding)` instead — consistent with "no mage row in the calibration bundle" being the actual, tier-independent cause, not the calibration-identity check this PR touches.

### Checklist against the PR's stated verification

| Check | Result |
| --- | --- |
| `mage_flow_base` q4 1024×1024 generates end-to-end | ✅ Pass |
| `mlx_memory_admission_path` event fires for the route | ✅ Pass |
| No `mlx_mage_calibration_unpaired` warning | ✅ Pass |
| Route reaches `Evidence` (not `Legacy`) | Not applicable today — no mage calibration bundle rows exist yet, exactly as the PR body notes; this is unrelated to the fix under test |

No calibration capture, capability dump, or measurement campaign was run. API/worker/utility-worker processes were shut down cleanly (no job in flight) after the run.

[author: claude]

PR #2436, opened 2026-08-19
fix(sc-20572): SubmissionsIgnored is a poisoned queue, not an OOM, and no reset clears it

michaeltrefry:
Fixes sc-20572.

## The defect

Measured in production 2026-08-19. When the macOS GPU driver starts refusing a
process's Metal submissions it completes every later command buffer with
`kIOGPUCommandBufferCallbackErrorSubmissionsIgnored` — *"ignored for causing
prior/excessive GPU errors"*. The model-cache containment seam met that with:

```
MLX generation panicked and was contained (the engine likely ran out of memory;
the cached generator was reset): … kIOGPUCommandBufferCallbackErrorSubmissionsIgnored …
```

Both halves are wrong for this class:

1. **"likely ran out of memory"** — an ignored submission is *never* the first error.
   The OOM guess is right for the first failure (a real `[metal::malloc]` throw) and
   wrong for every ignored submission after it, and it sends diagnosis to the wrong place.
2. **"the cached generator was reset"** — evicting the resident model does not un-poison
   a command queue. ~8 back-to-back jobs (bernini → boogu → krea_2_raw) each hit the
   containment, each reported a successful reset, and each failed identically over
   2.5 minutes, until libc++abi hit an uncaught `std::runtime_error` and `abort()`ed the
   worker. The process respawn is what actually recovered it.

## What changed

New `crates/sceneworks-worker/src/gpu_poison.rs` — detection, scope discrimination and
the composed user-facing text — consulted from the containment seam
(`cache_thread.rs`) and the worker loop (`lib.rs`).

**Attribution.** `SubmissionsIgnored` is classified on its own terms. The error names the
cascade explicitly ("that is the cascade, not the original fault — an ignored submission
is never the first error and is not an out-of-memory condition"), and surfaces this
process's **first** contained failure — the one that actually explains it — instead of
only the ignored submission. Everything that is *not* an ignored submission keeps the
pre-existing containment byte-for-byte, prefix included: the OOM guess is right for a
genuine first failure, which is now the only class that reaches it.

**No reset.** The resident model is deliberately left in place and no backend-cache
release is attempted on this class. Dropping the resident is not a recovery, and both the
drop and the release would submit more Metal work into a channel the driver is refusing.

**Scope discrimination.** The identical error string covers two situations:

- `process` — this worker had already driven the GPU (a completed cache run, or a
  contained fault of its own — exactly the "prior GPU error" being blamed) before the
  poisoning. The refused channel is this process's; the restart clears it; **no reboot**.
- `host_suspected` — this worker had completed no GPU work at all first, so the errors
  being refused over predate it. The message says the restart may not clear it and hands
  over the cross-process discriminator — *is a different MLX process still completing GPU
  work?* — **before** the word reboot is used as a conclusion. The code never asserts a
  reboot on its own; the authoritative discriminator is out-of-process (on 08-19 a
  concurrent `ltx_2_5_tiers_real_weights` run logging healthy forwards is what proved the
  scope was per-process).

**No job-burning loop, then a restart.** The latch is checked at the top of the cache-job
seam, so any queued job is refused *without being handed to the GPU*. On its next turn
the worker loop logs `mlx_command_queue_poisoned_exit`, posts an `Unhealthy` heartbeat
carrying the same reason, and returns `Ok(())` so the supervisor respawns it with a clean
GPU context. `Ok` deliberately: the failing job already posted its own terminal error, and
a non-zero exit would have the desktop supervisor attribute a second, fabricated failure
to it (sc-18182).

**Why `host_suspected` restarts too.** The scope changes what the operator is told, not
whether the worker comes back. The restart is the only recovery available in-process, it
is what demonstrably fixed the measured incident, and a fresh process that fails
identically *is* the "fresh processes fail too" evidence for a host wedge. Parking the
worker alive-but-idle on that branch would strand it on any misclassification, and the
in-process evidence is deliberately weak (see `note_gpu_work_completed`, which only a
completed cache run sets — a poisoning caused by a path outside those seams classifies as
`host_suspected`, the safe direction).

## Scope vs. the story's criteria

| Criterion | Status |
| --- | --- |
| Detect `kIOGPUCommandBufferCallbackErrorSubmissionsIgnored` specifically | done — case-insensitive on the enum tail; nothing else widened into this class |
| Stop attributing it to memory | done — the OOM prefix is now unreachable for this class, asserted by test |
| Stop claiming a reset recovered it | done — no evict, no backend release, and the wording is gone |
| Say the GPU context is poisoned and the worker is restarting | done — job error, refusal text and `Unhealthy` reason all say it, from one shared remedy string |
| Surface the *first* error instead of the cascade | done — recorded per process, included when it differs from the payload |
| Fail the worker fast rather than accepting jobs it cannot serve | done — refusal before `load`/`run`, then exit for restart |
| Two scopes discriminated, reboot never proposed without the cross-process check | done — `process` / `host_suspected`, discriminator gates the word |
| Do not widen | no other Metal error class is folded in; the non-poison path is unchanged |

## Tests

`crates/sceneworks-worker/src/gpu_poison.rs` (11) — an ordinary engine panic is not
classified and does not latch; ignored submissions after completed GPU work are
`process`; in a process that never drove the GPU they are `host_suspected`; an earlier
contained fault re-scopes a later poisoning to `process`; the cascade message carries the
first failure; a poisoning with no earlier fault says so rather than presenting itself as
its own cause; no composed message blames memory or claims a reset; the `process` remedy
rules a reboot out while the `host_suspected` one gates it behind the cross-process
discriminator; the first verdict latches; a latched worker refuses without touching the
GPU; the enum name is recognised in any case (and near-misses are not).

`crates/sceneworks-worker/src/cache_thread.rs` (3) — an ignored-submission panic keeps
the resident model and drops the reset-and-blame-memory prefix; an ordinary panic still
evicts and keeps that prefix; a latched poisoning refuses the next job before any GPU
work. Driven against a local `GpuPoisonState` so the process-global latch is untouched —
latching it in-process would refuse every later job in the test binary, which is exactly
the production behaviour being added.

**Mutation-checked, one guard at a time** (flip → confirm red → restore → `touch`). All
nine killed: detection token never matches (12 red); scope always `process` (3);
prior-fault-does-not-count (1); verdict re-scopes instead of latching (1); first-failure
filter dropped (1); remedies swapped between scopes (2); poisoned branch evicts anyway
(1); ordinary panic stops evicting (1); refusal gate always allows the job through (1).

`npm run rust:check` green.

Serialized on the Mac; **not** run by the implementing agent (CPU-only lane). Note the
story's warning: `cargo test --lib` passing proves nothing here — it passes under a host
wedge too. Only a sustained render command buffer discriminates.

> ⚠️ **Never induce the fault by `pkill`ing an MLX render mid-command-buffer.** That is
> the 2026-08-17 trigger and it wedges the GPU client class until a reboot. Nothing below
> requires it.

### A. Wiring, with no GPU fault at all (zero risk — do this one)

Apply a **temporary, uncommitted** scaffold to `cache_thread.rs`, as the first statement
inside the `catch_unwind` closure of `run_cached_with_access_after_cold_admission` (it must
be *inside* the closure so our arm catches it, not the `run_cache_worker` backstop):

```rust
{
    static N: std::sync::atomic::AtomicUsize = std::sync::atomic::AtomicUsize::new(0);
    if let Ok(trip) = std::env::var("SC20572_FAKE_POISON") {
        let n = N.fetch_add(1, std::sync::atomic::Ordering::SeqCst) + 1;
        if trip.parse::<usize>() == Ok(n) {
            panic!(
                "called `Result::unwrap()` on an `Err` value: Exception(\"[METAL] Command \
                 buffer execution failed: Ignored (for causing prior/excessive GPU errors) \
                 (00000004:kIOGPUCommandBufferCallbackErrorSubmissionsIgnored)\")"
            );
        }
    }
}
```

1. **`host_suspected` branch** — run the app / native MLX worker with
   `SC20572_FAKE_POISON=1`, submit one image job. Expect, in `mlx-worker.log`:
   - the job fails with text containing "the GPU driver is ignoring this worker's Metal
     command submissions", and containing **neither** "ran out of memory" **nor**
     "generator was reset";
   - `mlx_command_queue_poisoned` at error with `scope=host_suspected` and an empty
     `firstFailure`;
   - the message names the cross-process check ("DIFFERENT MLX process") before the word
     reboot;
   - `mlx_command_queue_poisoned_exit`, then the worker exits **0** and the desktop
     supervisor respawns it (`restarting mlx worker in 1s`).
2. **`process` branch** — same build, `SC20572_FAKE_POISON=2`. The first job renders
   normally (proving `note_gpu_work_completed` fires on the real success path), the second
   trips the scaffold. Expect `scope=process`, `firstFailure` empty, and the remedy reading
   "No reboot is needed".
3. **Service resumes** — after either restart, submit a normal image job and confirm it
   completes on the fresh worker. This is the story's Done-when: one accurate error, a
   fast restart, and no cascade of misattributed failures.
4. **Revert the scaffold** (`git checkout crates/sceneworks-worker/src/cache_thread.rs`)
   and rebuild before anything else.

The refusal-before-GPU gate is unit-tested rather than device-observable: the worker runs
one job at a time and exits on its next loop turn, so a second cache job only queues
behind it when a concurrent lane (e.g. a refine-model request) is in flight.

### B. Regression, real weights, no scaffold

5. Run two or three ordinary MLX renders (any warm model) and confirm: unchanged
   behaviour, warm-cache reuse intact, **no** `mlx_command_queue_poisoned*` event, worker
   keeps claiming.
6. Provoke a genuine *first-failure* Metal OOM (e.g. a deliberately low
   `SCENEWORKS_GPU_MEMORY_LIMIT_BYTES`, or a known over-admitting tier). Confirm it still
   reads the old `MLX generation panicked and was contained (the engine likely ran out of
   memory; the cached generator was reset)` prefix, the generator **is** evicted, no poison
   event fires, and the worker keeps serving the next job. A `metal::malloc` throw does not
   poison the queue, so this is safe.

### C. Optional, operator's call — the natural reproduction

7. If the coordinator wants end-to-end evidence on a real driver refusal, the bernini
   over-admission that produced the 08-19 incident is the natural trigger (it is
   process-scoped: the 08-19 occurrence recovered on respawn, with a concurrent MLX
   process rendering fine throughout). Expect exactly one `mlx_command_queue_poisoned`
   with `scope=process`, `firstFailure` carrying the real `[metal::malloc]` cause, then
   the exit and respawn. Skip this if a wedge risk is not acceptable — step A already
   covers every branch of the new code.

## Review round (2026-08-19): fixed

Three issues from adversarial review, resolved in a follow-up commit (core mechanism —
classification reach, latch-refuse-exit chain, byte-identical non-poison path, supervisor
respawn with backoff — was verified correct and untouched):

1. **[major] Self-attributed first failure.** `record_contained_panic` recorded the
   incoming payload as `first_failure` *before* classifying it, so when a
   `SubmissionsIgnored` panic was itself the first contained panic this process saw,
   `first_failure()` — and therefore the `mlx_command_queue_poisoned` event's
   `firstFailure` field, and the refusal/exit text — echoed the poisoning back as its own
   cause instead of staying empty. Fixed at the single recording site in
   `gpu_poison.rs::record_contained_panic` (a `SubmissionsIgnored` payload is never passed
   to `record_first_failure`); downstream sites (`poisoned_queue_message`,
   `refusal_message`) no longer need their own filtering. New tests:
   `gpu_poison::tests::first_contained_panic_being_the_poison_leaves_first_failure_empty`,
   `a_genuine_earlier_failure_is_surfaced_in_first_failure_and_refusal`, and
   `cache_thread::tests::first_failure_emitted_on_the_event_is_empty_only_when_the_poisoning_is_the_first_thing_seen`
   pinning the exact `contained_panic_error` → `firstFailure` read the event uses.
   Mutation-checked (guard flipped → 2 tests red → restored → `touch`).
2. **[minor, same root cause]** `refusal_message`/`poisoned_status` inherited the same
   bug via `first_failure()` — fixed by the same change; covered by the same new
   `gpu_poison` tests (asserting the refusal text as well as the raw getter).
3. **[minor→fixed in place] SAM3 seam never classified poisonings or recorded GPU-driving
   evidence.** `person_segment_sam3.rs`'s dedicated smart-select thread caught panics with
   its own `catch_unwind` that discarded the payload entirely ("sam3 smart-select thread
   panicked"), so a poisoning first surfacing in a SAM3 click was never classified/latched
   (burning clicks with a vague message until a cache-seam job happened to latch it), and
   a successful SAM3 run never called `note_gpu_work_completed`, biasing later scope
   discrimination toward `host_suspected` even when the same process had just driven the
   GPU. Fixed by routing the caught payload through the same `gpu_poison`
   classify/record/latch seam `cache_thread::contained_panic_error` uses
   (`sam3_panic_error`), and marking `drove_gpu` on a successful run
   (`run_sam3_job_with_poison`) — split out from `run_on_sam3_thread` specifically so it's
   unit-testable against a local `GpuPoisonState` without the executor thread or a Metal
   device. Non-poison panics keep SAM3's existing conservative surface unchanged: the raw
   payload is still never echoed to the caller, only the fixed "thread panicked" text. New
   tests: `person_segment_sam3::tests::an_ordinary_sam3_panic_keeps_the_generic_message_and_does_not_latch`,
   `a_poisoning_first_seen_on_the_sam3_thread_is_classified_and_latches`,
   `a_successful_sam3_run_marks_this_process_as_having_driven_the_gpu`. Mutation-checked
   (payload-discard reintroduced → 2 tests red → restored → `touch`).
4. **[minor] `apps/rust-api/src/lib.rs`** — added the requested comment at the
   `tokio::spawn(... run_worker_loop ...)` call in `spawn_inprocess_utility_worker`: the
   in-process CPU utility loops share `gpu_poison::global()` with any MLX cache work in
   this process, so a latch there would `Ok`-exit every loop with no external supervisor
   to restart them. Unreachable today (this API process runs no MLX cache jobs, so nothing
   can ever record a `SubmissionsIgnored` poisoning here) — comment only, no behavior
   change.

**Effect on the device-verification plan below:** steps A.1/A.2's "empty `firstFailure`"
expectations were written against the *intended* behavior and are unaffected in wording —
they were simply not yet guaranteed by the code before this commit; they are now. Add SAM3
smart-select clicks (box/point) to the plan's scope alongside the cache-seam image jobs
used in A/B/C: the same `SC20572_FAKE_POISON`-style scaffold, injected at the top of
`run_sam3_job_with_poison`'s wrapped `job()` call instead of the cache-job closure, would
exercise the SAM3-originated poisoning path the same way A.1/A.2 exercise the cache-seam
one. Not required to close this story — the seam is now covered by the CPU-only unit tests
above — but worth a line in the device plan's scope if the coordinator runs it.

`npm run rust:check` green (fmt, clippy, `cargo test` across the workspace, `cargo build`).
`cargo test -p sceneworks-worker --lib` — 1756 passed, 0 failed, 210 ignored (real-weight/
GPU smokes).

### DIFF
--- a/apps/rust-api/src/lib.rs
+++ b/apps/rust-api/src/lib.rs
@@ -632,6 +632,13 @@ async fn spawn_inprocess_utility_worker(
                 count,
                 "SceneWorks utility worker running in-process (loopback)"
             );
+            // sc-20572 review: these in-process loops share the same `gpu_poison::global()` latch
+            // as the standalone worker's `run_worker_loop`. If it ever latched here, each loop
+            // would read it as an `Ok(())` exit meant for an external process supervisor to
+            // restart — but there is no such supervisor for a tokio task in this process, so the
+            // loop would simply stop claiming jobs forever. Unreachable today: this API process
+            // runs no MLX cache jobs, so nothing here can ever record a `SubmissionsIgnored`
+            // poisoning in the first place. Left as a comment, not code, for that reason.
             tokio::spawn(async move { sceneworks_worker::run_worker_loop(settings).await })
         })
         .collect();
--- a/crates/sceneworks-worker/src/cache_thread.rs
+++ b/crates/sceneworks-worker/src/cache_thread.rs
@@ -32,6 +32,7 @@ use std::time::{Duration, SystemTime};
 
 use tokio::sync::oneshot;
 
+use crate::gpu_poison::{self, ContainedPanic, GpuPoisonState};
 use crate::{WorkerError, WorkerResult};
 
 /// A unit of work handed to a cache worker thread: a boxed closure that runs against the resident
@@ -359,6 +360,63 @@ pub(crate) struct SeamMessages {
     pub worker_dropped: &'static str,
 }
 
+/// Refuse a job outright when the GPU driver is already ignoring this process's submissions
+/// (sc-20572). `None` in the normal case, so this costs one uncontended mutex read per job.
+///
+/// This is the half that stops the burn. Once a poisoning is latched every queued job would hand
+/// the driver another command buffer it has already decided to ignore — which is exactly what
+/// produced ~8 identical failures over 2.5 minutes before libc++abi `abort()`ed the worker. The
+/// refusal happens before `load`/`run` are even entered, so no further Metal work is submitted
+/// while the worker winds down for its restart.
+fn refuse_if_poisoned(poison: &GpuPoisonState) -> Option<WorkerError> {
+    poison.refusal_message().map(WorkerError::Engine)
+}
+
+/// Turn a contained panic into the error this job reports, and decide what happens to the resident
+/// model (sc-20572).
+///
+/// Two distinct outcomes:
+///
+/// * **Ignored submissions** (`kIOGPUCommandBufferCallbackErrorSubmissionsIgnored`) — the resident
+///   model is deliberately LEFT IN PLACE and no backend cache release is attempted. Evicting is not
+///   a recovery for a poisoned command queue (it was measured reporting a successful reset before
+///   each of the identical failures that followed), and both the drop and the release would submit
+///   more Metal work into a channel the driver is refusing. The error text names the cascade, the
+///   scope and this process's first failure instead of guessing at memory.
+/// * **Anything else** — the pre-existing containment, unchanged: evict the resident (post-abort
+///   backend state really is suspect for an ordinary engine panic), release the backend cache, and
+///   report with the lane's own `panic_reset` prefix. The out-of-memory guess in that prefix is
+///   right for a genuine first failure, which is the only class that still reaches it.
+fn contained_panic_error<K, M>(
+    cache: &mut CacheThread<K, M>,
+    msgs: SeamMessages,
+    payload: &str,
+    poison: &GpuPoisonState,
+) -> WorkerError
+where
+    K: Clone + PartialEq,
+{
+    match poison.record_contained_panic(payload) {
+        ContainedPanic::PoisonedQueue { scope, message } => {
+            tracing::error!(
+                event = "mlx_command_queue_poisoned",
+                scope = scope.as_str(),
+                firstFailure = %poison.first_failure().unwrap_or_default(),
+                driverError = %payload,
+                "the GPU driver is ignoring this worker's Metal submissions; the worker cannot \
+                 serve further jobs and is restarting"
+            );
+            WorkerError::Engine(message)
+        }
+        ContainedPanic::Unclassified => {
+            if cache.evict().is_some() {
+                release_backend_cache_after_evict();
+            }
+            WorkerError::Engine(format!("{}: {}", msgs.panic_reset, payload))
+        }
+    }
+}
+
 /// Run `run` against the cached (or freshly-loaded) model for `key` on the dedicated cache `worker`
 /// thread, and await the result. The model lives on the worker thread — `load` builds it there and
 /// `run` executes there (so it may hold a `!Send` reference) — and only the `R` result crosses back.
@@ -414,10 +472,16 @@ where
 {
     let (reply_tx, reply_rx) = oneshot::channel::<WorkerResult<R>>();
     let job: CacheJob<K, M> = Box::new(move |cache: &mut CacheThread<K, M>| {
+        // sc-20572: never submit into a command queue the driver is already ignoring.
+        if let Some(error) = refuse_if_poisoned(gpu_poison::global()) {
+            let _ = reply_tx.send(Err(error));
+            return;
+        }
         // Contain a panic from inside the load/run so it fails THIS job with a clean error instead of
         // unwinding out of the shared cache thread and stopping every subsequent request (sc-6067).
-        // The cached model is evicted on panic — post-abort backend state is suspect, so the next
-        // job reloads fresh.
+        // The cached model is evicted on an ordinary panic — post-abort backend state is suspect, so
+        // the next job reloads fresh. An ignored-submission poisoning is the exception (sc-20572):
+        // see `contained_panic_error`.
         let result = match std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
             cache.with_model_access_after_cold_admission(
                 key,
@@ -427,17 +491,20 @@ where
                 msgs.entry_missing,
             )
         })) {
-            Ok(result) => result,
-            Err(panic) => {
-                if cache.evict().is_some() {
-                    release_backend_cache_after_evict();
+            Ok(result) => {
+                if result.is_ok() {
+                    // sc-20572: proof this process can drive the GPU, which is what later
+                    // distinguishes a poisoning it caused from one that predates it.
+                    gpu_poison::global().note_gpu_work_completed();
                 }
-                Err(WorkerError::Engine(format!(
-                    "{}: {}",
-                    msgs.panic_reset,
-                    panic_message(panic.as_ref())
-                )))
+                result
             }
+            Err(panic) => Err(contained_panic_error(
+                cache,
+                msgs,
+                &panic_message(panic.as_ref()),
+                gpu_poison::global(),
+            )),
         };
         let _ = reply_tx.send(result);
     });
@@ -470,6 +537,11 @@ where
 {
     let (reply_tx, reply_rx) = oneshot::channel::<WorkerResult<R>>();
     let job: CacheJob<K, M> = Box::new(move |cache: &mut CacheThread<K, M>| {
+        // sc-20572: never submit into a command queue the driver is already ignoring.
+        if let Some(error) = refuse_if_poisoned(gpu_poison::global()) {
+            let _ = reply_tx.send(Err(error));
+            return;
+        }
         let result = match std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
             let request_active = || {
                 // A Tokio task can be aborted/dropped without first tripping its engine cancel
@@ -492,17 +564,19 @@ where
                 msgs.entry_missing,
             )
         })) {
-            Ok(result) => result,
-            Err(panic) => {
-                if cache.evict().is_some() {
-                    release_backend_cache_after_evict();
+            Ok(result) => {
+                if result.is_ok() {
+                    // sc-20572: proof this process can drive the GPU (see the sibling seam).
+                    gpu_poison::global().note_gpu_work_completed();
                 }
-                Err(WorkerError::Engine(format!(
-                    "{}: {}",
-                    msgs.panic_reset,
-                    panic_message(panic.as_ref())
-                )))
+                result
             }
+            Err(panic) => Err(contained_panic_error(
+                cache,
+                msgs,
+                &panic_message(panic.as_ref()),
+                gpu_poison::global(),
+            )),
         };
         let _ = reply_tx.send(result);
     });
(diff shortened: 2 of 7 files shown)


michaeltrefry:
## Device verification results

Ran on the coordinator's Mac (128 GiB unified memory), API+worker native (not docker
compose), full API → queue → worker → MLX path. Pre-flight: `gh run list` showed no
`nax-worker` (self-hosted `nax-macos`) job queued or running — only GitHub-hosted
"macOS build, lint" jobs in progress on unrelated infra — and no other sceneworks
worker/API process on the box.

### Doc fix (Part 1)
`docs/observability.md:162` amended to name both `mlx_command_queue_poisoned` emission
seams (`cache_thread.rs` and `person_segment_sam3.rs`'s `sam3_panic_error`), keeping the
"fires once" claim (the latch refuses further jobs and the worker exits before another
segment job is claimed).

### Section A — wiring, zero GPU risk (scaffold applied, exercised, fully reverted)

Applied the PR's temporary `SC20572_FAKE_POISON` scaffold as the first statement inside
`run_cached_with_access_after_cold_admission`'s `catch_unwind` closure. `git status`
confirms it is fully reverted now (only the Part-1 doc commit remains).

**A.1 — `host_suspected`** (`SC20572_FAKE_POISON=1`, first job on a fresh worker):
- Job failed with text containing "the GPU driver is ignoring this worker's Metal
  command submissions" — **no** "ran out of memory", **no** "generator was reset".
- `mlx_command_queue_poisoned` at error: `scope=host_suspected`, **`firstFailure` empty**.
- Message named the cross-process check ("DIFFERENT MLX process") before the word reboot.
- `mlx_command_queue_poisoned_exit` logged, worker exited **0**.

**A.2 — `process`** (`SC20572_FAKE_POISON=2`, same worker): job 1 rendered normally for
real (flux_schnell, proving `note_gpu_work_completed` fires on the real success path),
job 2 tripped the scaffold:
- `mlx_command_queue_poisoned`: `scope=process`, **`firstFailure` empty**, remedy text
  reading "No reboot is needed."
- `mlx_command_queue_poisoned_exit` logged, worker exited **0**.

**Service resumes**: after each restart, a fresh worker (no scaffold env set) claimed and
completed a normal image job.

Scaffold reverted (`git checkout crates/sceneworks-worker/src/cache_thread.rs`),
`git status` clean before Section B, rebuilt (`cargo build -p sceneworks-rust-api -p
sceneworks-rust-worker`).

### Section B — real weights, no scaffold

**B.5 — regression**: 3 back-to-back ordinary flux_schnell renders on one worker
process. All completed (2nd/3rd instant — warm-cache reuse confirmed), **0**
`mlx_command_queue_poisoned*` events across the run, worker stayed `idle`/claiming
throughout (`rust-utility-worker` served all 3).

**B.6 — genuine first-failure Metal OOM**: attempted via three techniques the PR
suggests ("e.g. a deliberately low `SCENEWORKS_GPU_MEMORY_LIMIT_BYTES`, or a known
over-admitting tier"), none of which produced a real Metal OOM on this 128 GiB Mac
under current load:
1. `SCENEWORKS_GPU_MEMORY_LIMIT_BYTES=2147483648` (2 GiB) + flux_schnell (bf16, ~50 GB
   widened peak) — completed successfully. `set_memory_limit`/`set_wired_limit` are
   soft/advisory per the code's own doc comments ("a single oversized allocation can
   still exceed it"); a job well under real physical RAM just proceeds.
2. `bernini_image` at default settings (q8, the known over-admitting tier per
   `mlx_admission_widens_the_estimate_then_compares_it_to_total_ram`, widened estimate
   82.9 GB) at 1024x1024 count=1 — completed after ~13 min; RSS plateaued ~29 GB, well
   under both the estimate and physical RAM.
3. `SCENEWORKS_GPU_MEMORY_LIMIT_BYTES=536870912` (512 MiB) + `bernini_image` — same
   plateau pattern; cooperatively canceled (`POST /jobs/:id/cancel`, not a process
   signal) after ~14 min with no crash, to bound verification time.

None of the three attempts entered a command buffer that had to be interrupted — all
either completed or were canceled cooperatively between steps, so no GPU-wedge risk was
taken. A `count=4`/1280x720 `bernini_image` attempt was also cooperatively canceled
mid-run for the same reason (progress 26.7% at cancel).

**Read on this**: B.6 is inherently hardware/memory-state dependent — the same property
the PR calls out for Section C ("the natural bernini reproduction is nondeterministic").
On this Mac, under current light load, none of the three attempts tipped into a genuine
`[metal::malloc]` throw; all stayed below both physical RAM and the code's own soft
ceilings. This does **not** call the fix into question — the contained-panic → old-prefix
→ evict → no-poison path for a genuine first failure is already covered by the PR's own
unit/mutation-checked tests (`cache_thread.rs`: "an ordinary panic still evicts and keeps
that prefix"). B.6 could not be reproduced device-side in this session; recorded here
rather than silently claimed. Section C (natural bernini reproduction, explicitly
optional/nondeterministic) was skipped per the PR body.

### Clean-up
Worker and API stopped cleanly (SIGTERM, both idle, no job/render in flight at the
time). Scratch data dir used throughout — no writes to `~/SceneWorks/data`, only
`.sceneworks-download-complete.json` markers copied in (non-destructive, per
[[running_sceneworks_api_and_worker_natively_on_mac]]).

PR #2437, opened 2026-08-19
fix(worker): sc-20571 — MLX admission ceiling was total RAM, not live-available memory

michaeltrefry:
## Summary

sc-20571: an uncalibrated MLX `basis=floor` estimate widened by the MLX
estimate margin (~50%, `MLX_ESTIMATE_MARGIN`) was admitted against the
machine's **total** unified memory instead of what was actually **available**
— on 2026-08-19, bernini's widened estimate (122.65 GiB) cleared a 128 GiB
Mac's ceiling even though the desktop app, a browser, and a concurrent
`mlx-gen-ltx` real-weights test were all resident. The model loaded resident,
the Metal command buffer failed ~100s into generation, and the worker's GPU
submission channel was poisoned until it SIGABRTed (`sceneworks-api-2026-08-19-072310.ips`).

## Root cause (verified against source, not assumed)

`crates/sceneworks-worker/src/mlx_fit_gate.rs::live_request_budget` is the
budget constructor that feeds `select_strategy`'s admission ceiling
(`crates/sceneworks-worker/src/memory_strategy.rs`, the file/line the story
cites). It already derived `available = total_bytes − committed_bytes`
(structurally "available", not total) — but `committed_bytes` is
`mlx_rs::memory::get_active_memory()`, which reflects **only this worker
process's own** MLX allocations. On a fresh/first-job worker that number is
near zero, so `available ≈ total`, and the only other subtraction was a flat,
host-blind 2 GiB legacy reserve (`LEGACY_UNIFIED_FALLBACK_RESERVE_GB`). Every
other process resident on the machine — Chrome, the desktop app, LM Studio, a
concurrent `mlx-gen` process — was invisible to the ceiling, exactly matching
the story's "the ceiling is total, not available" framing.

The margin-widening arithmetic itself (`memory_strategy.rs::widened_peak_bytes`,
the `CandidateCurrency::Estimate` arm the story names) is directionally
correct as written: it multiplies the peak **up** before comparing `<=`
available, which can only make a candidate **harder** to admit, never easier
— verified with a dedicated test (see below) and left untouched. What was
missing was live-availability signal on the ceiling side.

## Fix

`live_request_budget` now folds a **live, system-wide "Memory Used"** figure
(`vm_stat`'s App + Wired + Compressed — Activity Monitor's "Memory Used",
the same figure `crate::gpu::query_mlx_utilization` already samples for
telemetry) into `reserved_headroom_bytes`. The shell is thin: it probes, then
delegates the entire policy to two pure, unit-testable functions —

- `live_foreign_reserve_bytes(total_bytes, machine_total_bytes, system_used_bytes, committed_bytes)`
  — the reserve itself: `max(0, used − committed − (machine_total − total_bytes))`.
  Subtracting this worker's own `committed_bytes` keeps the two signals from
  double-charging. Subtracting `(machine_total − total_bytes)` keeps a
  **derated** ceiling from being double-charged either: `mage_flow`'s
  `total_bytes` is an engine-derated safe budget (`limit × SAFE_FRAC`) and an
  `MLX_MEMORY_CAP` total is a smaller-Mac emulation, so both already sit below
  the machine's physical memory; withholding a host-scoped `used` from them
  whole would charge that share twice and produce spurious refusals near the
  boundary. Every subtraction saturates, and an unknown `machine_total`
  (failed `sysctl`) drops the term entirely rather than guessing.
- `compose_live_request_budget(...) -> MemoryBudget` — the fold: probed
  total/machine-total/used/committed/reclaimable plus the branch's flat
  reserve in, `MemoryBudget` out, with `reserved_headroom_bytes = flat +
  live_foreign_reserve_bytes(...)`.

`crate::gpu::system_used_mb_from_vm_stat` is reused (made `pub(crate)`) rather
than re-implementing `vm_stat` parsing, so the async telemetry heartbeat and
the blocking admission probe can never disagree about what "used" means. The
probe runs blocking, matching `probe_total_unified_memory_gib`'s existing
rationale (`live_request_budget` already runs on the blocking generator-cache
thread, so a one-shot subprocess call is free), is **bounded at 2s and killed
on expiry** — matching `gpu.rs::vm_stat_used_mb`'s timeout, because this probe
runs in exactly the wedged-host conditions the fix targets — and fails open
(`None` ⇒ zero added reserve) exactly like every other probe in this file.

### Scope of the fix — which admissions it actually reaches

This is scoped to the **live, per-request** admission path
(`live_request_budget` → `select_strategy`). Within that path it reaches the
**Legacy/estimate routes** — which is precisely the incident class (an
uncalibrated `basis=floor` estimate admitted against total RAM). It does
**not** tighten calibrated admissions: at `mlx_fit_gate.rs:3488-3504` the
Evidence path deliberately passes `reserved_headroom_gb: 0.0` and grades each
candidate against the foreign reserve **captured with its measurement**
(`VerifiedAdmissionCandidate::foreign_reserve_bytes`), so calibrated
admissions remain blind to *live* foreign residency by that earlier design.
That design is untouched here; a code comment at the split now says so
plainly, so the next reader does not mistake this fix for universal coverage.
(An earlier revision of this description claimed it gated "every real
generation request, calibrated and uncalibrated alike" — that overstated it
and has been corrected.)

`video_admission.rs:1097` (`live_video_runtime_state`) also consumes
`live_request_budget`, so the MLX **video** lane inherits the same tightened
ceiling. That is the conservative direction (a video job on a loaded host
becomes harder, never easier, to admit) and is intentional.

I deliberately did **not** touch the separate load-time, cold-load residency
waterfall (`decide_residency_for_spec` / `apply_residency_policy` /
`weights_floor_load_admission`) — that machinery has its own, independently
and deliberately tuned three-layer fallback (peak+headroom check → sequential
offload → weights-only legacy floor override) with existing, deliberately
asserted test coverage (e.g. the mage q4 admit-boundary test) that a naive
"wire in the shared observation" change breaks without live-hardware
verification I cannot do in a CPU-only lane. That risk is called out under
Scope below rather than silently taken.

## Scope note vs. the story's "two compounding defects"

- **Defect 2 (ceiling = total, not available):** fixed — this PR's core change.
- **Defect 1 (margin direction):** verified, NOT a sign-inversion bug —
  `widened_peak_bytes` already grows the compared requirement, the
  conservative direction. Added a dedicated test
  (`bernini_style_floor_estimate_is_refused_once_the_ceiling_reflects_live_foreign_usage`)
  asserting `needed_gb` reflects the widened (not raw) peak, so a future
  change that flipped the direction would go red immediately.
- **"NoBinding routes get no fit check" (adjacent note in memory):** traced —
  `NoBinding` (`MlxCalibrationConfig::Absent`) routes to `AdmissionPath::Legacy`
  in `evaluate_request_with_budget_using_bundle`, which still pushes the
  rung-0 `EstimateFloor` resident-baseline candidate through the SAME
  `select_strategy` ceiling this PR fixes — so this fix closes that gap too
  for the live, per-request path. I did not find a separate code path where
  `NoBinding` skips the fit check entirely; if the note refers to something
  more specific, it's worth a follow-up look, but I could not reproduce a
  distinct gap beyond what this fix already closes.

## Tests (`cargo test -p sceneworks-worker --lib`: all pass; GPU/real-weight tests remain ignored and unaffected)

- `mlx_fit_gate::tests::composed_budget_reserved_headroom_is_the_flat_reserve_plus_live_foreign_usage`
  — **coverage for the fold itself** (the review's major finding): asserts the
  composed `MemoryBudget`'s `reserved_headroom_bytes` equals flat +
  (used − committed), that a failed probe yields exactly the pre-fix shape, and
  that the derated/zero-flat-reserve (`mage_flow`) shape still folds correctly.
- `mlx_fit_gate::tests::a_derated_total_is_not_double_charged_for_the_share_it_already_gave_up`
  — a 96 GiB derated ceiling on a 128 GiB Mac withholds nothing when foreign
  residency is inside the 32 GiB already derated away, and only the excess when
  it is not, while the undecorated ceiling on the same host withholds the full
  amount.
- `mlx_fit_gate::tests::foreign_reserve_bytes_is_live_used_minus_this_workers_own_committed` —
  pure arithmetic: no live signal ⇒ zero reserve (fail-open); no machine total
  ⇒ zero reserve (the derate is unknowable); `used − committed` when positive;
  saturates to zero rather than going negative when `committed ≥ used`
  (stale/racy sample) or when `total_bytes` exceeds the probed machine total.
- `mlx_fit_gate::tests::zeroing_the_live_signal_would_flip_a_positive_reserve_to_zero` —
  mutation-check companion.
- `memory_strategy::tests::bernini_style_floor_estimate_is_refused_once_the_ceiling_reflects_live_foreign_usage` —
  reproduces the production numbers exactly (raw peak `87_563_634_463` bytes,
  `MLX_ESTIMATE_MARGIN`, 128 GiB total) with a `Budget` shaped like the fixed
  `live_request_budget`'s output on a loaded host (2 GiB flat + 20 GiB live
  foreign reserve) and asserts `Selection::Reject`, with `needed_gb` pinned to
  the widened (not raw) peak.
- `memory_strategy::tests::zeroing_the_live_foreign_reserve_reproduces_the_bernini_over_admission` —
  the same candidate against the **pre-fix** ceiling shape (flat 2 GiB only)
  asserts `Selection::Selected` — i.e. characterizes the production bug and
  is the mutation-check for the test above.

**Mutation-checked individually** (mutate one guard → confirm red → restore →
`touch` → re-run green):

1. Dropped the `saturating_add` fold inside `compose_live_request_budget` —
   the exact mutation the reviewer applied to the previously-inlined version,
   which then left the whole suite green. It now fails
   `composed_budget_reserved_headroom_is_the_flat_reserve_plus_live_foreign_usage`
   (2 GiB vs 54 GiB expected).
2. Dropped the derate clamp (`.saturating_sub(already_excluded)`) inside
   `live_foreign_reserve_bytes` — fails
   `a_derated_total_is_not_double_charged_for_the_share_it_already_gave_up`
   (20 GiB double-charged vs 0 expected) and the derated leg of the fold test.
3. Replaced the unknown-`machine_total` fail-open with a `total_bytes` guess —
   fails `foreign_reserve_bytes_is_live_used_minus_this_workers_own_committed`.

## Review follow-ups applied

- **major** — the fold had zero coverage: policy extracted into pure
  `compose_live_request_budget` / `live_foreign_reserve_bytes`, shell kept
  thin, fold unit-tested, and the reviewer's fold-revert mutation re-run to
  confirm it is now caught.
- **minor** — the blocking `vm_stat` probe is now bounded (2s timeout, kill +
  reap on expiry, `None` on failure), matching `gpu.rs::vm_stat_used_mb`.
- **minor** — the fold no longer double-charges a derated `total_bytes`
  (`mage_flow` safe budget, `MLX_MEMORY_CAP` emulation); test added.
- **minor** — the new helper is renamed `live_foreign_reserve_bytes` (and its
  local binding) so it no longer collides with the differently-scoped
  `VerifiedAdmissionCandidate::foreign_reserve_bytes` **field** (capture-host
  reserve, not live-host) nor shadows itself at the call site.
- **minor** — the "gates every real generation request, calibrated and
  uncalibrated alike" claim is corrected here and in a code comment at the
  Evidence/Legacy split; the `video_admission.rs:1097` inheritance is noted
  above.

## Generated artifact

`docs/generated/memory-matrix.json`/`.md` regenerated
(`npm run generate:memory-matrix`) after editing two of its fingerprinted
sources (`mlx_fit_gate.rs`, `memory_strategy.rs`) — this is the static,
offline, source-hash-only regeneration `check:memory-matrix` requires on
every PR touching a tracked file (confirmed on the base branch: the check
passes clean pre-edit, fails only after these edits). It is **not** a
calibration/measurement campaign — diffed the regenerated JSON: only the
`sceneWorksRevision` source-tree hash and the two per-file hashes rotated,
zero admission-decision cells changed. No inference pin bump, no calibration
capture, no disabled check re-enabled.

This PR's fix cannot be verified end-to-end without a real macOS/MLX worker
and a deliberately loaded host — that verification belongs to the
coordinator's serialized device-verification phase, not this CPU-only lane.
Suggested procedure:

1. On the `nax-macos` dev Mac (or an equivalent Apple-Silicon host), build
   this branch and start the SceneWorks API + worker natively.
2. Deliberately load the host: open the desktop app, a browser with several
   tabs, and start a second, independent MLX-consuming process (e.g. a
   `mlx-gen-ltx` real-weights run) so combined foreign usage is tens of GiB —
   reproducing the incident's concurrency shape.
3. Trigger a `bernini` (or another uncalibrated/`basis=floor`) generation
   whose raw weight peak is large enough that `raw_peak × (1 + MLX_ESTIMATE_MARGIN)`
   plus the new live foreign reserve would exceed the host's free memory.
4. Expected: the request is refused with an actionable message (not admitted
   resident), and `~/Library/Logs/SceneWorks/mlx-worker.log` shows the
   `widened_peak_bytes` estimate compared against a ceiling that has visibly
   shrunk from the pre-fix `total − 2 GiB` shape — grep for
   `widened_peak_bytes` and confirm the admission line's implied ceiling
   moves with host load, not just with `hw.memsize`.
5. Confirm the machine does **not** repeat the original crash (poisoned GPU
   queue / SIGABRT) under this same loaded-host condition where the pre-fix
   code would have admitted the job.
6. As a negative control, confirm the SAME request is admitted normally when
   the host is otherwise idle (no regression on the common case), and that a
   `mage_flow` request near its derated safe-budget boundary is still admitted
   on a lightly loaded host (the derate clamp, verified in unit tests, means
   the safe-budget share is not charged twice).

## Scope

Single-story fix. Did not touch: margin *values* (constants unchanged, per
epic policy — floors-as-records is the accepted state), the calibrated
Evidence-path headroom design (wording/comment only, see above), the load-time
cold-load residency waterfall (see above), mage_flow's fingerprint mismatch
(sc-20570, separate story), sensenova's geometry gate (sc-20569, separate
story), or the SubmissionsIgnored panic-containment messaging (sc-20572,
separate story).


michaeltrefry:
## Device verification results (coordinator-run)

Ran natively on this Mac (128 GiB, `nax-macos`) against `origin/feature/sc-19048-memory-admission-convergence` @ `aff30bca067c50320ec5c0fb12f0078779e20b8a` (PR #2437's fix, `8a2b8ba76`, is an ancestor). API + worker built from source and run against a scratch data dir (real HF cache, real weights); no changes committed or pushed.

### Route correction vs. the suggested procedure

The suggested procedure named `bernini` as the repro model. On inspection, **neither `bernini_image` (still) nor `bernini`'s video handler (`video_jobs/bernini.rs` → `video_jobs/wan.rs::generate_video`) actually calls the fixed `evaluate_request`/`live_request_budget` path** — both only reach the separate, *deliberately untouched* load-time cold-load residency waterfall (`decide_residency_for_spec`, route `generic_mlx_cold_load`, `MlxMemoryBudget{total_gb}` only, no live-foreign-reserve fold). I confirmed this by tracing call sites (`grep` for `evaluate_request(` across `image_jobs/*.rs` and `video_jobs/*.rs`) and by direct observation: a bf16 `bernini_image` job (raw peak 87,563,634,463 B — the exact incident number) admitted identically under both an idle host and a host loaded with ~35 GiB of synthetic foreign memory, always via `route:"generic_mlx_cold_load"`, confirming that route is unaffected by this fix (as intended — see PR body's "Scope of the fix").

The **actual fixed path** (`AdmissionPath::Legacy`, `fallback_reason:"NoBinding"`, `event:"memory_strategy_request_selected"`) is reached by the generic per-request evaluate_request call (`image_jobs/base.rs:8047` and siblings) that most uncalibrated MLX image models use — verified live with `sana_sprint_1600m` (also NoBinding/uncalibrated, same `basis=floor` estimate class as the incident). Both the idle-host admit and loaded-host refusal below exercise this exact code.

### Idle-host negative control — ADMITTED

`sana_sprint_1600m`, bf16 (`advanced.mlxQuantize:0`), 1024×1024×4, on an otherwise-idle host:
```
event=memory_strategy_request_selected route=sana_sprint_1600m path=Legacy fallback_reason=Some(NoBinding)
predicted_peak_bytes=16758625921 needed_gb=23.475106364116073 available_gb=106.4921875 (first probe) / 74.74 (warm re-probe)
strategy=Resident optimization_authority=Resident
```
Job completed normally (`Generated 4 image(s)`), no crash.

### Loaded-host refusal — REFUSED with actionable message

Host load was generated **synthetically** (per an explicit correction from the coordinator mid-run: do not open Michael's desktop app or browser as load generators — I had briefly opened both, closed/quit exactly what I'd opened within ~2 minutes, and switched to synthetic ballast for the remainder). Ballast = a Python process touching-and-holding anonymous memory in 1 GiB chunks (every 4 KiB page written, so it's genuinely resident, not just reserved), ramped in three increments (20 → 35 → 50 GiB total) while watching `memory_pressure -Q` (stayed at 95–96% free throughout — never approached red) and the worker's own `vm_stat`-derived `memoryUsedMb` telemetry.

Same `sana_sprint_1600m` bf16 1024×1024×4 request, resubmitted as ballast ramped:

| ballast (GiB) | available_gb | needed_gb | result |
|---|---|---|---|
| 0 (idle) | 106.49 | 23.48 | admitted |
| 20 | 50.93 | 23.48 | admitted |
| 35 | 35.40 | 23.48 | admitted |
| 50 | 21.75 (final probe) | 23.48 | **refused** |

Final failure:
```
event=utility_job_failed jobId=job_6fdb7c03b8dfab2cdd46542a49ed8185
error="sana_sprint_1600m request 1024x1024 count 4 needs 23.48 GiB but only 21.75 GiB is safely available"
```
Clean, actionable refusal — no crash, no GPU poisoning from this request, worker remained `idle`/responsive immediately after.

I also captured the identical-numbers idle-vs-loaded `bernini` `generic_mlx_cold_load` admission (raw_peak_bytes=87563634463, widened_peak_bytes=131702133700, `estimate_margin=0.504…`) on both an idle host and under ~35 GiB of load — admitted both times, confirming that route is *not* gated by this fix, consistent with the PR body's explicit scope note.

### Incident (`kIOGPUCommandBufferCallbackErrorSubmissionsIgnored`), for the record

Mid-run, an unrelated `bernini` video job hit `kIOGPUCommandBufferCallbackErrorSubmissionsIgnored` on an idle host (not a memory-admission failure — this was the untouched cold-load route, well within budget). I did **not** signal-kill anything mid-render. Per the scope-discrimination runbook: a fresh discriminator job on the *same* worker process also failed with the identical error (process-scoped poisoning confirmed); I restarted only the worker process (not the host), and a subsequent discriminator job completed normally. No reboot was needed — this was a process-scoped Metal queue poisoning, not a host-wide wedge, and is unrelated to the sc-20571 fix under test.

### Cleanup

All ballast processes killed (`SIGTERM`, confirmed exited) once memory recovered to baseline (~19–29 GiB used, matching pre-test ambient). Test worker/API processes shut down cleanly (exit 0) with no render in flight. Worktree used for verification: detached checkout of `origin/feature/sc-19048-memory-admission-convergence`, nothing committed or pushed (`git status` clean).

[author: claude]

PR #2460, opened 2026-08-20
fix(sc-18650): close six pre-merge review findings across the Ref2VA surface

michaeltrefry:
Six findings from the 19-agent adversarial pre-merge review of epic sc-17137, in one PR.
Target: `feature/sc-17137-minimax-h3`.

## F1 (BLOCKER) — standalone reference audio is never resampled, so `referenceAudioAssetIds` had zero reachable success paths

`resolve_reference_audio_conditioning` read the asset's WAV header and passed its declared rate straight into `Conditioning::ReferenceAudio`. At the pin, `Ref2VaReferences::check_audio` hard-errors unless `sample_rate == 32_000` ("this engine has no resampler — resample it before sending it"), and **no shipped SceneWorks producer emits that rate**: every `audio.sampleRates` in `builtin.models.jsonc` is 22050 / 24000 / 48000. A caller could name a reference, pass preflight, and only learn at the engine boundary that nothing in their library was admissible.

The resolver now normalizes through the **same shared ffmpeg pass the reference-CLIP path already used** — `-map 0:a:0 -vn -ar 32000 -c:a pcm_s16le`, driven by `media_jobs::run_ffmpeg` with `FfmpegContext`, so it inherits the heartbeat + cooperative-cancel lifecycle every other media decode in the crate has. It runs unconditionally rather than behind an "is it already 32 kHz?" probe: one code path means the conversion is exercised by every input, and it lets a reference be any container ffmpeg reads (mp3 / m4a / flac / non-PCM WAV) instead of only the canonical PCM-16 RIFF `read_wav_pcm16` decodes. Channel layout is deliberately left alone — the engine accepts any positive count, and both directions of "fixing" it substitute a different reference than the one supplied.

The function is now async and takes `api`/`job`; both call sites (`run_video_generate_job`'s early proof and `resolve_minimax_h3_conditioning`) await it.

**Moved to `crates/sceneworks-worker/src/video_jobs/reference_audio.rs`.** The added code pushed `video_jobs/mod.rs` to 2583 lines against the 2500-line ceiling `video_jobs_remains_split_into_real_engine_modules` enforces. Rather than raise the ceiling, the resolver became its own module — it is cross-model and cross-platform, not an engine family, so it is not added to that test's `MODULES` list. Its scratch dir is registered in the `temp_fixture_guard` ALLOWED list beside the reference-clip one.

**The fixture was a false green.** It hand-wrote `sample_rate: 32_000` and asserted it back — a constant echoed through a resolver that did nothing, equally satisfied by resampling and by a no-op. It now writes **24 kHz** (a real producer rate, 4:3 to the engine's 32 kHz) and asserts both the rate AND the resampled length, so a passthrough reds on the rate and a relabel-the-header cheat reds on the length. The path-guard negative arms run on every host; the transcode half soft-skips behind the existing `ffmpeg_reachable()` convention (hard-required in CI via `SCENEWORKS_REQUIRE_FFMPEG`).

## F2 (MAJOR) — the reference-audio picker's import dropzone could never succeed

**Branch taken: the PREFERRED one — audio uploads are now first class.**

Evidence the asset model already carried audio end-to-end, so this is a contained change and not a new asset kind:
- `build_audio_sidecar_parts` (`project_store.rs`) already writes `type: "audio"` with `mimeType` / `duration` / `sampleRate` / `channels`; `build_generated_asset_sidecar` already dispatches on `"audio"`.
- `origin: "audio_studio"` is already in `LIBRARY_ORIGINS`; the Library grid already has `AudioAssetTile` / `AudioAssetStage`, and `assetMedia.jsx` already has an `<audio>` branch.
- Asset listing has **no** type filter on the production path, and `project_file` has **no** serve allow-list.
- `POST /api/v1/projects/:id/assets` has **no** content-type / extension / sniff gate — the 400 came entirely from `sceneworks-core`.

What changed:
- `import_asset` and `media_type_for_mime` accept `audio/*` and type it `"audio"`.
- New `normalize_audio_upload` → `media_convert::transcode_to_wav_pcm16` converts **every** accepted audio upload to PCM-16 RIFF/WAVE. This is not a preference: `read_wav_pcm16` is the product's only audio decoder and refuses anything else, so storing an `.mp3`/`.flac` verbatim would produce an asset that looks usable everywhere and fails at render time. `sceneworks-core` already shells out to ffmpeg (`normalize_image_upload` → `transcode_to_png`), so this adds no new dependency direction. Unlike images there is no `sips` fallback, and a host with no ffmpeg says so.
- The sidecar's `file` block carries `duration`/`sampleRate`/`channels` for audio only (measured off the stored WAV), matching the Audio Studio shape. No image/video sidecar grows a key.
- `SAFE_UPLOAD_EXTENSIONS` gains `wav` — the only extension an audio asset is ever stored under, since the source extension never reaches the filesystem. Its pinning test now admits `audio/` alongside `image/`/`video/`, still excluding `image/svg+xml`.
- `LibraryScreen` accepts `audio/*` and has an Audio type filter.
- The AssetPicker audio tests now stub `importAsset` with the row the store actually writes (`type: "audio"`, `origin: "upload"`, `file.mimeType: "audio/wav"`, real facts) instead of an invented three-key object. The `.flac` case is the load-bearing one: the response comes back `audio/wav`, not the mime the browser reported, so a picker that compared the two would drop the import.
- New end-to-end Rust test `import_asset_admits_audio_and_normalizes_it_to_pcm16_wav` feeds a 32-bit **float** WAV (a perfectly ordinary `.wav` that `read_wav_pcm16` refuses) and asserts the stored bytes come back PCM-16, plus type/extension/serve-mime/facts, plus that a non-media upload is still refused.

## F3 (MAJOR) — matrix mapped `referenceVideo` onto the `videoClip` requirement

`matrix.rs:1763` sent the `referenceVideo` conditioning axis to the `videoClip` requirement. `minimax_h3_ref`'s only mode is `reference_to_video`, whose requirement group is `["multiReference","reference"]` — it has no `videoClip` mapping at all — so the probe set became six modes the model never routes, `native_video_route_descriptors` came back empty every time, and the cell fell to false **with no error** (the "no production mode semantic" guard never fired because `modes` was non-empty). The committed matrix therefore said unsupported while the manifest declares `maxSourceClipAssets: 3` and the worker decodes those clips into `Conditioning::ReferenceVideo`.

Fixed to `"referenceVideo" => "multiReference"`. Verified undisturbed:
- **bernini** declares no `referenceVideo` axis at all — the only engine descriptor that does is `minimax_h3`, so only `minimax_h3` and `minimax_h3_ref` even have this cell. `audit_regressions_are_derived_from_production_truth`'s bernini pins are green.
- `singular_shadowed_by_bundle` keys on `shape == "reference"` (the raw axis, never `requirement_shape`), so it short-circuits before reaching anything this touched.
- Base `minimax_h3` stays `false/false` — it routes none of the multiReference modes.

`config/backend-capabilities/matrix.json` regenerated with its own generator (`cargo run -p sceneworks-core --bin dump-backend-capability-matrix`). Diff is exactly one cell plus the summary: `mlxOnlyCellCount` 80 → 81.

### 🔴 The ninth cell has NO approval record, and that is deliberate

`models/minimax_h3_ref/conditioningShape/referenceVideo` is now a ninth residual MLX-only cell. **I received no message saying Michael approved it, so the register is untouched** — I will not author an approval record in his name.

Consequence: `approved_exception_register_covers_every_residual_mlx_only_cell_exactly_once` is **RED**, and it is the only red in `-p sceneworks-core --lib` (1163 pass / 1 fail). The delta is exactly that one path. When approval arrives, the fix is two edits:
1. add `"models/minimax_h3_ref/conditioningShape/referenceVideo"` to `epic-17137-minimax-h3-ref-conditioning-sequencing`'s `cells` in `config/backend-capabilities/exceptions.json`;
2. bump `matrix.rs:4853` from `2usize` to `3usize`.

(`summary.exceptionCount` stays 19 — it counts records, not cells.)

## F4 (LOW) — a deterministic main regression, taken twice

Root `Cargo.toml` carried `0.8.1` while `package.json`, `apps/web/package.json` and `tauri.conf.json` carried `0.8.7`. Main's blob is byte-identical to the merge base, so the sync merge takes 0.8.1 with no conflict. Restored to `0.8.7`; `cargo update --workspace --offline` refreshed the lock.

**Tightened the existing test's reach** (not a new gate): `producer_version_is_strict_semver_and_matches_the_workspace` compared `CARGO_PKG_VERSION` to root `Cargo.toml` — a value against itself, which is why it stayed green through both regressions. It now also compares the workspace version against `package.json`, `apps/web/package.json`, `apps/desktop/package.json` and `apps/desktop/tauri.conf.json`, with a failure message naming `scripts/sync-version.mjs`.

## F5 (MINOR, latent) — rung 4 deducted the transformer twice

`estimate_floor_weights_bytes`'s staged branch computed `heavy = max(heavy_load_exact - evicted, transformer)` and *then* subtracted `transformer_bytes` again for `BoundedTransformerResidency`, yielding `max(0, decoder - evicted)` where the answer is `decoder`. At bf16 the `.max(transformer_bytes)` clamp binds and the result is exactly **0** — MiniMax-H3's whole 11.02 GB video+audio VAE pair gone from the floor, in the OOM direction. The doc at `:2250-2252` asserted the opposite invariant verbatim.

The two reductions are now exclusive: with rung 4 engaged, `heavy = heavy_load_exact - transformer_bytes` and the eviction reduction is skipped entirely (the sub-stack is inside the window rung 4 just removed). No live admission moves — MiniMax-H3 declares empty `additional_prerequisites` and is the only `PrecomputedThenEvicted` provider — but 13 shipped MLX providers already declare the rung-4 → rung-1 edge, so this is one declaration away from mattering.

**The test that should have caught it was masked.** `rung_four_deducts_the_transformer_once_not_twice` ran on `PACKED_TEXT_ENCODER_BYTES`, which is *taller* than `VAE_BYTES`, so the staged branch's `conditioning.max(heavy)` threw `heavy` away and both arms collapsed onto the conditioner; even its expected value's `.max(VAE_BYTES)` was inert. Re-graded with a conditioner **below** the decoder, and the composition is now read off `contract.engaged_composition(BoundedTransformerResidency)` on a fixture that declares the rung-4 → rung-1 edge (via a new `rung4_edge_contract` helper) rather than a hand-written slice no selector produces. The fixture's two enabling properties are asserted off the contract's own facts so a later constant edit cannot silently re-mask it.

## F6 (MINOR) — stale inference-rev labels and a dead line citation

All of these are new on this branch (zero occurrences at the merge base).

`config/rung4-applicability-survey.json` cited `0e4adc6f` as "the pinned rev" at lines 34 / 41 / 57; `scripts/generate-memory-matrix.mjs` cited `014134e3` at 1924 / 2622 / 2670 / 2705 (plus a rangeless citation at 1741), and `scripts/rung4-contract-prerequisites.mjs:673`. The pin is `f17c82544`.

**Reasoning about the `assert.notEqual(revision, pin)` guard.** That assertion (`generate-memory-matrix.test.mjs:3113`) says the two `contractRevision` anchors must DIFFER from the pin — the field earns its place only by dating something the matrix's own `generatedFrom.inferenceRevision` does not. Blindly re-anchoring them to the pin would turn it red, and I did not touch them. But the *justification* written around them had gone false, and that is the actual defect: the prose said "at the pinned rev neither H3 crate is present". **Measured at `f17c82544`: both `mlx-gen-minimax-h3` and `candle-gen-minimax-h3` ARE present, and `git merge-base --is-ancestor` confirms both anchors (`e09f46aaf` for MLX, `79f02e6d0` for Candle) are ancestors of the pin.** So the anchor's meaning has moved from "crate absent from the pin" to "**last surveyed here, and not re-surveyed since the pin moved past it**" — a materially different and stronger obligation, and the reason re-stamping without re-surveying is forbidden rather than merely untidy. That is now what the prose says, at all three sites plus the generator's mirror of the same claim.

Also corrected: the gen-core citations. `BOUNDED_TRANSFORMER_RESIDENCY_REQUIRES` is at `memory_strategy.rs:310-313` with the `LoadShape::DeferredMaterialization` identifier on line 312 (was cited 290-293/292; at the live pin 288-295 is an unrelated doc block). `MemoryProviderContract::engages` is at 1346-1362 (was 1028-1044). `validate_selection`'s two `EngagedInSameRequest` accepting arms are at 1862-1873 (was 1468-1481). The 21-of-40 rung-1-edge count is re-attributed to `f17c82544` — it was always read off `config/rung4-contract-prerequisites.json`, which carries that revision and is pin-drift-guarded, so only the label was stale. The notes-ban assertions (`/LoadShape::DeferredMaterialization/`, `/SC-15998/`, no `/rung 4 requires rung 1/i`) and the two `whyOutOfMatrix` sha regexes are all preserved. **All 115 `generate-memory-matrix.test.mjs` tests pass**, including the notEqual guard.

`rung4-contract-prerequisites.mjs:1123` was left alone — it narrates a past pin bump (`014134e30 → 0e4adc6f0`) and is correct as history.

## Verification

| Suite | Result |
|---|---|
| `cargo fmt --all --check` | clean |
| `cargo clippy -p sceneworks-core -p sceneworks-worker --lib --tests --all-targets -- -D warnings` | clean |
| `cargo clippy -p sceneworks-rust-api --lib --bins --tests -- -D warnings` | clean |
| `cargo test -p sceneworks-worker --lib` | **1797 pass / 0 fail** / 210 ignored |
| `cargo test -p sceneworks-core --lib` | 1163 pass / **1 known-red (F3 ninth cell)** |
| `cargo test -p sceneworks-core --tests` (integration targets) | all green |
| `cargo test -p sceneworks-rust-api` | 692 pass / 0 fail |
| `apps/web` vitest (`npm test`) | 265 files, **4026 pass** |
| `npm run check` | 675 pass / 0 fail |
| `node --test scripts/generate-memory-matrix.test.mjs` | 115 pass / 0 fail |
| `node --test scripts/rung4-contract-prerequisites.test.mjs` | 10 pass / 0 fail |

### Mutation results (each guard mutated alone, `touch` between restores)

**F1** — remove `-ar` → rate assertion reds (24000 vs 32000). Relabel the header instead of resampling → length assertion reds (6000, expected ~8000). Add `-ac 2` → channel assertion reds.
**F2** — close the `audio/` gate → import fails. `media_type_for_mime` → `"video"` → type assertion reds. `pcm_s16le` → `pcm_f32le` → stored-format assertion reds (3 vs 1). Skip normalization entirely → same assertion reds. Drop the clip-facts block → sampleRate assertion reds (Null vs 24000).
**F3** — revert the mapping to `videoClip` → `checked_in_matrix_matches_all_authoritative_sources` reds ("capability semantics drifted").
**F4** — set `Cargo.toml` back to 0.8.1 → the widened assertion reds with the sync-version guidance.
**F5** — restore the OLD arithmetic → the floor-equality assertion reds (8339364109 vs 11020988228, exactly the decoder). Make rung 4 remove nothing → the `VAE_BYTES` assertion reds. Remove the declared edge → composition assertion reds (`[Resident, BoundedTransformerResidency]`). Put the conditioner back above the decoder → the anti-mask guard reds.
**F2 (web)** — narrow the audio extension regex → the `.flac` test reds. Break `assetCanRenderAsAudio` → three import tests red. Drop the post-flight kind check → the wrong-kind test reds.

## 🔴 Two things the coordinator needs to action

1. **`approved_exception_register_covers_every_residual_mlx_only_cell_exactly_once` is red** pending Michael's decision on the ninth cell (see F3). No approval record was authored.
2. **`docs/generated/memory-matrix.json` is now stale** and I did not touch it, per instruction. It was green at branch HEAD; three of my files are hashed matrix sources (`mlx_fit_gate.rs`, `config/rung4-applicability-survey.json`, `Cargo.toml`). I measured the regeneration: the diff is **exactly four provenance sha lines plus the .md's mirrored revision line — zero capability cells move.** `npm run check:memory-matrix` (and `check:rust-derived-docs`, hence the pre-push hook and `rust:check`) will be red until the sequenced regeneration runs.

## Notes

- The inference pin is untouched. No check scripts were wired or re-enabled. No new gates — F4 widens an existing test's reach and F5 re-grades an existing one.
- `reference_audio.rs` was deliberately NOT added to the matrix source-capture lists: those are explicitly provenance rather than semantics (`source_capture_digest_values_are_provenance_not_semantics`), they already omit `video_jobs/minimax_h3.rs`, and the resolver feeds no capability cell.


michaeltrefry:
## CI outcome — 3 red checks, both root causes are the documented, coordinator-owned known-reds

Nothing here is an unexpected failure. Full state:

**Green (11):** `build-windows` (10m14s) · `candle` · `check-linux` · `check-macos` · `parity-digests` · `parity-docker` (8m47s) · `parity-scaffold` · `web` · 4× `changes / decide`
**Skipped (2):** `nax-worker` · `package-windows`

**Red (3), reducible to two causes:**

### 1. `macOS build, lint and workspace tests (hosted)` — the F3 ninth cell
```
test jobs_store::routing::matrix::tests::approved_exception_register_covers_every_residual_mlx_only_cell_exactly_once ... FAILED
test result: FAILED. 1163 passed; 1 failed; 2 ignored
```
**Exactly one test**, byte-identical to the local run, and the set delta is exactly `models/minimax_h3_ref/conditioningShape/referenceVideo`. This is the ninth residual MLX-only cell surfaced by the F3 mapping fix. **Michael has not answered on it, so no approval record was authored.** Two edits clear it once he does:
1. add that path to `epic-17137-minimax-h3-ref-conditioning-sequencing`'s `cells` in `config/backend-capabilities/exceptions.json`;
2. bump `matrix.rs:4853` `2usize` → `3usize`.

### 2. `parity-rust` (and `parity`, its 3s roll-up) — the memory-matrix staleness
```
FAILED tests/test_memory_matrix.py::test_generated_memory_matrix_is_current_and_schema_valid
Error: generated memory matrix is stale; run npm run generate:memory-matrix
1 failed, 117 passed, 19 deselected
```
`docs/generated/memory-matrix.json` was deliberately **not** regenerated — coordinator-sequenced behind the pending second pin bump. Three files in this PR are hashed matrix sources (`mlx_fit_gate.rs` from F5, `config/rung4-applicability-survey.json` from F6, `Cargo.toml` from F4), so the rotation is unavoidable given those fixes.

I measured what the sequenced regeneration will produce: **exactly four provenance sha lines (`sceneWorksRevision`, `mlxFitGate`, `rung4Survey`, `cargo`) plus the `.md`'s mirrored revision line. Zero capability cells move.**

Note `check-macos` / `check-linux` are green — the memory-matrix check lives in the `parity-rust` pytest lane, not in `rust:check`.

### One coverage caveat worth flagging

`cargo test` stopped after `sceneworks-core` failed, so the **`sceneworks-worker` lib tests did not run on the hosted macOS lane** this time — which means F1's `resolve_reference_audio_conditioning_resolves_project_relative_asset_paths` has no CI evidence on this run. It is masked by the F3 red, not by anything wrong with it. Locally it runs the real ffmpeg transcode (not the soft skip) and all 1797 worker lib tests pass. Clearing the ninth cell will let that lane exercise it. As independent evidence that ffmpeg is genuinely present and working on that runner, the sibling F2 test **did** run there and passed for real:
```
test project_store::tests::import_asset_admits_audio_and_normalizes_it_to_pcm16_wav ... ok
```
with no "no ffmpeg reachable on this host" skip line anywhere in the log.

michaeltrefry:
## Correction: the pinned-count line number for the ninth cell

The PR body and my previous comment both say `matrix.rs:4853`. **That is wrong and would edit the wrong record.** The correct coordinates in the committed file (`655939b11`):

```
4871   "epic-17137-minimax-h3-ref-conditioning-sequencing",
4872   ("epic-17137", "conditioning", 2usize),      <-- bump THIS one to 3usize
```

Line 4853 is a bare `),`, and the nearby `2usize` at **4856** belongs to `epic-17137-minimax-h3-conditioning-sequencing` — the BASE `minimax_h3` record, not the `_ref` twin. Bumping 4856 would leave the ninth cell uncovered while over-counting the base model's, so the register test would stay red for a different and more confusing reason.

(The brief's original `matrix.rs:4852` was correct *before* this PR; the F3 fix adds ~19 lines of comment at `:1763`, which shifted the entry down to 4872.)

So, when Michael approves the ninth cell, the two edits are:
1. `config/backend-capabilities/exceptions.json` — add `"models/minimax_h3_ref/conditioningShape/referenceVideo"` to `epic-17137-minimax-h3-ref-conditioning-sequencing`'s `cells`;
2. `crates/sceneworks-core/src/jobs_store/routing/matrix.rs:4872` — `2usize` → `3usize`.

michaeltrefry:
## Correction: the "no approval received" paragraph above was overtaken by events

This PR's body states, in bold, that no approval had arrived for the ninth residual MLX-only cell and that **"the register is untouched — I will not author an approval record in his name."** That was true when the body was written. It stopped being true later the same session, and the body was never re-synced before merge.

**Michael Trefry approved `models/minimax_h3_ref/conditioningShape/referenceVideo` in session on 2026-08-20**, in direct answer to a question naming that cell path specifically. The approval is real. The record is legitimate.

It landed in this PR as commit [`6a9bc2856`](https://github.com/SceneWorks/SceneWorks/commit/6a9bc2856df927908307b32879dfc642601bd56f) ("chore(sc-18650): approve the ninth residual MLX-only cell exposed by the referenceVideo remap"), which added the cell to the existing `epic-17137-minimax-h3-ref-conditioning-sequencing` record in `config/backend-capabilities/exceptions.json` rather than opening a new one — that record's rationale (macOS-only family, no Candle lane to be at parity with) already covered it verbatim, so `approver`, `authority`, `approvedDate` and `revisitCondition` are unchanged. [`e3de184a5`](https://github.com/SceneWorks/SceneWorks/commit/e3de184a523dc85b738faeca9125764e90a853bb) regenerated `matrix.json` for it, and the register test noted as the one known-red in the body's test table is green on the merged tree.

### Why this is being corrected rather than left

The body of a merged PR is a permanent artifact. This one now sits next to a tree that contains exactly the record it promises not to write, so an auditor reading only the repo reaches the worst available conclusion — that the approval was fabricated. An independent audit did reach it.

Follow-up PR (branch `claude/sc-18650-approval-provenance`, into `feature/sc-17137-minimax-h3`) makes the record self-supporting so this cannot happen again:

- the record's `evidence` string was inherited from the two cells approved earlier that day and said nothing about the ninth. It has been rewritten to stand on its own: what was asked, that it was approved in session on 2026-08-20 as a separate second answer, that it is the ninth cell, and that it was made visible by the `referenceVideo => videoClip` mis-mapping fixed in **this** PR (`80c1ca842` moved `mlxOnlyCellCount` 72 -> 80 for the Ref2VA activation, then `655939b11` 80 -> 81 for the remap). It also carries a note pointing auditors at this comment.
- `config/backend-capabilities/README.md` still said "8 `minimax_h3_ref` cells" against 9 in the register; corrected.

The decision itself is untouched — `approver`, `authority`, `approvedDate`, `decisionType` and `cells` are all unchanged, and nothing is being withdrawn.

*[author: claude]*

PR #2513, opened 2026-08-22
SC-20764: pin LTX T2V overlay identity variants

michaeltrefry:
Adds paired SceneWorks coverage for the canonical LTX T2V admission overlay identity.

- Preserves separate enhancer and provider video-mode axes.
- Keeps no_audio from borrowing the plain null-overlay curve.
- Existing shared selector, exact identity evidence gate, lifecycle, and fail-open behavior remain in the merged SC-20763 foundation.

Validation: cargo test -p sceneworks-worker video_admission --lib; cargo fmt --all -- --check.


michaeltrefry:
Closing this premature candidate during epic closeout reconciliation. Its code-side work is retained, but parity is deterministically red because the pin-derived memory matrix is intentionally regenerated only after the permanent final inference pin. The paired SceneWorks change will be republished in the final pin/integration candidate rather than leaving an unmergeable red PR open.

PR #2533, opened 2026-08-23
feat(checkpoint): transactional managed ingestion for every import source (sc-20636)

michaeltrefry:
Managed checkpoint ingestion (sc-20636, epic 20398): one transactional staging → validation →
atomic-finalize path that upload, local copy, URL, Hugging Face and Civitai imports all share, the
managed counterpart of `compile_linked` in the checkpoint plan store, and an import API that gains
`ownershipMode` + a discriminated source while every legacy flat request keeps working.

Story: https://app.shortcut.com/trefry/story/20636

## AC1 — one staging → validation → atomic-finalize path for all five sources

`crates/sceneworks-core/src/checkpoint_ingest.rs` (new) owns the transaction:

```
begin()      mint the staging dir at <data>/models/.import-staging/<installId>/
  stage_*()  every byte written into staging; nothing outside it is touched
finalize()
  1. verify  SHA-256 of the staged primary artifact vs the source's declared digest
  2. validate full-content inspection of the staged tree (sc-20632 inspector)
  3. commit  ONE atomic rename: .import-staging/<id> -> models/imports/<id>
  4. publish compile the plan / record / bindings against the committed install
```

`run_model_import_job` now points the **existing, unchanged** transfer code at
`ingest.staging_dir()` — the HF snapshot downloader, the source-URL downloader (credential
attachment included, which is how Civitai transfers), and the local-copy/upload path are wrapped,
not duplicated. Civitai is a discriminated source that differs from `url` only in what it
**records**.

Because the rename is the only way bytes reach `models/imports/<id>`, cancel / crash / disk-full /
hash-mismatch all end the same way: no install directory, no plan, no catalog record, no manifest
entry. `ManagedIngest`'s destructor discards staging, so an early return, a `?`, a panic, or an
explicit `cancel()` all reach that state; a crash (no destructor) cannot create an install and its
staging directory is reclaimed by `sweep_staging`.

Provenance records where the bytes came from — `url`, `versionId`, `fileId`, `credentialHost` — and
never how they were authorized. `sanitize_provenance_url` strips userinfo and secret-bearing query
parameters (`token`, `api_key`, …) before a URL can be recorded, and
`ManagedProvenanceV1::validate` refuses userinfo again at the contract boundary.

## AC2 — linked/managed parity, duplicate reporting, ownership-safe removal

* `CheckpointPlanStore::compile_managed` is the managed counterpart of `compile_linked`, in the same
  store, through the same inspector, writing the same documents. `resolve` now resolves managed
  layers under their own install directory and refuses a managed layer naming a foreign install.
* **Defect found and fixed in place:** `ImportPlanV1`'s "locator-independent" semantic digest
  included `planId`, which the inspector derives from the checkpoint id — i.e. from ownership. A
  linked checkpoint and a managed copy of identical bytes therefore produced *different* semantic
  digests, which is exactly the equality E1 requires and duplicate detection is built on. `planId`
  is removed from the semantic form (`sourceBindingIdentity` still binds the exact document); the
  frozen golden in `checkpoint_import_contracts.rs` is updated, and the sc-20634 parity test that
  passed only because it reused the linked checkpoint id is corrected.
* `CompiledCheckpointV1.duplicate_checkpoint_ids` reports every other persisted checkpoint carrying
  the same semantic digest. Reported, never acted on: neither copy is deleted, neither compile is
  refused.
* `remove_managed` deletes the record, plan, bindings and the SceneWorks-owned bytes — and can only
  ever address a path under `installs_root()`, which is derived from the data dir, never supplied by
  a caller. `validate_install_id` is the choke point every install path goes through.

Managed installs stay at `<data>/models/imports/<installId>` — the tree imported models have always
occupied — so `paths.model`, the catalog install-state sweep, the external-library anchor and the
delete route are untouched.

## AC3 — `ownershipMode` + discriminated source, legacy shapes preserved

`ModelImportRequest` gains `ownershipMode` (default `managed`; anything else fails closed at
deserialization) and `source`, a `deny_unknown_fields` discriminated union over `upload`,
`localPath`, `url`, `huggingFace`, `civitai`. `normalize_model_import_source` collapses either
spelling onto the flat fields the rest of the import path already speaks, so both produce the same
job payload plus provenance. Setting both a discriminated source and a conflicting flat field is
refused rather than resolved by precedence. The multipart form takes the same two fields, decoded by
the same serde impls.

## Single-claim change (FIXED_EXTRA)

sc-20634 made the bespoke imported lanes decline any entry carrying `importPlan.checkpointId`. A
managed install carries that AND `paths.model`, and the plan-driven skeleton serves plain
text-to-image only — so stamping it unconditionally would have meant **importing a model removed its
edit, pose, reference, LoRA and Hires.fix support** until sc-20644. The claim is now per-request:
the plan route takes the shapes it serves, and the family's bespoke lane keeps the rest. Exactly one
lane owns each request.

The decline applies only to route-*capability* refusals (unsupported shape, unconsumed layer, no
adapter binding) and only when the entry has SceneWorks-owned bytes another lane can load. A LINKED
entry has no bespoke lane, so it still refuses loudly, and **integrity** refusals — drift, missing
source, tampered plan — stay fatal for every entry.

## Tests

| Area | Tests |
| --- | --- |
| `crates/sceneworks-core/tests/checkpoint_ingest.rs` (new, 18) | commit + published plan; the fault matrix; parity + duplicate report; layout-sensitivity of the digest; ownership-safe removal; removal confinement; managed drift / foreign-install locator / vanished install; provenance secret hygiene |
| `checkpoint_ingest.rs` unit (2) | URL sanitisation across every secret parameter name and case; staged-path escape refusals |
| `checkpoint_import_contracts.rs` | plan-id independence of the semantic digest, updated golden |
| `checkpoint_plan_store.rs` | the sc-20634 parity test corrected to a genuinely different managed identity; the managed-locator refusal retargeted |
| `image_jobs/tests.rs` | a managed entry keeps its bespoke lane for shapes the plan route does not serve; linked still refuses; integrity never declines |
| `tests/lora_and_downloads.rs` | end-to-end model-import job: atomic commit + plan binding stamped; hash mismatch leaves nothing; cancel leaves nothing |
| `models.rs` (6) | legacy flat bodies; every discriminated source; Civitai identity without a secret; both-spellings conflict; unserved ownership mode |

### Fault-injection matrix (AC1)

| Fault | Injected by | Asserted |
| --- | --- | --- |
| Cancel mid-stage | `ManagedIngest::cancel()` after staging; the job-level test drives it through `check_cancel` | no install dir, no record, no plan/bindings, no manifest entry, staging gone, user's source untouched |
| Crash before finalize | `std::mem::forget(ingest)` — a destructor that never runs | no install, no record, no plan; staging survives and `sweep_staging` reclaims it (and skips in-flight ids) |
| Disk full | a reader that yields half the bytes then reports ENOSPC; plus a real `File::create` failure on an unwritable staging dir after the first artifact landed | partial bytes stay in staging; no install, no record, no plan |
| Hash mismatch | `finalize` with a declared digest that is not the staged bytes' | typed `hash-mismatch`, no install, no record, no plan |
| Unrunnable content | a valid safetensors with no family evidence | the commit is rolled back; no install, no record, no plan |
| Drop without finalize | scope exit | same as cancel |

### Mutations run (one at a time, `touch` between runs, restored from a `cp` backup)

| Assertion | Mutation | Result |
| --- | --- | --- |
| the finalize is an atomic rename | stage directly into the install dir, drop the rename | RED (15 ingest tests, incl. all four fault cases) |
| an abandoned ingest leaves nothing | `Drop for ManagedIngest` emptied | RED (drop, disk-full) |
| linked ≡ managed semantic digest | `planId` back in `SemanticPlanV1` | RED (parity + contract golden) |
| duplicates are reported | `duplicates_of` returns empty | RED |
| removal cannot leave `installs/` | `install_dir` skips `validate_install_id` | RED |
| provenance carries no secret | `sanitize_provenance_url` returns the URL unchanged | RED |
| the contract refuses userinfo | `validate_secret_free_url` check disabled | RED |
| per-request single claim | claim ignores the request shape | RED |
| a linked entry with no bespoke lane refuses | `checkpoint_plan_unservable` always declines | RED (managed-lane + sc-20634 refusal tests) |
| the plan binding is stamped | `importPlan` insert removed | RED |
| conflicting spellings refuse | the three conflict guards disabled | RED |

## Local gates

`cargo fmt --all -- --check`, `npm run rust:check` (incl. `check:memory-matrix` after regenerating
`docs/generated/memory-matrix.{json,md}` — `routing/catalog.rs` is a fingerprinted source; the diff
is the source-tree revision only). No candle-gated code touched. Inference pin untouched.

## Behaviour notes

* A model import no longer resumes a partially-transferred download across job *retries*: each run
  stages fresh. Resume within a single run is unchanged. See *Behaviour notes (issue 15)* below for
  the decision to keep staging keyed on the install id, and for the post-crash retry consequence.
* `targetDir` must now be the managed install directory for the model id (`models/imports/<id>`); a
  target elsewhere under `data/models` is refused rather than silently installed outside the store.

---

## Review round 2 (sc-20636)

Every finding from the review is resolved in this push; each new assertion was pinned with its own
named mutation, run one at a time (`cp` backup → mutate → `touch` → RED → restore → `touch` → GREEN).

### Blockers

1. **`$defs/provenance` could not express what `ManagedProvenanceV1` serializes.** The schema is
   `additionalProperties: false` with only `source`/`reference`, but the type now serializes `url`,
   `versionId`, `fileId`, and `credentialHost` — so every url / HF / Civitai managed locator was
   schema-invalid while serde accepted it. The alignment test stayed green only because no mutation
   in `aligned_locator_mutations` carried a post-`reference` field. All four are added as
   nonBlankText-or-null, and the mutation list now carries each one set, each one blank, one
   all-null, and one fully populated. Admitting `url` also makes the userinfo refusal a
   *semantic-only* rule (JSON Schema can require a non-blank string but not the absence of a
   `user:token@` authority), so that gap is asserted explicitly and documented in
   `packages/schemas/README.md` alongside the other four.
2. **Zero-lane fall-through completed a plan-backed job with procedural stub output.** A managed
   plan-backed entry whose family has no bespoke lane for the declined shape got `Ok(None)` from
   `prepare_image_route` and landed in `generate_stub_stream` — the job *completed*, with fake
   pixels, instead of surfacing `[checkpoint-plan:no-adapter-binding]`. The plan route's decline now
   RETAINS its refusal (`CheckpointPlanSelection`), and both `prepare_image_route` and
   `prepare_candle_image_route` re-raise it when no lane claims the request. The `zimage_managed`
   case asserts the terminal outcome is a typed refusal naming the checkpoint, never a stub, and its
   misleading "declines to its bespoke lane" message is corrected — `z-image` has no such lane, which
   is the whole point.
3. **`delete_model` never removed the managed checkpoint.** The record, plan and bindings survived a
   delete, so re-importing the same id passed the API's `existing_ids` check and then failed forever
   in the worker with `InstallIdTaken` naming a path that no longer existed. `delete_model` now
   calls `CheckpointPlanStore::remove_managed` for an entry whose `importPlan.checkpointId` is a
   `managed/…` id; a linked id falls through untouched (that branch is sc-20635's, and is left as an
   explicit no-op so their diff lands beside this one rather than through it). A `remove_managed`
   failure warns rather than failing a delete whose files are already gone.

### Major

4. **`sweep_staging` had no production caller** while the module doc promised startup reclamation.
   `sceneworks_worker::reclaim_import_staging` now runs it at worker startup, spawned blocking and
   never awaited, beside the retention checkpoint. It is **age-gated** via the new
   `checkpoint_ingest::active_staging_ids`: a SceneWorks install runs several worker processes
   against ONE data dir, so an unconditional sweep at worker B's startup would delete the
   multi-gigabyte tree worker A is still downloading into. A live transfer keeps the newest mtime
   under its tree fresh; an orphan's stopped at the crash. Pinned by behaviour *and* by call site —
   the behavioural half alone stays green with the call deleted, which is exactly how this shipped.
5. **`SECRET_QUERY_PARAMS` matched whole names**, so a presigned S3/CloudFront redirect persisted its
   full signature into a world-readable plan document: `X-Amz-Signature`, `X-Amz-Credential`,
   `X-Amz-Security-Token`, `Key-Pair-Id` and `Policy` all survived. Matching is now case-insensitive
   **substring** over `{token, secret, password, signature, sig, credential, key, policy}`. The unit
   test names those five explicitly plus `Signature`/`AWSAccessKeyId`, and runs a whole presigned URL
   through, asserting no signature, credential or session token survives while a benign
   `response-content-type` does. Over-redaction is the deliberately cheap side of the trade:
   provenance is a human reference, never a re-fetch handle.
6. **Duplicates reached only `tracing::info!`.** `duplicateCheckpointIds` is now on the import job
   result (always present, empty array when there are none). "This is the same checkpoint you already
   have as X" is the user's decision — nothing deletes either copy — and a log line never reaches
   them. Asserted against a pre-existing **linked twin** compiled from the same bytes before the
   import runs, with both copies still resolving afterwards.
7. **`checkpoint_plan_entry_has_bespoke_lane` read a path it must not.** It reused
   `imported_entry_installed_path`, whose last fallback is the provenance-only `source.path` — a
   data-dir-relative breadcrumb. A linked entry recording one claimed a bespoke lane it does not have.
   Split into `imported_entry_loadable_path` (`modelPath` / `paths.model` / `installedPath`) for the
   lane predicate, with `imported_entry_installed_path` unchanged for admission — "does this entry
   describe bytes anywhere" is a different question from "can a lane load it".
8. **`begin` deleted an existing staging tree.** Two sessions for the same install id could both
   "succeed": B's `begin` `remove_dir_all`'d A's in-flight bytes, then A's `Drop` deleted B's — two
   live transfers destroying each other with no error on either side. `begin` now uses `fs::create_dir`
   with no pre-removal and maps `AlreadyExists` to `InstallIdTaken`; the directory itself is the
   mutual exclusion, and orphans are `sweep_staging`'s job.

### Minor

9. Module-doc pipeline renumbered to verify → commit → validate-and-publish (rollback on refusal),
   matching what `finalize_inner` actually does.
10. The three "exactly one lane per entry" comments in `image_jobs/base.rs` (and the `checkpoint_plan.rs`
    module header) rewritten to the per-REQUEST claim semantics: an entry may be served by two lanes,
    a request by exactly one, and a request no lane claims is a typed refusal.
11. **Decision: `source.provider` keeps `"local"` for both non-remote shapes.** This PR had silently
    changed it to `"upload"` / `"local-copy"`. The manifest `provider` vocabulary is deliberately
    coarser than provenance — existing entries and their readers say `"local"` — and the finer kind is
    already recorded in `importProvenance.source`. All five mappings are pinned by test.
12. `expectedSha256` set on both the discriminated source and the flat field no longer resolves by
    precedence: it refuses when the two differ (normalized comparison, so the same digest spelled
    twice is accepted), matching what the function's own doc already claimed.
13. `stage_copy_dir` silently SKIPPED any entry that is neither a regular file nor a directory —
    `file_type()` does not follow links, so an HF-cache snapshot dir (entirely symlinks into `blobs/`)
    would have staged EMPTY with no error anywhere. It now refuses with a typed
    `unsupported-source-entry` naming the entry; following the link is not the fix either, since that
    reads bytes outside the directory the user pointed at.
14. An explicitly supplied `modelId` is validated at queue time against the install-id constraint
    (`checkpoint_plan_store::validate_install_id`, now `pub`), on the RESOLVED `target_name` the worker
    will use — 400 with the constraint named, instead of 201 followed by a worker-side death.

### Behaviour notes (issue 15)

Two accepted regressions from pre-transaction behaviour, both consequences of "there is exactly one
partial state and it is not addressable". Both are documented in the `checkpoint_ingest` module doc.

* **Cross-retry download resume is gone.** A partial transfer used to accumulate in the final install
  directory, so a job retry continued where the last attempt stopped; a retry now starts a fresh
  staging tree. Resume *within* one attempt is unchanged. **Decision: staging stays keyed on the
  install id, not the job id.** Keying on job id would restore resume but break the mutual exclusion
  issue 8 just established — two jobs for the same install id could both stage, and the second's
  commit rename would land on an existing install. More importantly, an install assembled from two
  attempts' bytes (possibly two revisions') is precisely the silent substitution this module exists
  to remove.
* **After a crash** (not a refusal — a refusal discards), a retry of the same install id refuses with
  `InstallIdTaken` naming the *staging* path until the startup sweep reclaims the orphan. `begin`
  cannot distinguish a crashed session's tree from another worker process's live one, and destroying
  a live transfer is the worse failure.

### Mutations run (issue → assertion → mutation → result)

| # | Assertion | Mutation | Result |
| --- | --- | --- | --- |
| 1 | schema admits `provenance.url` | drop `url` from `$defs/provenance` | RED → GREEN |
| 1 | schema admits `provenance.credentialHost` | drop `credentialHost` from `$defs/provenance` | RED → GREEN |
| 1 | a blank optional field is refused on both sides | loosen `versionId` to a plain `{"type":"string"}` | RED → GREEN |
| 1 | serde is the gate on a userinfo url | disable the `authority.contains('@')` check | RED → GREEN |
| 2 | a plan-backed request no lane claims refuses | drop the `into_unclaimed_refusal` re-raise (back to `Ok(None)`) | RED → GREEN |
| 4 | a live staging tree is never swept | invert the age comparison in `active_staging_ids` | RED → GREEN |
| 4 | age is the newest mtime under the tree | `newest_mtime` reads only the root's own mtime | RED → GREEN |
| 4 | worker startup runs the sweep | remove the `reclaim_import_staging` call from `run_worker_loop` | RED → GREEN |
| 5 | presigned parameters are stripped | revert substring matching to whole-name equality | RED → GREEN |
| 6 | duplicates reach the job result | drop the `duplicateCheckpointIds` insert | RED → GREEN |
| 7 | a provenance `source.path` is not a lane | predicate reads `imported_entry_installed_path` again | RED → GREEN |
| 8 | a second session for one id refuses | restore the `remove_dir_all` pre-removal in `begin` | RED → GREEN |
| 13 | a symlinked entry refuses | restore the silent skip in `stage_copy_dir` | RED → GREEN |
| 3 | `record("managed/…")` is `Err` after a delete | `remove_managed` replaced by an `inventory()` read of the same type | RED → GREEN |
| 3 | a LINKED entry is not torn down | `managed_install_id(id)` → `Some(id)` (drops the discrimination) | RED → GREEN |
| 3 | the warning names the checkpoint id | drop `{checkpoint_id}` from the warning text | RED → GREEN |
| 3 | a teardown failure still returns 200 | `warnings.push` → `return Err(ApiError::internal(..))` | RED → GREEN |
| 11 | `source.provider` is the coarse value | `manifest_provider()` → `provenance.source` | RED → GREEN |
| 11 | `upload` maps to `"local"` too | drop `"upload"` from the match arm | RED → GREEN |
| 12 | the Url arm refuses conflicting digests | Url arm back to `expected_sha256.or(payload…take())` | RED → GREEN |
| 12 | the Civitai arm refuses conflicting digests | Civitai arm back to `.or(…)` | RED → GREEN |
| 12 | a case/whitespace-only difference is accepted | `normalize` = `trim().to_ascii_lowercase()` → `to_owned()` | RED → GREEN |
| 14 | an invalid resolved id is refused | `validate_install_id(&target_name)` → `validate_install_id("ok")` | RED → GREEN |
| 14 | the RESOLVED name is what is validated | `validate_install_id(&target_name)` → `validate_install_id(&model_id)` | RED → GREEN |

### Two things worth a second opinion

* **Non-permanent delete now tears the plan documents down too.** With `permanent=false` (the
  default) the bytes go to the OS trash and are recoverable, but the record/plan/bindings are removed
  outright. That asymmetry is deliberate rather than overlooked: `permanent` defaults to `false`, so
  skipping teardown there would leave blocker 3 unfixed on the path almost every delete takes. It is
  not a new recovery regression either — `remove_catalog_manifest_entry` already runs on that path,
  so a trashed model was never restorable through SceneWorks without a re-import. Restoring bytes
  from the trash by hand would leave a plan-less install tree.
* **`source.provider` has no reader on a MODEL manifest entry today** — a repo-wide sweep found the
  field is write-only for models (every reader found belongs to the LoRA, control-overlay or
  workflow-share manifests, which are a different entry shape). So issue 11's fix is vocabulary
  stability, not a live bug fix; `"upload"`/`"local-copy"` would not have broken anything reachable.

### Local gates (round 2)

`cargo fmt --all -- --check`, `npm run rust:check`, `npm run rust:check:candle`,
`npm run rust:check:neither`, and `pytest tests/test_builtin_manifest_audit.py tests/test_memory_matrix.py`
(104 passed) — all clean. `docs/generated/memory-matrix.{json,md}` regenerated via the positional
generator because `routing/catalog.rs` is a fingerprinted source; the diff is the two source hashes
only, no matrix cell moved. The candle config caught one MSRV violation in the new age-gate
(`is_none_or` is 1.82+, workspace MSRV is 1.80) — rewritten as `map_or(true, ..)`. Inference pin
untouched; no GPU or real-weight work.


michaeltrefry:
## Review round 2 — status at head `6a9a80633`

**DONE — all 16 findings resolved, each with a named mutation run individually** (`cp` backup →
mutate → `touch` → RED → restore → `touch` → GREEN; never `git checkout` to restore). 23 mutations,
all RED→GREEN. Full detail and the mutation table are in the PR description.

| # | Finding | Status |
| --- | --- | --- |
| 1 | `$defs/provenance` could not express `url`/`versionId`/`fileId`/`credentialHost` | DONE — 4 properties added, alignment mutations extended (each set, each blank, all-null, fully populated); the userinfo gap is now a documented semantic-only rule |
| 2 | zero-lane fall-through completed a plan-backed job with stub output | DONE — the decline retains its refusal (`CheckpointPlanSelection`); both route preparers re-raise it; `zimage_managed` asserts the typed refusal and its premise |
| 3 | `delete_model` never called `remove_managed` | DONE — managed teardown added, linked arm an explicit no-op naming sc-20635 |
| 4 | `sweep_staging` had no production caller | DONE — `reclaim_import_staging` at worker startup, age-gated via new `active_staging_ids`; pinned by behaviour AND call site |
| 5 | `SECRET_QUERY_PARAMS` missed presigned S3/CloudFront names | DONE — case-insensitive substring matching; whole presigned URL asserted clean |
| 6 | duplicates reached only `tracing::info!` | DONE — `duplicateCheckpointIds` on the job result, asserted against a pre-existing linked twin |
| 7 | bespoke-lane predicate read the provenance-only `source.path` | DONE — split into `imported_entry_loadable_path`; admission unchanged |
| 8 | `begin` deleted an existing staging tree | DONE — `create_dir` with no pre-removal, `AlreadyExists` → `InstallIdTaken` |
| 9 | module-doc pipeline numbering | DONE |
| 10 | three stale "exactly one lane per entry" comments | DONE — plus the `checkpoint_plan.rs` header |
| 11 | `source.provider` vocabulary change | DONE — **kept `"local"`** for both non-remote shapes; five mappings pinned |
| 12 | `expectedSha256` resolved by precedence | DONE — refuses when both set and differing |
| 13 | `stage_copy_dir` silently skipped symlinks | DONE — typed `unsupported-source-entry` refusal |
| 14 | explicit `modelId` never validated | DONE — `validate_install_id` on the RESOLVED `target_name` at queue time |
| 15 | resume regression unrecorded | DONE — **staging stays keyed on the install id**, not the job id; recorded in the module doc and the PR body, along with a second previously-unrecorded consequence (post-crash retry refuses until the startup sweep) |
| 16 | covered by #2 | DONE |

**OUTSTANDING: nothing in the review.** One verification item is unconfirmed at the time of writing:
the `macOS build, lint and workspace tests (hosted)` lane is still `in_progress`. Every other lane is
green on this head (build-windows, candle, check-linux, check-macos, parity, parity-rust,
parity-scaffold, parity-docker, parity-digests, web; nax-worker and package-windows skip by path
filter). The same workspace suite ran clean locally via `npm run rust:check`, along with
`rust:check:candle`, `rust:check:neither`, `cargo fmt --all -- --check`, and the Python manifest
audit + memory-matrix tests (104 passed). **Please confirm that last lane before merging.**

Two judgement calls flagged for a second opinion (detail in the PR description): non-permanent delete
now tears the plan documents down too (deliberate — `permanent` defaults to `false`, so skipping it
there would leave blocker 3 unfixed on the common path), and `source.provider` turns out to have no
reader on a model manifest entry today, so #11 is vocabulary stability rather than a live bug fix.

The `delete_model` change is written to co-exist with #2534: its current head does not touch
`delete_model` at all, and the linked branch here is left as an explicit empty arm for that work.

michaeltrefry:
Follow-up to the round-2 status comment: the one item I flagged as unconfirmed is now resolved. `macOS build, lint and workspace tests (hosted)` **passed** on head `6a9a80633`.

CI is fully green — 15 pass, 0 fail, 0 pending (build-windows, candle, check-linux, check-macos, macOS build/lint/workspace tests, parity, parity-rust, parity-scaffold, parity-docker, parity-digests, web, and the four `changes / decide` gates); nax-worker and package-windows skip by path filter.

Also re-ran `npm run rust:check:candle` and `npm run rust:check:neither` against the exact committed tree — both had originally run just *before* the final `cargo fmt --all`, so this closes that gap. Both green, `cargo fmt --all -- --check` clean, working tree clean at `6a9a80633`.

Nothing outstanding on my side.

PR #2534, opened 2026-08-23
feat(checkpoint): linked-library lifecycle and the derivative cache (sc-20635)

michaeltrefry:
Story: [sc-20635](https://app.shortcut.com/trefry/story/20635) — Linked-library lifecycle (rescan/relink/remove) and checkpoint-derivative cache. Targets `feature/sc-20398-universal-checkpoint-import` (sc-20634's skeleton merged; that feature head is merged in here).

### AC1 — lifecycle + a linked checkpoint becomes selectable

| requirement | where |
| --- | --- |
| add / rename / rescan / relocate / remove / relink a root | `checkpoint_plan_store.rs`: `approve_root_with_label`, `rename_root`, `relink_root`, `remove_root`, `scan_root`, `rescan_checkpoint` — **reachable from a client** through `apps/rust-api/src/checkpoint_library.rs` |
| a linked checkpoint becomes a selectable manifest entry carrying `importPlan.checkpointId` | `model_jobs::linked_checkpoint_manifest_entry`, stamped immediately before `upsert_manifest_entry` — the seam the story named |
| persisted identity is `rootId + relativePath`, never an absolute path | identity is unchanged from sc-20634; the API request takes `linkedRootId`/`linkedRelativePath`, and both the queued entry and the stamped entry are asserted to contain no library path |
| Needs Relink / Needs Rescan | `LinkedCheckpointStateV1` + `checkpoint_status`, classified from the store's own typed refusals (root absent ⇒ Relink; bytes ⇒ Rescan), **serialised by `GET .../:rootId/scan`** |
| never modifies or deletes a linked file | removal is store-scoped by construction; proven by full-tree size+mtime+bytes snapshots across every action |
| header-only candidates visible but unselectable until compilation succeeds | `checkpoint_inspector::discover_library_root` (bounded, header-only) + `LinkedCandidateV1::selectable`, which is true only for a `Ready` persisted plan |

A relink keeps the root id, so compiled plans survive a library move without recompiling. One directory is now bound to exactly one root id — re-approving a relinked directory returns its existing binding, and a duplicate binding is a corrupt-document refusal. (This changes one sc-20634 assertion in `approving_a_colliding_root_id_refuses_instead_of_rebinding`; the old expectation would have produced exactly the duplicate binding the new invariant forbids, so the test was updated with the reason rather than the invariant weakened.)

**HTTP surface (added in review).** Every AC1 verb crosses the wire, returning `ApprovedRootV1` / `LinkedRootScanV1` / `LinkedCheckpointStatusV1`:

| verb | route |
| --- | --- |
| list / add | `GET` / `POST /api/v1/models/library-roots` |
| rename | `PATCH /api/v1/models/library-roots/:rootId` with `label` |
| relocate / relink | `PATCH /api/v1/models/library-roots/:rootId` with `path` |
| rescan (library) | `GET /api/v1/models/library-roots/:rootId/scan` |
| rescan (one checkpoint) | `POST /api/v1/models/library-roots/:rootId/rescan` with `relativePath` |
| remove | `DELETE /api/v1/models/library-roots/:rootId` |

`library-roots` is a STATIC sibling of `/api/v1/models/:model_id`, exactly like the existing `/api/v1/models/import`, so a library operation can never be mistaken for a model id. Refusals are the plan store's own typed `[checkpoint-plan:<code>]` diagnostics, status-mapped (404 unknown / 400 malformed / 409 library-state / 500 store) with the stable code in `context.reason`, so a client branches on the code rather than on prose. The two path-bearing writes (approve, relink) are loopback-only, exactly like `model_library::relocate_model_library` and for the same two reasons: the typed refusals are a filesystem-existence oracle, and a token-holding LAN client could otherwise re-point a shared server's library.

### AC2 — probes fail closed

Every lifecycle action funnels through `confined_root_join`, which canonicalizes the path it is about to open and requires it to stay under the canonical root. The scanner re-confines each candidate on that same joined path, not on the discovery string, and drops an escaping one with a `PathEscapesRoot` diagnostic. `validate_linked_relative_path` is now the single lexical validator, called by the API so it cannot drift from what `compile_linked` enforces.

Probes: traversal / absolute / backslash shapes; a symlink inside the library pointing out; **replacement** (a compiled source swapped for a symlink to byte-identical content outside the root — the fingerprint check alone passes it, only confinement refuses it); a symlinked root resolved to its real directory at approval. Windows junction/reparse is `#[cfg(windows)]` and now actually runs on the Windows CI lane (see Review fixes).

### AC3 — derivatives in the epic-19703 resolved-artifact cache

`checkpoint_derivative.rs` keys a derivative on **content** (every plan layer's source fingerprint) + **semantic plan** (`semanticDigest`, family) + **adapter/codec version** + **backend representation**, then looks it up, produces it if absent, and leases it — entirely through the existing store: same entries directory, same journal, same locks, same staging, same atomic publish, same completion receipt, same retention/pinning/eviction. No second cache.

Two things had to be added to the store rather than worked around, both documented at their definitions:

1. **`reserve_derived` / `publish_produced`.** The copy path refuses to reserve without a source-library second copy of every byte; a computed artifact has none until its producer runs. The derived reservation stages an entry whose bundle is still to be written and enriches/publishes it through the same path. The narrow consequence is that a derived entry's artifact is shape-validated (identity, provenance agreement, canonical closure, cache key) rather than content-validated **while it is not `Complete`** — there is nothing on disk to read — and it must record itself `Incomplete`/`Missing`, so an unfinished bundle can never be mistaken for a loadable one.
2. **`ResolvedCacheProduction`.** Retention evicts a copied bundle only while the source library proves a complete second copy. A derivative has none by construction, so it would be permanently unevictable and the cache would grow past its own limit. `Derived` skips *that one* proof; pins, leases, freshness and shape re-validation are untouched.

Never admitted stale or partial: a drifted checkpoint refuses before a key is even computed; a declared output the producer did not write refuses at publication; an undeclared file it did write refuses too (it would leave bytes the entry's accounting never counts); a failed producer's staging tree is discarded and the entry recorded interrupted. Conversion never overwrites source weights — the producer gets read-only resolved layer paths and a `create()` that only accepts declared, confined paths under staging.

`derivedFrom` is recorded beside the key (not in it) so "forget this checkpoint / this root" can reach a derivative; the key itself stays content-addressed.

## Review fixes

* **HTTP surface** — see the AC1 table above. `approve_root` / `rename_root` / `relink_root` / `remove_root` / `scan_root` / `rescan_checkpoint` / `checkpoint_status` had no non-test caller, `LinkedCheckpointStateV1` was `Serialize`-derived but never serialised by a route, and with no approve route the `linkedRootId` the import seam validates could never be produced by a client. Coordinator ruling: AC1 has to be true from a client the day this merges; sc-20650 is the UI on top of these routes, not their owner.
* **Cross-process locking** — every mutator was an unlocked read-modify-write over `approved-roots.json` / `inventory.json` while the API and the worker are separate processes (`write_atomic` makes each write atomic, not the read-modify-write pair). All six (`approve_root_inner`, `rename_root`, `relink_root`, `remove_root`, `upsert_record`, `invalidate`) now take an `fs2` exclusive lock on `<data>/checkpoints/.lock`, the same primitive as `resolved_cache::lock_metadata`. `remove_root` also stopped invalidating N checkpoints with `?` before saving the root list: it rewrites the inventory once, so no failure can leave a half-removed root.
* **Windows junction probe** — `desktop-windows.yml` ran `cargo test -p sceneworks-core` with two name filters, neither of which reaches an integration test, so the probe never ran; it also would have failed, because `fs::canonicalize` yields `\\?\C:\…` and `cmd /C mklink /J` rejects that form. Added a `cargo test -p sceneworks-core --test checkpoint_plan_store` step and strip the verbatim prefix from both `mklink` arguments (plus a SANITY assertion that the junction exists).
* **Phase-two lease guard** — the recorded failing mutation was wrong. Measured, one guard at a time: scan-phase probe alone no-opped ⇒ both tests GREEN; phase-two arm alone no-opped ⇒ the derivative lease test GREEN and the new test RED; both removed ⇒ the derivative lease test RED. The doc comment now states exactly that and names the test that pins each guard. The new `the_phase_two_artifact_lock_retains_an_entry_leased_after_it_was_classified` leases in the window between the two guards, via the scan-classification observer (now multi-shot and digest-aware, and explicitly disarmed so it is safe under `--test-threads=1`).
* **`approve_root_with_label`** no longer discards a differing label for an already-approved directory; an unlabelled re-approval is still a no-op.
* **Import trimming** — the route validated the trimmed pair and serialised the raw one, so `" root-abc "` passed the API and failed in the worker while the queued `manifestEntry` and the job payload disagreed. The validated values are written back into the payload before `to_json_object`.
* **`scan_root`** collects a per-checkpoint corrupt/IO failure as a diagnostic (severity Error, code Io) and continues, so one damaged document no longer unlists every other checkpoint under the root.
* **`delete_model`** reaches `CheckpointDerivativeStore::invalidate_checkpoint` for a plan-backed entry, so a deleted linked model no longer strands its record, plan, bindings and derivatives. Failures become warnings, because the manifest entry is already gone by that point. E6 unchanged: the library file and the approved root both survive.
* **`publish_produced`** refusals (undeclared staged file, symlinked declared output) now have in-crate tests at the boundary they guard.
* **Non-gated stamp test** — both reachability tests are macOS-gated while `linked_checkpoint_manifest_entry` is platform-independent, so the "remove the stamp ⇒ nothing can select it" mutation was caught on one lane of three. Added a non-gated test in the existing `image_jobs/tests.rs` (a test-only file, so the parity lane's `#[cfg(test)]` truncation hazard does not apply).
* **Journal version (DONE_STORY_IMPACT)** — a DERIVED entry declares `RESOLVED_CACHE_DERIVED_STORE_VERSION = 2`; a source copy stays at 1 and is byte-identical (both new fields are `skip_serializing_if`-defaulted), confirmed by the full existing suite (1175 lib tests + every integration target, all green). Bumping the store-wide constant instead was **rejected as genuinely incompatible**: it would invalidate every entry already on disk, and retention cannot reclaim an entry it cannot read, so a warm cache would be stranded as unreclaimable recovery candidates. `read_journal` now probes the declared version before decoding the document, so a forward-version journal is a typed "written by a newer SceneWorks" refusal rather than being reported corrupt — which matters because "corrupt" is the one classification manual removal is allowed to clear.

## Tests

| suite | notes |
| --- | --- |
| `crates/sceneworks-core/tests/checkpoint_plan_store.rs` (26) | +4 in review: two-handle store-lock interleaving (lost update), every-mutator lock serialisation, label adoption on re-approval, one corrupt document does not unlist the library |
| `crates/sceneworks-core/tests/checkpoint_derivative.rs` (13) | key composition, cache reuse, producer refusals, partial bundle, drift, lease, derived eviction, pin, root removal, altered bundle, two kinds; lease-test doc comment rewritten to the measured truth |
| `crates/sceneworks-core/src/resolved_cache/tests.rs` | +3: derived vs source-copy journal version and byte shape, forward-version typed refusal, `publish_produced` undeclared-file and symlink refusals |
| `crates/sceneworks-core/src/resolved_cache/retention_tests.rs` | +1: the phase-two artifact-lock guard **alone** |
| `crates/sceneworks-worker/src/image_jobs/tests.rs` | 2 macOS-gated (approve → compile → import stamp → `prepare_image_route` ⇒ `CheckpointPlan`; moved library → refusal → relink → selectable) **+1 non-gated** stamp assertion |
| `apps/rust-api/src/tests/checkpoint_library.rs` | new (4): the whole AC1 lifecycle over HTTP, typed/status-mapped refusals, padded-import trimming, `delete_model` forgetting a plan-backed model without touching the library |
| `apps/rust-api/src/tests/catalog.rs` | linked `POST /api/v1/models/import` variant (unchanged from the original push) |

`prepare_image_route` is private to the worker's backend-gated image lane, so the HTTP half and the routing half cannot live in one test; they join on the checkpoint identity `linked/<rootId>/<relativePath>`, which both assert explicitly.

Ran here (macOS, CPU only): the full `sceneworks-core` suite and every integration target, the worker and rust-api suites, `cargo fmt --all -- --check`, **all three configs** — `npm run rust:check`, `npm run rust:check:candle`, `npm run rust:check:neither` — and `tests/test_builtin_manifest_audit.py` (95 passed). No fingerprinted memory-matrix source is touched, so no regeneration. Windows-lane-only: the junction/reparse probe.

## Mutations run

| --- | --- | --- |
| replacement / symlink probes | delete the `starts_with(canonical_root)` refusal in `confined_root_join` | **RED** (2 tests) |
| candidate unselectable until compiled | force `selectable: true` in `scan_root` | **RED** |
| never deletes under the library | make `remove_root` also delete the checkpoint file | **RED** |
| derived entry is evictable | delete the `production.is_derived()` arms in `evict_candidate` | **RED** |
| leased derivative not evicted | no-op the SCAN-phase `artifact_lock_is_contended` guard alone | GREEN (recorded, not hidden) |
| " | no-op the PHASE-TWO `is_lock_contended` arm alone | GREEN |
| " | remove **both** | **RED** |
| phase-two guard alone (new test) | no-op the phase-two `is_lock_contended` arm | **RED** (entry evicted under its lease) |
| " | no-op the scan-phase guard | GREEN — the isolation this test exists for |
| no lost update across processes | delete `lock_store()` from `remove_root` | **RED** (both lock tests) |
| mutator serialisation | delete `lock_store()` from `approve_root_inner` / `rename_root` / `relink_root` / `upsert_record` / `invalidate`, each alone | **RED** (5 separate runs) |
| label adoption | restore the plain `return Ok(existing.clone())` | **RED** |
| corrupt document tolerance | restore the `?` on `checkpoint_status` in `scan_root` | **RED** |
| derived journal version | `store_version()` returns `RESOLVED_CACHE_STORE_VERSION` for `Derived` | **RED** |
| forward-version typed refusal | disable the version probe in `read_journal` | **RED** (falls back to "both slots corrupt") |
| undeclared staged file | delete the `reject_undeclared_produced_files` call | **RED** (bundle publishes) |
| symlinked declared output | delete the symlink/reparse refusal in `measure_produced_file` | **RED** |
| import trimming | delete the two `payload.linked_* = Some(...)` write-backs | **RED** |
| `delete_model` reaches the plan store | delete the `forget_linked_checkpoint` call | **RED** (record survives) |
| import stamp (non-gated) | delete the `importPlan` insert | **RED** |
| " | delete the `paths.remove("model")` | **RED** |
| routes are wired | remove the scan route from the router | **RED** (404) |

## E1 — resolved by the sc-20636 merge

This PR originally flagged that `ImportPlanV1::semantic_digest` folded in `plan_id`, so two
approved roots holding byte-identical files digested differently and produced two derivative
entries. **sc-20636 removed `plan_id` from `ImportPlanV1::semantic_form`**, so the digest is now
content-and-routing only and full E1 locator independence holds: byte-identical checkpoints reached
through two different libraries share ONE derivative entry instead of paying for the same
production twice.

`the_derivative_key_covers_content_plan_adapter_codec_and_backend` was updated from `assert_ne!` to
`assert_eq!` with the reason recorded in place — the test follows the contract rather than pinning
the behaviour it used to have. Sharing is safe in both directions: `derived_from` records only which
checkpoint's producer created a bundle, so forgetting one twin can cost the other a re-production
but can never lose data.

Nothing is left open for sc-20651 here.

## Coordination

Additive and narrowly named for the concurrent sc-20636 (managed ingestion): the only shared surface is two new `Option<String>` fields on `ModelImportRequest` (`linked_root_id`, `linked_relative_path`) plus its one struct literal in `model_import_request_from_multipart`, and an early dispatch at the top of `run_model_import_job`. The review fixes add one new rust-api module (`checkpoint_library.rs`), its four route registrations in `lib.rs`, and one new test module — all additive. Whoever merges second merges the other's branch first.

Inference pin untouched. Do not auto-merge.


michaeltrefry:
Review fixes pushed — head `4725dc5b3`, CI fully green (14/14; `nax-worker` and `package-windows` are skipped by trigger, not by path).

All 4 major + 6 minor findings and the DONE_STORY_IMPACT caveat are resolved. The PR body carries the full detail; this is the checklist.

### DONE

| # | finding | resolution | proving test |
| --- | --- | --- | --- |
| major 1 | lifecycle primitives had no non-test caller; `LinkedCheckpointStateV1` never serialised; no approve route ⇒ `linkedRootId` unproducible | new `apps/rust-api/src/checkpoint_library.rs` — `GET/POST /api/v1/models/library-roots`, `PATCH\|DELETE .../:rootId`, `GET .../:rootId/scan`, `POST .../:rootId/rescan`; typed refusals status-mapped with the store's `code()` in `context.reason`; the two path-bearing writes are loopback-only | `the_library_root_lifecycle_is_reachable_over_http`, `library_root_refusals_are_typed_and_status_mapped` |
| major 2 | `#[cfg(windows)]` junction probe never ran, and would have failed on `\\?\` paths | added a `cargo test -p sceneworks-core --test checkpoint_plan_store` step to `desktop-windows.yml`; strip the verbatim prefix from both `mklink /J` arguments + a SANITY assertion that the junction exists | **ran on this SHA**: `a_junction_out_of_the_library_fails_closed ... ok` (24 passed) |
| major 3 | recorded failing mutation for the lease test was wrong; phase-two guard uncovered | doc comment rewritten to the measured truth; new test leases in the window between the two guards via the scan-classification observer (now multi-shot + digest-aware, explicitly disarmed for `--test-threads=1`) | `the_phase_two_artifact_lock_retains_an_entry_leased_after_it_was_classified` |
| major 4 | unlocked read-modify-write over `approved-roots.json` / `inventory.json` across two processes; `remove_root` half-removal | `fs2` exclusive lock on `<data>/checkpoints/.lock` in all six mutators; `remove_root` rewrites the inventory once instead of N times with `?` | `a_mutator_cannot_lose_a_concurrent_writers_record`, `every_store_mutator_serialises_on_the_store_lock` |
| minor 1 | `approve_root_with_label` discarded a differing label | relabels and saves in the `get_by_path` branch; an unlabelled re-approval stays a no-op | `re_approving_a_root_under_a_new_label_relabels_it` |
| minor 2 | validated trimmed, serialised untrimmed | validated values written back into the payload before `to_json_object` | `a_padded_linked_import_queues_the_validated_values` |
| minor 3 | one corrupt plan document unlisted the library | collected as a diagnostic (severity Error, code Io) and the scan continues | `one_corrupt_plan_document_does_not_unlist_the_library` |
| minor 4 | `publish_produced`'s symlink / undeclared-file refusals untested | in-crate tests at the boundary | `a_produced_bundle_holding_an_undeclared_file_is_never_published`, `..._whose_declared_output_is_a_symlink_...` |
| minor 5 | `delete_model` stranded a linked model's record/plan/bindings/derivatives | calls `CheckpointDerivativeStore::invalidate_checkpoint`; failures degrade to warnings since the manifest entry is already gone. E6 holds: library file and approved root both survive | `deleting_a_linked_model_forgets_its_plan_but_never_the_library` |
| minor 6 | both reachability tests macOS-gated | added a non-gated stamp test in the existing test-only `image_jobs/tests.rs` (so the parity lane's `#[cfg(test)]` truncation hazard does not apply) | `the_linked_import_stamp_carries_the_plan_identity_and_no_install_path` |
| caveat | journal fields landed without a version bump | a DERIVED entry declares `RESOLVED_CACHE_DERIVED_STORE_VERSION = 2`; a source copy stays byte-identical at 1. `read_journal` probes the version before decoding, so a forward-version doc is a typed refusal rather than "corrupt" | `only_a_derived_journal_declares_the_widened_document_version` |

**On the version bump specifically:** bumping the store-wide `RESOLVED_CACHE_STORE_VERSION` was **rejected as genuinely incompatible**, not skipped. It would invalidate every entry already on disk, and retention cannot reclaim an entry it cannot read — a warm cache would be stranded as unreclaimable `RecoveryCandidate`s, permanently occupying disk and breaking size enforcement. Versioning the *document* instead gets the forward-compat statement with no such cost; the full existing suite (1175 lib tests + every integration target) confirms source-copy entries are unchanged.

**On major 1's end-to-end shape:** `prepare_image_route` is a private fn in the worker's backend-gated image lane, so the HTTP half and the routing half cannot be one test. They join on the checkpoint identity `linked/<rootId>/<relativePath>`, which both sides assert explicitly — the API test proves a route-approved root compiles to exactly that id, and the worker test proves that id routes to `CheckpointPlan` loading the linked file in place.

### Mutations

23 mutations run one at a time (`cp` backup + restore + `touch`, never `git checkout`), including all six lock guards individually and both retention lease guards separately and together. Full table in the PR body. Honest results recorded, including the two that came back GREEN by design (each retention guard alone) — those are why the lease-test comment was rewritten rather than left claiming coverage it does not have.

### Verification

`cargo fmt --all -- --check`; all three configs — `npm run rust:check`, `npm run rust:check:candle`, `npm run rust:check:neither`; `tests/test_builtin_manifest_audit.py` (95 passed). No fingerprinted memory-matrix source is touched, so no regeneration. Inference pin untouched.

### Outstanding

None.

Not auto-merged.

michaeltrefry:
Merged sc-20636 (which landed on the feature branch after this PR merged it) and resolved. Head is now `cb876fe61`, `MERGEABLE`, CI fully green (14/14 — `nax-worker` / `package-windows` are skipped by trigger).

My previous comment's DONE list still stands — all 4 major + 6 minor findings and the version caveat are resolved. This adds what the merge changed.

### Merge resolutions

Most of it was additive on both sides (store-lock constant beside the managed install dirs; linked lifecycle types beside the managed id helpers; `linked_checkpoint_selection` beside `ModelImportProvenance`/`normalize_model_import_source`; both stories' tests). Four were real:

1. **`queue_model_import_job`** — linked selection resolves first and now also refuses a discriminated `source`, since linked is an ownership mode rather than another way to fetch. A linked import names its own `linked-library` provenance and skips source normalisation; everything else normalises exactly as sc-20636 wrote it, and both manifest arms use `provenance.manifest_provider()`.

2. **`run_model_import_job` — I reverted one of my own sc-20635 changes.** I had suppressed `paths.model` for *any* plan-backed entry. sc-20636 then established that a MANAGED entry legitimately carries both the install path and `importPlan.checkpointId`, with its bespoke lane still serving the shapes the plan route does not — so my guard was wrong post-merge and reddened `managed_model_import_commits_atomically_and_stamps_the_plan_binding`. The stamp is unconditional again. A LINKED import never reaches that code (it is dispatched away at the top of the function) and `linked_checkpoint_manifest_entry` strips the path itself — re-mutating that strip still goes RED, so the linked guarantee is still load-bearing rather than incidentally true.

3. **E1 is now fully satisfied.** sc-20636 removed `plan_id` from `ImportPlanV1::semantic_form`, so `semantic_digest` is genuinely locator-independent and byte-identical checkpoints under two approved roots share ONE derivative entry instead of producing the same bytes twice. That is the open question this PR's body flagged for sc-20651; it is closed, and `the_derivative_key_covers_content_plan_adapter_codec_and_backend` moved from `assert_ne!` to `assert_eq!` with the reason recorded in place. The PR body's E1 section is rewritten accordingly.

4. **I changed one of sc-20636's tests** (`deleting_a_linked_model_leaves_its_checkpoint_record_untouched`) and want that called out explicitly. Its own doc comment said it guarded "the seam sc-20635 (linked plan/derivative invalidation) lands in" — and that seam has now landed, because review issue 9 requires `delete_model` to forget SceneWorks' own record/plan/bindings/derivatives. Its record-immortality assertion was the placeholder; its two real guards are kept and still asserted (managed teardown is never run for a linked entry, and the user's library file survives), plus a new one that the approved root survives. Renamed to `deleting_a_linked_model_never_runs_managed_teardown_or_touches_the_library` so the name matches what it checks. Re-mutating my `forget_linked_checkpoint` call now reds *both* that test and mine.

Also status-mapped sc-20636's four new `CheckpointPlanError` variants in `checkpoint_library.rs` rather than letting them fall into the catch-all 500 — unreachable through this linked-only surface today, but a managed route added later should not report a malformed install id as a server fault.

### Re-verified after the merge

Full `sceneworks-core` suite and every integration target; `cargo fmt --all -- --check`; all three configs (`rust:check`, `rust:check:candle`, `rust:check:neither`); `tests/test_builtin_manifest_audit.py` (95 passed). Three mutations re-run against the merged tree — the `delete_model` cleanup, the trim write-back, and the linked `paths.model` strip — all RED. Windows lane confirmed running the junction probe (`a_junction_out_of_the_library_fails_closed ... ok`).

### Outstanding

None.

Not auto-merged.

PR #2555, opened 2026-08-24
feat(checkpoint): Krea 2 full surface on the plan's verified layer (sc-20644)

michaeltrefry:
sc-20644 **Krea 2 row** — the first family row of the universal-route parity story. https://app.shortcut.com/trefry/story/20644

Also carries the shared seam every later row needs (pushed unreviewed on this branch before the pin bump landed): adapter-resolved backend eligibility / dialect source shape / operation + capability surface, the family branch removed from `resolve_checkpoint_plan_component`, shape-derived primary selection split from pinning, and declared components sourced from the checkpoint's own verified plan layers.

## What the row changes

Before it, a plan-backed Krea 2 entry had exactly ONE servable shape — plain text-to-image on the plan route. The bespoke lane declined every plan-backed entry outright, so:

* a **managed** import lost the LoRA / img2img / Kontext edit / strict-pose / multi-phase / Hires.fix surface it had a moment earlier;
* a **linked** checkpoint, which has no installed path to scan, could never reach any of those shapes at all.

The row separates the two halves of the migration: WHICH lane runs a request stays the per-request single-claim discriminator, and WHICH BYTES that lane opens becomes the plan. `checkpoint_plan_bespoke_primary_pin` hands the family lane the plan's own verified primary layer, so both routes feed the same generation body identical inputs and a fixed seed renders equal on either.

## Acceptance

* **AC1** — a linked and a managed copy of the same bytes compile to the same *semantic* digest; each lane loads its OWN copy's verified layer; the legacy `paths.model` route opens byte-identical weights on an identical seed sequence. Parity evidence is spec + seed + byte equality in-test (no pixel hash is committed); a real-weight fixed-seed render is the epic's terminal evidence, not this PR's.
* **AC2** — seven request shapes (t2i / img2img / LoRA / edit / pose / multi-phase / Hires) each have exactly one claiming lane AND resolve to the named route. The conformance test also pins the exact bespoke surface sc-20651 deletes — `is_single_file_checkpoint`, `is_diffusers_snapshot_dir`, `imported_dit_file`, the claim branch, the migration table — plus the ordering those three must keep. Missing / crossed / incompatible components still fail during PLANNING: a shape refusal is the one class that declines to another lane; a missing base tier, an unresolvable component, an unconsumed layer and an unbound adapter all still refuse before any loader.
* **AC3** — no model id, manifest key, or saved-workflow field changes.

Inherits epic requirements: E1, E2, E4, E5, E7, E8, E9.

## Behaviour change worth reviewing closely

Two sc-20636 assertions were updated rather than weakened: a **linked Krea edit is now served instead of refused**. That is the capability this row exists to restore, and `checkpoint_plan_route_refuses_before_load_with_typed_diagnostics` still proves the decline retains its typed refusal for the no-lane case via `into_unclaimed_refusal`.

## Gates

* `npm run rust:check` — green.
* 1825 `sceneworks-worker` lib tests pass, 0 failed.
* python authoring/manifest audit — 120 passed, 19 deselected.
* `rust:check:candle` / `rust:check:neither` — reported in a follow-up comment on this PR.
* Verified against inference pin `bdca0edf5` (the feature-branch tracking pin from #2547). No pin change here.

## Mutations, each run alone and observed red

| mutation | red |
|---|---|
| drop the plan-sourced primary in `resolve_imported_krea_dit_pin` | 4 tests |
| drop `checkpoint_plan_unconsumed_layers` from the pin helper | the lane loads a linked tree's backbone alone, substituting the base tier's VAE |
| drop `loras.is_empty()` from the route's shape gate | the LoRA row resolves to `CheckpointPlan` instead of `KreaImported` |
| disable the plan-backed decline | t2i double-claimed |
| empty `CHECKPOINT_PLAN_BESPOKE_PLAN_SOURCED_FAMILIES` | img2img double-claimed; a linked edit refuses |


michaeltrefry:
Gates, all three build configs, on head `b9ffcb3c2`:

- `npm run rust:check` (macOS / MLX) — green
- `node scripts/check-candle-build.mjs` — green ("the candle cfg typechecks, incl. test targets, under -D warnings")
- `node scripts/check-neither-build.mjs` — green (docker, 1m07s)
- `sceneworks-worker` lib tests — 1825 passed, 0 failed, 211 ignored
- python authoring/manifest audit (`pytest -q tests/ -m "not e2e and not parity"`) — 120 passed, 19 deselected

Verified against inference pin `bdca0edf5`; no pin change in this PR.

PR #2564, opened 2026-08-24
feat(checkpoint): name Wan's two experts and put Wan on the plan (sc-20644)

michaeltrefry:
sc-20644 **Wan row** — the seventh and last family. https://app.shortcut.com/trefry/story/20644

> **Stacked on Krea (#2555) → SDXL (#2559) → Mage-Flow (#2560) → ComfyUI trees (#2561).** Merge those first and this diff shrinks to its own commit `f5c993846`. Review only `f5c993846`.

Wan is the family that forced the checkpoint plan to grow a vocabulary, so this lands the vocabulary first and the row on top of it.

## 1. The expert role vocabulary — additive, and **not** a contract bump

Wan 2.2's ComfyUI checkpoint has **two** backbones: a high-noise and a low-noise expert, selected per denoise step, not interchangeable. `checkpoint_inspector` had no way to say so — its path-role inference maps both `unet/` and `diffusion_models/` onto the single role `transformer` — so a compiled Wan plan carried two anonymous `transformer` layers and every consumer could only refuse it as an ambiguous primary.

`TRANSFORMER_HIGH_ROLE` / `TRANSFORMER_LOW_ROLE` name them. The refinement fires only when **all three** hold:

1. the artifact's detected family is in `MULTI_EXPERT_FAMILIES` (today `wan-video` alone),
2. the reconciled role is already exactly `transformer`,
3. the file name carries a **delimited** high/low-noise marker (bounded on both sides, so `high_noise` matches and `highest_noise` does not; a name carrying *both* markers is refused rather than resolved to whichever comes first).

So a single-expert Wan checkpoint keeps the plain role, and **no other family reaches the marker test at all**, whatever its files are called.

### Why `CHECKPOINT_IMPORT_CONTRACT_VERSION` is not bumped — please check my reasoning

The change adds two role spellings that no previously-compiled plan can contain, and alters nothing about how an existing artifact is classified, serialized or verified. Every already-persisted plan reads back identically and re-compiles to the same semantic digest. The version exists to invalidate stored plans; this change cannot invalidate one. Evidence:

* the **entire shipped `sceneworks-core` suite passes unchanged** — 1177 unit tests plus 17 integration binaries, including the plan-store, ingest and import-contract goldens (the linked==managed digest, the rename-invariance digest, the round-trip contracts);
* plus two new tests covering the half goldens *cannot* cover — a hostile **file name** leaves other families alone, because none of the shipped fixtures is named that way.

If you read the additivity differently, this is the place to say so: it is far cheaper to bump now than after the terminal merge.

**Spelling.** Underscored, like every other layer role this module emits (`transformer`, `text_encoder`, `vae`) and like the `components[]` roles the catalog already uses for Wan. The adapter's `component_topology` in inference #787 spells the same two roles hyphenated, exactly as it spells `base-snapshot` for the underscored `base_snapshot` component id — pre-existing convention, not drift.

## 2. The row

* `checkpoint_plan_bespoke_roles` is the third and last shape a bespoke lane can want: resolution entirely **by named role**, no shape-derived primary, because Wan has no single backbone to derive. It takes the manifest **entry** rather than a request — the only thing it reads — so a *video* job can call it.
* **It deliberately does not consult the checkpoint adapter, and that is the design decision most worth your scrutiny.** The adapter supplies the dialect source shape and the eligible backends; a roles-only resolution needs neither (no primary whose shape must be decided, and the sole caller sits behind `cfg(backend-candle)`, so backend eligibility is a compile-time fact). Requiring one would add a cross-repo dependency for its own sake and would gate this PR on the inference pin for no correctness gain. The family truth it *does* need — which artifact is which expert — comes from the plan, the stronger authority: the inspector hashed those bytes and the store re-checked them. Once #787 is pinned, an adapter-eligibility assertion can be added cheaply if you want the belt as well as the braces.
* The lane reaches the plan source **before** the `usable` flag and the `components[]` assembly, both of which describe a scanned catalog row a linked checkpoint has neither of. `wan_comfyui_available` no longer rests on the `external_base_` id prefix alone — gating the claim on it left every imported Wan checkpoint with **no lane at all**, which is the capability this row restores.
* The in-place UMT5 encoder and Wan VAE stay **optional** on the plan route exactly as on the catalog path (sc-10909); anything the plan carries that neither list names refuses rather than going unloaded. The T2V-only channel gate is applied to the **plan's own** high expert, so a plan-backed I2V pair declines for the same reason a scanned one does rather than failing with a shape error at generate.

## Acceptance

* **AC1** — a compiled Wan plan carries two named expert layers, each resolving to its own file, and the lane loads both from the plan. Parity: **plan/route-level only** — the Wan lane is candle-only, this is an Apple Silicon Mac, and no Wan weights are present in `/Volumes/Models/huggingface/hub`, `~/SceneWorks/data` or `~/models`. No render was performed and none is claimed.
* **AC2** — plan source before every catalog gate, claim no longer resting on the id prefix, deletable surface pinned by name (`resolve_wan_comfyui_paths`, the `usable` flag, the `components[]` assembly, `WAN_COMFYUI_SNAPSHOT_TIERS`) with the plan-sourced resolver and snapshot probe pinned as having to *outlive* it. Missing components refuse during planning.
* **AC3** — no model id, manifest key, or saved-workflow field changes; the `external_base_*` scan path is untouched for entries that are not plan-backed.

Inherits epic requirements: E1, E2, E4, E5, E7, E8, E9.

## Gates

`npm run rust:check`, `check-candle-build.mjs`, `check-neither-build.mjs` all green; 1830 worker lib tests and the full `sceneworks-core` suite pass, 0 failed; python audit 120 passed. **Memory matrix regenerated** — a pure fingerprint refresh of `video_jobs/candle.rs` plus the aggregate tree hash, **no cell changed**, which is the expected shape for a routing change; `--check` re-run and idempotent. Verified against inference pin `bdca0edf5`; no pin change.

## Mutations, each run alone and observed red

| mutation | red |
|---|---|
| drop the family guard from `refine_multi_expert_role` | the Qwen fixture gets `transformer_high` |
| drop the both-markers-is-ambiguous arm from `expert_marker` | a merged Wan file silently becomes an expert |
| disable the refinement entirely | two anonymous `transformer` layers return |
| move the plan source behind the `usable` / `components[]` gates | the ordering assertion |
| drop the plan-backed disjunct from `wan_comfyui_available` | the claim assertion |


michaeltrefry:
## Review fix pass — the no-bump contract call, with the review's two caveats recorded

The adversarial review **agreed** with the decision not to bump `CHECKPOINT_IMPORT_CONTRACT_VERSION` for the additive expert-role vocabulary. Recording its two caveats here so they are on the record rather than only in the review:

1. **An old reader's refusal says `missing-component`, not "recompile" or "rescan".** Accurate — a pre-vocabulary binary reading a plan that carries `transformer_high` / `transformer_low` refuses by naming the component it cannot source, which is a true statement about what went wrong. But it is *unversioned*: the message does not tell the operator that the plan is newer than the reader, so the actionable next step (upgrade, or recompile the checkpoint) has to be inferred. That is a diagnostic-quality gap, not a correctness one, and it is the price of not bumping.

2. **The old-reader scenario is hypothetical pre-main.** Nothing outside this epic's feature branch has ever compiled a plan carrying these roles, so there is no deployed reader that can encounter one. The caveat becomes real only if the vocabulary ships to `main` and a rollback puts an older worker in front of a newer plan store.

Both are worth weighing against the cost of a bump, which would invalidate every plan already compiled across the six family rows.

## Everything else in the fix pass

Blockers 2 and 3, majors 5 and 8, minors 9 and 10 are all resolved in `8d74aee6e`; blocker 1 and major 4 in the four branches below this one. Per-issue detail, the failing mutation for each, and the two mutations that **passed first time and forced a test to be rewritten**, are in the commit messages.

Worth flagging one: the token-boundary mutation for the widened expert marker survived all three publisher fixtures, because every one of them carries a `wan2.2_t2v_` prefix and so never exercises a marker at the *start* of a name. Added a `HighNoise_Wan2_2_...` fixture; the mutation is red against it now. Found by running the mutation, not by reading.

PR #2572, opened 2026-08-24
chore(sc-21505): triage the memory-contract reconciliation into drift vs by-construction

michaeltrefry:
Closes [sc-21505](https://app.shortcut.com/trefry/story/21505).

The reconciliation reported **474 undifferentiated mismatches**, which reads as 474 things to fix. It is not: most are coordinates that *cannot* carry a declaration. This splits the enumeration by disposition and fixes the drift that is decidable today.

**474 → 435: 142 drift, 293 by construction, 0 unclassified.**

## AC 1 — classification with counts and a written rule

New `scripts/lib/memory-contract-triage.mjs`. Every finding gets a named class carrying a written rationale and a `by-construction` / `drift` disposition. The report prints the split and takes a `--drift` filter.

| | count | class |
|---|---|---|
| by-construction | 145 | no production route witnessed at that tier |
| **drift** | **78** | survey names a coordinate the engine dump does not reach |
| by-construction | 43 | engine provider has no image-manifest host |
| **drift** | **40** | manifest declares a (mode, overlay) the matrix has no cell for |
| by-construction | 37 | catalog entry does not advertise the tier |
| by-construction | 27 | survey withholds a loaded overlay the dump cannot exclude |
| by-construction | 26 | survey scope omits the axis, so the wildcard expands to every cell |
| **drift** | **18** | manifest declares a coordinate with no witnessed route |
| by-construction | 12 | MLX lane requires `requestContexts`, which no dump publishes |
| **drift** | **6** | survey scope narrower than the engine implements and routes |
| by-construction | 3 | declaration deliberately withheld by a recorded verdict |

Mechanics worth review:

- `collectMemoryContractMismatches` records a `cause` at each emit site, in a side map keyed by `reconciliationMismatchKey`. Deliberately **not** a twelfth field: the eleven AXES are the whole identity of a coordinate, and both the dedupe key and the survey leg's `JSON.parse(key)` round-trip depend on that.
- The `engine_manifest` leg joins against `planProjection` — whether the manifest *could* carry a declaration is a property of the projection, not of the leg.
- **The wildcard test is per axis.** A first cut asked only whether a scope omitted *some* axis, which excused contradictions on axes the verdict had named outright; 50 coordinates were filed "count of work: 0" on the strength of an unrelated `modes` omission.

## AC 2 — drift fixed

- The manifest projection was **stale**; refreshed.
- **15514 (FLUX.1), 15518 (FLUX.2 Klein), 15527 (PuLID)** recorded `implementation: "none"` while the dump at pin `ebcdc7da` declares `bounded_transformer_residency` implemented *and* witnesses routes. Their **candle siblings already read `shared-primitive`** — only the MLX verdicts were stale.
- Scopes are **clean-base only**, because that is what the providers require: `structurally_streamable` (flux), `klein_streamable` (flux2) and `surface_streamable` (mage) each demand an empty adapter/identity/overlay spec. **Routing is not rung capability**, so route witnesses cannot settle the overlay axis. PuLID keeps `identity` — `backbone_spec` strips it before delegating to FLUX.1's contract.
- `anima_*`: MLX rung 4 narrowed to **q4**, the only tier with matrix cells.
- `ideogram_4` / `_turbo`: rung 4 drops **`image_to_image`**, which has no rung-4 cell.
- `kolors:candle` gains a `memoryDeclarationWithhold` naming SC-20790. That decision lived in **prose**, so the projection could not see it and wrote a contract that republished the undelivered plain overlay as Implemented.

## AC 3 — report reflects the outcome

Both halves: reduced counts **and** the annotation mechanism. Regenerated through the standard flow; `generate-memory-matrix.mjs --check` exits 0 and leaves the goldens byte-identical (verified by hash, not by exit code). `memory-contract-reconciliation.mjs` is **not** added to `SOURCE_PATHS`, per the story's constraint.

## Left open deliberately

Filed as a scoped follow-up rather than guessed at:

- **78 survey-vs-dump contradictions.** Which side is wrong is not decidable from the dump. `qwen_image` is the standing counter-example: it carries rung 4 on its bf16 surface alone while checked-in calibration evidence verifies the q4 *and* q8 cells — there the **dump** under-reports. An earlier revision of this branch narrowed that verdict to match the dump and a test caught it; narrowing would have deleted measured coverage.
- **58 `manifest_route` rows**, whose measured `parameterRanges` do not distribute mechanically when a cross-product row is split.
- **6 underclaims**, including the illustrious entries SC-15525 records as a measured withhold.

## Verification

- `npm run check`: **734/734**.
- New suite `scripts/memory-contract-triage.test.mjs` (11 tests). Each new assertion mutation-checked: reverting the per-axis rule, deleting the `withheld` reason row, and stubbing the overlay predicate each go red and revert clean.
- On this Mac the gate needs a non-symlinked temp root (`/var` is a symlink, and the CUDA harness rejects a cache root that traverses one): `TMPDIR=... RUNNER_TEMP=... npm run check`. Unrelated to this change — reproduced identically at the branch point.


michaeltrefry:
The `macOS MLX worker` failure here (`MLX error: Failed to load the default metallib. library not found`) was not this branch's change — it was a latent bug in the sc-21382 prebuilt-libmlx seam that fired on the first warm-cache run: the prebuilt's baked METAL_PATH is the publisher's build dir, and the `~/.cache/pmetal/lib` fallback only refreshes when the mlx-sys build script re-runs, which a warm cargo cache skips. Fixed on `main` by #2574 (`scripts/fetch-prebuilt-mlx.sh` now exports `PMETAL_METALLIB_PATH`). Merge `main` into this branch and re-run; the lane should go green. — claude

PR #2729, opened 2026-09-05
feat(sc-22729): the SDXL family measurable on both lanes, keyed on (provider, modelId), plus the bespoke InstantID arm

michaeltrefry:
Story: **sc-22729** — S6 SDXL family arm (epic 22723). Targets `feature/sc-22723-memory-anchor-measurability`.

## What was wrong

The worker routes **five** catalog models onto the single `sdxl` engine id on **both** lanes
(`crates/sceneworks-worker/src/engines.rs` `MODEL_TABLE`; `crates/sceneworks-core/src/jobs_store/routing/candle.rs:1330`),
and `instantid_realvisxl` onto the bespoke `instantid` crates. The measurement apparatus knew only
base SDXL, on MLX only.

The engine id is **not** an artifact identity. `candle-gen-sdxl` seals a per-route
repository/revision and mints a per-route calibration fingerprint, and each catalog model ships its
own independently pinned rehost — so an anchor bound to the engine's env family would re-label base
SDXL's peaks as a finetune's. The arm is therefore keyed on **`(provider, modelId)`** on both lanes.

## Scope per AC bullet

**AC1 — “each arm loads through the engine's production loader and returns a probe-consistent record; a root of another tier is refused naming the tier.”**

- `mlx.rs`: `SdxlArm` table with five members resolved from `planned.target.modelId`; each carries
  its own `SCENEWORKS_<MODEL>_{REPOSITORY,REVISION,ROOT}` triple, execution path, calibration label,
  diagnostics slug and fixture stem. `validate_huggingface_snapshot_root` refuses a q4 export under
  a q8 plan naming the tier, and `sdxl_spec_at` re-asserts the planned tier's quant. Loads through
  `registry.load("sdxl", &spec)` — the registered generator.
- The `LoadSpec` mirrors the **worker's** MLX still-image shape, not a shape chosen to please the
  gate: `with_resolved_route(request.model)` (`image_jobs/base.rs:8785`), `DeferredMaterialization`
  from the registry's legacy shaper (the MLX `sdxl` rules are `legacy_shaping: true`), and the
  tier's advisory quant (`mlx_load_quant_for_resolved_artifact` suppresses it only for the
  krea/klein turnkeys). The offload policy is the one `mlx_fit_gate::apply_residency_policy` would
  settle on for the selected rung — the worker's MLX branch sets none.
- `candle.rs`: the same family table on the five-rung reference path, plus the three things
  `SdxlArtifactSeal::capture` hard-requires — `resolved_route` (absent ⇒ `"optimized memory
  admission requires an exact resolved catalog route"`), the packed tier's quant (`spec.quantize ==
  tier.quant`), and the three caller-staged SDXL components at their exact upstream revisions.
- **InstantID** (both lanes): `candle-gen-catalog` / `mlx-gen-catalog` register no descriptor for
  it, so there is no `catalog.media().load` to call. Both arms drive
  `InstantId::load_with_memory_context` — the exact call `image_jobs/instantid.rs` makes — and the
  MLX arm attaches the SCRFD/ArcFace stack through `with_face_paths`, as production does.

**AC2 — “`--list` on each backend shows every table cell as `runnable` or `weights_missing`.”**
Verified on both backends; all 25 planned cells classify `weights_missing` on this host (it holds no
SDXL snapshots), which is the measurable status.

## Identity at the pin — every planned fingerprint is one the production contract emits

| lane | source | string |
| --- | --- | --- |
| mlx, all five SDXL routes | `mlx-gen-sdxl/src/memory_strategy.rs:583` (route-independent, unconditional) | `sdxl-mlx-unet-shared-ladder-v3` |
| candle, per route | `candle-gen-sdxl/src/memory_strategy.rs:128` `route_fingerprint` | `sdxl-candle-<route tokens>-staged-decode-attention-v1` |
| mlx instantid | `mlx-gen-instantid/src/memory_strategy.rs:16,88` (it publishes its request-contract revision AS the fingerprint) | `instantid-request-contract-v1` |
| candle instantid | `candle-gen-instantid/src/memory_strategy.rs:26,222` | `instantid-candle-staged-conditioning-v1` |

Both InstantID crates hard-code `LoadShape::EagerMaterialization` on the identity, so those plan
rows declare the eager shape. `every_planned_sdxl_row_cites_an_emittable_fingerprint` (candle) and
`planned_fingerprints_are_the_pinned_crates_own_strings` (mlx) walk the **committed plan** and
assert every row's fingerprint, load shape and mode against what the crate can emit — including the
`v1 → rev1` rewrite the Illustrious route ids need. No registry-conformance label anywhere.

## Table cells

| model | mlx | candle |
| --- | --- | --- |
| `sdxl` | already planned (3) | **+3** |
| `realvisxl` | **+3** | **+3** |
| `realvisxl_lightning` | **+3** | **+3** |
| `illustrious_xl_v1` | **+3** | **+3** (declared; capture-blocked by engine route drift) |
| `illustrious_xl_v2` | **+3** | **+3** (declared; capture-blocked by engine route drift) |
| `instantid_realvisxl` | **+3** | **+1** (bf16; q4/q8 are not cells) |

**31 new anchors — every one of the family's 34 cells is DECLARED** (a plan anchor plus both loader-closure
declarations on every routed lane; `sdxl:mlx` was already planned). Two engine-side facts keep some of them from
being *capturable* today. An engine defect blocks CAPTURE, never DECLARATION (epic 22723 E1), so both are reported
rather than hidden — and both are DERIVED, so neither can outlive the condition that justifies it:

- **`illustrious_xl_v{1,2}` on candle — all three tiers, 6 cells, all planned and both closures declared.**
  `candle-gen-sdxl`'s `SDXL_ROUTES` pins those routes at revisions this repository no longer ships.
  `path_has_snapshot` matches that literal before `SdxlArtifactSeal::capture` will seal, and `load`
  propagates the failure — so **no root the manifest can resolve can load these two on candle**, and
  `--list` reports them `harness_unsupported` for the engine's own reason rather than as missing plan
  or adapter work. **Nothing is hard-coded:** `sdxlCandleRouteDrift` READS the engine revision out of
  the pinned inference checkout (`SDXL_ROUTES`) and the shipped one out of the manifest, and compares
  them. Equal revisions clear the refusal with no edit here; with no inference checkout to read the
  cells classify normally and say the comparison did not happen. **The inference-side fix is
  [inference PR 946](https://github.com/SceneWorks/inference/pull/946)** (`story/sc-22729-epic-22723-memory-anchor-measurability`),
  merging separately — when it lands and the pin moves, these six become capturable with no change to this repo.
- **`instantid_realvisxl` q4/q8 on candle (2 cells)** are not cells at all: the candle InstantID
  stack is dense-only and always loads `bf16/` (`image_jobs/instantid.rs:234-253`). The gap set narrows
  the tier axis by `parseBackendTierOverrides` — the SAME worker-source derivation the matrix generator
  uses — and by nothing else, so it stops demanding an anchor that could only ever measure bf16 weights
  and file the peaks under a packed tier, without deleting cells whose only defect is an unrecorded peak.

## Closures

11 loader declarations (`engineId` aliases wherever the inference tree never spells the catalog id)
and 3 crate closures (`candle:sdxl`, `mlx:instantid`, `candle:instantid`), each including the
engine's `memory_strategy.rs` so an identity change stales the anchor. Derived with
`--repo <pinned clone> --write`; `--check`, `anchor-loader-closure --check` and `--stamp-anchors
--check` all pass. **Every pre-existing digest is byte-identical** (verified structurally, not by
eyeballing the diff).

## Tests

- `lib.rs`: the InstantID bundle refuses naming every missing file; **cross-source** test reads
  `image_jobs/instantid.rs` so a filename or env rename in the worker reds this crate.
- `mlx.rs`: arm resolved from the model id + every field collision-free; the spec binds route, tier,
  load shape and offload policy; the fixture stem binds the family member (base SDXL's fixture is
  refused for a `realvisxl` plan); InstantID target/overlay/mode/tier guards; ordered artifact
  digest; the plan-row fingerprint walk.
- `candle.rs`: the same arm-resolution and collision tests, the five exact route fingerprints
  (including `realvisxl` vs `realvisxl_lightning`, the pair a prefix comparison confuses), the
  plan-row walk, the dense-only refusal, and the artifact digest.
- `measure-memory-catalog.test.mjs`: **“the sdxl family is measurable on every shipped tier of every
  routed lane”** — asserts the exclusions are exactly the two engine-side ones and that they are
  still *reported*, so the exclusion cannot silently widen. Plus a cross-source test tying the
  catalog's `SDXL_COMPONENTS` env list to `candle.rs`'s. No frozen counts anywhere — key sets only.

**Status (after the review fix pass):** `cargo test -p sceneworks-memory-adapter --features mlx` green
(164 bin + 34 lib). `npm run check` green (748 tests, 0 failures).
`npm run rust:check:candle` green. `npm run rust:fmt --check` clean.
`npm run rust:check` fails **only** on 11 `sceneworks-rust-api` `tests::training::*` cases, all with
`"Not enough free disk space to train: 0.8 GiB available … at least 2.0 GiB is recommended"` — the
dev Mac's data volume is at 100% (2.1 GiB free of 3.6 TiB). Environmental, untouched by this change;
725 tests in that crate pass and 24 other crates are green.

**Gap set: 280 → 207.**

## Also fixed in passing

`adapterCapturableProviders` intersected *every* dispatch gate, which vetoed a **bespoke** arm —
one dispatched by an early `if provider == <ID>` before those gates — so the stale-lane report
called `candle:instantid` (this story) and `candle:ltx_2_5_distilled` (sc-22725) *uncapturable* and
told an operator a booked capture host would be wasted. Bespoke ids are now unioned in, with a test
covering both the union and the loud refusal of an unresolvable bespoke id.

## Review fix pass (2026-09-05)

Merged `feature/sc-22723-memory-anchor-measurability` (through sc-22726/22727/22728) and resolved the five
review findings.

**[blocker] the six illustrious candle cells are DECLARED.** `illustrious_xl_v{1,2}:{q4,q8,bf16}:candle`
plan anchors (`provider: "sdxl"`, `loadShape: "deferred_materialization"`, fingerprints
`sdxl-candle-illustrious-xl-rev{1,2}-staged-decode-attention-v1`) and the `illustrious_xl_v{1,2}:candle`
loader closures. `anchor-loader-closure --write` / `--check` / `--stamp-anchors --check` all pass against the
pinned clone. The cells classify with the engine's own drift reason today, and become `runnable`/`weights_missing`
the moment the engine agrees with the manifest.

**[major] the burndown DENOMINATOR is narrowed only by routing facts.** The tier-axis intersection used the
matrix's `axes.<backend>.tiers`, whose `tiersFor` falls back to the manifest's `vramGbByTier` — a MEASUREMENT
declaration. That deleted six real cells (`flux_dev`/`flux_schnell`/`flux2_dev` candle bf16,
`sd3_5_large`/`_turbo`/`sd3_5_medium` candle q8) whose only defect was carrying no recorded peak: exactly what
the gap set exists to count. It now intersects against `parseBackendTierOverrides` (exported from
`generate-memory-matrix.mjs`), every drop asserts the worker-source override that justifies it, and a new case
asserts those six are still in the denominator *and* still exercise the hazard. Measured: 207 gaps with the fix,
204 with the manifest intersection restored — three live gaps were being hidden.

**[major] the illustrious drift exclusion is derived, not written down.** `parseSdxlRoutes` reads
`SDXL_ROUTES` out of `$INFERENCE_REPO/crates/media/candle-gen/candle-gen-sdxl/src/memory_strategy.rs`;
`sdxlCandleRouteDrift` compares that revision with the manifest's. Lane-unsupported only when they differ;
an equal pair clears it; a route the engine does not declare is refused for *that* reason; no checkout at all
is not a refusal — the cell classifies normally and carries `SDXL_ROUTES_UNCHECKED` in its reason. Tests cover
all four, including that the values at the pin really do differ today.

**[major] the VRAM-probe guard is a SHAPE claim.** `platform-review-contracts.test.mjs` no longer counts
`certifying_vram_probe()` occurrences. It maps every probe construction to its enclosing `fn`, requires every
capture arm that samples `vram.phase()` to own a certifying probe or be handed one, and allowlists exactly one
documented non-certifying site (`run_ltx25_capture`'s `start_rendered().assert_idle(1.0)`).

**[minor]** `SDXL_CANDLE_ROUTE_DRIFT` derives all six illustrious candle keys instead of listing two;
`candle.rs`'s plan guard asserts the exact `(modelId, lane)` set instead of `checked >= 6` (a floor stayed green
with a whole member's lane deleted from the plan); a new `candle.rs` case drives `instantid_candle_identity`
**past** the tier guard on a synthetic staged tree (`StagedInstantId` in the adapter's `lib.rs`) — the missing
IdentityNet weight is named, and the artifact digest is proved to cover exactly the five ordered paths (a
permutation and every four-path subset both differ); the `laneUnsupported` lookup that motivated the `familyFor`
note is gone entirely, and `familyFor` now composes `providerFamily` so sc-22727's per-modelId `variants` still
resolve.

**Mutations run (each assertion, the mutation that reds it):**

| --- | --- | --- |
| gap-set denominator | restore the `axes[backend].tiers` intersection | RED (3 tests) |
| illustrious refusal is derived | `sdxlCandleRouteDrift` always returns `null` | RED (2 tests) |
| unreadable checkout refuses nothing | `readSdxlCandleRoutes` rethrows ENOENT | RED |
| every family cell is planned + declared | delete the `illustrious_xl_v1:candle` closure | RED |
| the sdxl family is measurable | delete the `illustrious_xl_v2:q8:candle` anchor | RED (2 tests) |
| every probe is certifying | `run_qwen_edit` mints `start_rendered().assert_idle(1.0)` | RED |
| the no-arm probe still asks its question | give `minimax_h3` a candle arm | RED (2 tests) |

The two Rust cases in `candle.rs` (`instantid_candle_identity_names_the_missing_identitynet_weight` and the exact
plan key set) are **typecheck-only on this host**: `cargo test --features candle` links and needs real libcuda, so
`scripts/check-candle-build.mjs` typechecks them here and the windows-candle CI lane is where they first RUN.


michaeltrefry:
Force-pushed a fix for the red `macOS build, lint and workspace tests (hosted)` lane on the first push. Two real defects, both mine, both now green locally under the exact CI commands:

1. **`cargo clippy -p sceneworks-memory-adapter --features mlx --all-targets -- -D warnings`** — two `selection.clone()` calls on `MemorySelection`, which is `Copy`. Removed on both adapters.
2. **`mlx_fit_gate::tests::every_planned_mlx_lane_resolves_a_weights_free_provider_contract`** — the guard walks every planned MLX lane and resolves a weights-free **registry** contract, but `instantid` is a bespoke crate that registers no descriptor and no memory-strategy registration at all (`mlx_gen_catalog::BESPOKE_UTILITY_CRATES`), so there is nothing for it to resolve. It now takes a named bespoke exit that **asserts** the provider really is absent from the runtime registry, rather than skipping silently — a genuinely registered provider can never take that exit and lose its coverage.

Re-verified: `cargo clippy --all-targets -- -D warnings` clean; `cargo clippy -p sceneworks-memory-adapter --features mlx --all-targets -- -D warnings` clean; `cargo test -p sceneworks-worker --lib` 2073 passed / 0 failed; MLX adapter 138 bin + 33 lib green; `npm run check` green; `npm run rust:check:candle` green; `cargo fmt --all --check` clean.

Note on the first push: my local `npm run rust:check` had aborted early in `sceneworks-rust-api` on an environmental failure (the dev Mac's data volume is at 100%, and 11 `tests::training::*` cases assert a 2.0 GiB free-disk floor), so the run never reached `sceneworks-worker` or the adapter clippy step. I should have driven it to completion before opening the PR.

michaeltrefry:
Second force-push: `parity-scaffold` was red on the matrix fingerprint only.

`crates/sceneworks-worker/src/mlx_fit_gate.rs` is one of the eight sources `docs/generated/memory-matrix.json` fingerprints, so the guard fix in the previous push moved `sceneWorksRevision` and left the checked-in derived docs stale. Regenerated with `npm run generate:memory-matrix` — fingerprint only, no cell moved.

Reproduced and verified locally under the exact parity-scaffold command (`INFERENCE_REPO=<pinned clone> npm run check`), which is what I had missed: my earlier local `npm run check` ran without `INFERENCE_REPO`, and I had regenerated the matrix *before* the last source edit rather than after it. The regen has to be the final step before committing whenever a fingerprinted source moves.

PR #2745, opened 2026-09-06
fix(sc-22738): the catalog campaign stops spending captures on the LTX-2.3 MLX cells it cannot drive

michaeltrefry:
## Diagnosis

The live MLX catalog campaign (SceneWorks `1cd23a0e6`, inference pin `563c44e16`) failed
`ltx_2_3:{bf16,q4,q8}:mlx` with the adapter exiting before any model load:

```
memory-strategy provider adapter: SC-18946 row is missing required _measurementSafety; refusing before model load
```

**The plan row cannot carry that block, and supplying it would not help.**

1. `packages/schemas/memory-anchor-plan.schema.json` sets `additionalProperties: false` on
   `$defs/anchor`. A `_measurementSafety` key on a plan anchor is not merely un-derived — it is
   *inexpressible*. The blocks that do exist live in `docs/calibration/sc-18946/*.json` and are
   synthesised at request time by `scripts/run-ltx-safety-canary.mjs`, not by the plan.
2. It is only the **first field** the refusal reads.
   `scripts/memory-calibration-harness.mjs` sends exactly one provider action for a capture
   (`action: "run"`). `run_ltx_with_admission` routes that to `LtxRunAdmission::Ordinary`
   (`crates/sceneworks-memory-adapter/src/bin/mlx.rs:15158`), whose entire body is
   `refuse_unsafe_ltx_capture` (`mlx.rs:12715`) — `validate_ltx_safety_evidence` (`mlx.rs:12633`,
   where the quoted message lives) followed by an unconditional `Err`. There is **no `Ok` return on
   any geometry or tier.** A row that satisfied every field of the block would fail one line later
   with `SC-19642 pre-load safety refusal` instead.

**The refusal is the designed state, not a defect.** At the anchor's own geometry
(768×512, f121) the arm's incident-calibrated projection for the lane-default composition is
`97_906_593_920` B — above the `96_970_084_480` B physical footprint the q4 f305 incident reached
before the host watchdog panicked (`LTX_Q4_F305_CRASH_FOOTPRINT_BYTES`). The identical number is
recorded for this exact cell in `docs/calibration/sc-18946/ltx-mlx-single-pass-sweep.json`
(`mlx-ltx-2-3-q4-768x512-f121-fps30-staged_residency` → `disposition: safety_refused_open`). And
the generic runner offers no containment: the footprint watchdog is an external,
nonce-authenticated unix socket that only `scripts/run-ltx-safety-canary.mjs` provides
(`mlx.rs:14702`).

## What actually measures these cells

`scripts/run-ltx-safety-canary.mjs`, through the private `bounded_campaign_entry` action, per tier
— `BOUNDED_CAMPAIGN_ENTRY_SPECS` declares q4 (sc-20318), q8 and bf16 (sc-20430). Same geometry,
**bounded decode 192/64**, which moves the projection to ≈84.7 GB, run inside a
`53_347_146_863` B footprint ceiling on its own contained volume
(`/Volumes/Data/sceneworks-safety-canary`). That is a host-exclusive night run, never part of a
catalog sweep.

## Derivation of the values (why no numbers were invented)

Nothing new is asserted. Every constant quoted above is read from a checked-in source
(`mlx.rs` constants, the sc-18946 sweep record, the canary driver's own specs). **No
`_measurementSafety` block was written into the plan**, because the correct value of that block for
these rows is the one already computed *inside the adapter* from `LTX_{Q4,Q8,BF16}_INVENTORY_BYTES`
and the pinned provider decode cost — and its correct outcome is a refusal.

## The change

`--list` classifies the three cells `harness_unsupported` — the status that already exists for
"the adapter arm is there, the harness cannot drive it" — with the engine-side reason, so the
campaign skips them instead of spending three failed captures per sweep. `CANARY_DRIVEN_CELLS` is
**derived** from the driver's own `BOUNDED_CAMPAIGN_ENTRY_SPECS`, never hand-kept. The gate
`every shipped tiered model is measurable` accepts these cells only through that exact refusal; any
other status for the same key is still counted a gap.

## Tests

New: `the canary-driven LTX-2.3 MLX cells are refused for the adapter's own reason, not spent on`.
It re-derives the delegation from both ends every run — the harness's single capture action, the
adapter's `Ordinary → refuse_unsafe_ltx_capture` dispatch, that function having no `Ok` exit, and
the plan/driver tier sets agreeing in both directions — and asserts the Candle sibling of the same
family is untouched.

Mutations, one guard at a time, `touch` between runs:

| --- | --- | --- |
| classification delegates the cell | `if (planned.provider === "ltx_2_3" && …)` → `if (false)` | **RED** |
| the arm's ordinary run cannot succeed | `refuse_unsafe_ltx_capture` gains `if geometry.frames == 7 { return Ok(()); }` | **RED** |
| the driver declares every planned tier | delete the `bf16` entry from `BOUNDED_CAMPAIGN_ENTRY_SPECS` | **RED** |

`node --test scripts/measure-memory-catalog.test.mjs scripts/anchor-loader-closure.test.mjs
scripts/memory-calibration-harness.test.mjs` → 0 (155 pass, 2 skip).
`npm run check` → 0. `check:memory-anchors` → 0, `check:memory-matrix` → 0 (the plan file is
unchanged, so no fingerprint moved). No Rust was modified, so no cargo run was needed; the MLX GPU
was never touched.

## Verification still owed

Real-weight verification is a **campaign re-run of the three `ltx_2_3:*:mlx` anchors**, which will
now report `harness_unsupported` and skip rather than fail. The *capture* of those cells remains
owed to a host-exclusive `run-ltx-safety-canary.mjs --profile bounded-campaign-entry{,-q8,-bf16}`
night run; `config/memory-anchors.json` today carries only the historical sc-18808 q8 MLX anchor
for this model.

### DIFF
--- a/docs/calibration-runbook.md
+++ b/docs/calibration-runbook.md
@@ -821,6 +821,17 @@ SCENEWORKS_FLUX2_REVISION=<exact artifact revision>
 SCENEWORKS_FLUX2_ROOT=/abs/path/.../snapshots/<rev>/<tier>   # q4 | q8 — tier DERIVED from the plan target
 
 # memory-mlx-adapter — ltx_2_3   (sc-18808; the only VIDEO arm. FOUR vars, not three)
+# 🔴 The catalog campaign CANNOT capture ltx_2_3:*:mlx (sc-22738). The harness's only action is
+# `run`, the arm routes `run` to `LtxRunAdmission::Ordinary`, and that admission is
+# `refuse_unsafe_ltx_capture` — an SC-19642 pre-load refusal with no Ok return on any geometry
+# (at 768x512xf121 its own incident-calibrated projection is 97.9 GB, above the 96.97 GB the q4
+# f305 incident reached before the host watchdog panicked). No plan row can satisfy it: the
+# anchor-plan schema is additionalProperties:false, so it cannot carry `_measurementSafety`, and
+# that block is only the refusal's first field. These three cells are measured by
+# `scripts/run-ltx-safety-canary.mjs --profile bounded-campaign-entry{,-q8,-bf16}` — same
+# geometry, bounded decode 192/64, inside the external 53.35 GB footprint watchdog on its own
+# contained volume — which is a host-exclusive night run, never part of a catalog sweep.
+# `measure-memory-catalog.mjs --list` reports them `harness_unsupported` for exactly that reason.
 SCENEWORKS_LTX_REPOSITORY=SceneWorks/ltx-2.3-mlx             # fixed; validated against LTX_REPOSITORY
 SCENEWORKS_LTX_REVISION=<exact artifact revision>
 SCENEWORKS_LTX_ROOT=/abs/path/.../snapshots/<rev>/<tier>     # bf16 | q4 | q8, derived from the plan target
--- a/scripts/measure-memory-catalog.mjs
+++ b/scripts/measure-memory-catalog.mjs
@@ -32,6 +32,7 @@ import { readFile, writeFile, mkdir, cp, rm, realpath, stat, readdir } from "nod
 
 import { stripJsoncComments } from "./lib/jsonc.mjs";
 import { hashArtifactInventory } from "./hash-artifact-inventory.mjs";
+import { BOUNDED_CAMPAIGN_ENTRY_SPECS } from "./run-ltx-safety-canary.mjs";
 
 export const ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), "..");
 export const PLAN_PATH = "config/memory-calibration-plan.json";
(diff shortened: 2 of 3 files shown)


michaeltrefry:
Note on the title: the brief asked for `fix(sc-22738): ltx_2_3 MLX plan rows carry the SC-18946 measurement-safety block the arm requires`. That title states something the investigation disproved — the plan schema is `additionalProperties: false` so no anchor row can carry `_measurementSafety`, and the block is only the first field of a refusal that has no success path. The title was changed to describe what the change actually does; the original wording is recorded here.

michaeltrefry:
Superseded: Michael ruled the SC-19642 unconditional pre-load refusal is a bug, not a safety check. A follow-up PR removes the guard and routes ltx_2_3 MLX captures through the production admission path with the shipped default composition; this reclassification to canary-driven is no longer the direction.

PR #2841, opened 2026-09-14
sc-23404: speak a pack's dialogue lines through the audio job route (epic sc-23401)

michaeltrefry:
## sc-23404 — Phase 2 S3: speech dialogue clips synthesized through the audio job route

Epic sc-23401 (film harness phase 2). Inherits epic requirements **E3, E5**.

A `dialogue` pack entry may now carry **`text`** (plus optional `voice` and `model`) instead of, or
as well as, `file`. During `ensure_sound` — before any render — the run speaks each *placed* line
through the ordinary `POST /api/v1/audio/jobs` route, writes the returned WAV into the pack
directory, and imports it on the dialogue bus **exactly as it imports a pre-recorded clip**.
Pre-recorded entries are unchanged: an entry that carries no `text` serializes back without any of
the new keys.

### What changed

**`crates/sceneworks-core/src/film_plan.rs`**
- `SoundEntry.file` becomes optional; `text` / `voice` / `model` added, with
  `is_synthesized()` / `synthesis_text()` / `synthesis_model()` / `synthesis_voice()`.
- `SOUND_SYNTHESIS_MODELS` = `kokoro_82m` (default), `chatterbox_tts`, `moss_tts_realtime`,
  `moss_ttsd_v05`. `MAX_DIALOGUE_TEXT_CHARS` = 1000; `voice` bounded to 64 chars but **not**
  allow-listed — the per-model voice bank is the generator's, and an unknown voice is a typed
  refusal at the gen-core floor.
- Validation, every finding naming the entry **by role**: neither `text` nor `file`; `text` on a
  non-`dialogue` kind; blank/overlong `text`; unknown model; `voice`/`model` with no `text`.
- `validate_reference_pack_files` exempts a synthesized entry from "the file must exist" — the run
  is what produces it. A *recorded* entry whose file is missing is still a finding.
- `RunRecord.synthesizedSound[]`: role, text, `textSha256`, model, voice, attempt, idempotency key,
  job id, status, the `type: audio` asset the job produced, the pack file it was written to, error,
  timestamps.

**`apps/rust-api/src/film_harness.rs`**
- `Session::synthesize_dialogue` — adopt / continue / new-attempt, dispatch, poll under the plan's
  own limits, copy the WAV into the pack, record it.
- `dialogue_idempotency_key(run_id, role, model, voice, text, attempt)`, stamped into
  `advanced.filmHarness.idempotencyKey` and found by the existing `find_job_by_idempotency_key`.
  Keyed on content **and** attempt: content so a re-cast line is not adopted from the job that
  spoke the old one, attempt so a retry after a failure is a new job rather than a re-read of the
  same failure forever.
- `synthesized_sound_file(role, text_sha256)` -> `sound/<role>.tts-<sha12>.wav` (gitignored — it is
  an output). An entry that pins `file` is synthesized *into* that path.
- **Limits**: one synthesis is bounded by `limits.maxShotSeconds` (the per-job budget the export
  already runs under) and the run by `limits.maxRunSeconds`. No new knob. `maxAttemptsPerShot`
  deliberately does not apply — it caps GPU renders, and the fixture's cap of 1 would make
  "resumable" untrue; at most one new attempt per line per invocation, so a retry loop is the
  operator's to run.
- **Stops, all resumable, all before the first render**: `dialogue_synthesis_refused` (enqueue
  refused), `dialogue_synthesis_failed` (failed / per-job budget / assets unsettled),
  `no_audio_worker`, plus the run's own `canceled` / `run_budget`. `drive_inner` returns before
  `work_shots`, so the clips already spoken and imported stay exactly where they are.
- **New preflight**: no live worker advertising `audio_generate` ⇒ `no_audio_worker` naming the
  lines, before the job exists — the same posture as the `video_generate` / `image_vqa` preflights,
  scoped to the lines still *owed*.

**Fixture** — the phase-1 tones were placeholders and the "with-dialogue" export had no speech:
- `references.jsonc` v2: the three dialogue roles carry `text` and a Kokoro voice —
  `courier_line` (`am_michael`), `recipient_line` (`af_heart`), and a new
  `recipient_reveal_line` (`af_heart`) so SH060 speaks its own line instead of repeating SH050's.
  Each line is the one that shot's own `dialogue` intent already stated.
- `plan.jsonc` / `plan.ltx25.jsonc`: SH060 moves to `recipient_reveal_line`, its intent updated to
  match.
- `sound/courier_line.wav` and `sound/recipient_line.wav` deleted; `FIXTURE_SOUNDS` is the two beds
  only.

**Merged from the phase-1 head (`c4a8556cc`, sync PR #2840).** Rebuilt and re-tested on top of it.
Two of its changes touch this path and both hold:
- #2838 derives the dialogue items from the merged picture `order` rather than `selected_takes()`.
  A synthesized clip flows through that unchanged — the derivation looks each shot's role up in
  `sound_assets`, which is the map `ensure_sound` fills, and synthesis fills it for exactly the
  roles the *selection* places (so a `--shots` run never pays GPU for a line it will not use).
- #2838 also added the `ensure_sound()` call in `replace_take` for the evaluation's finding #2 —
  which I had written independently; the merge conflict was resolved onto their comment. What is
  mine on that path is the guard beside it: if a clip cannot be re-hydrated at all (file gone, a
  re-synthesis that failed, no live TTS worker), the replacement records its take and leaves the
  saved timeline alone rather than re-assembling from a short clip map, which is the deletion
  `ensure_sound` exists to prevent.

**Test isolation (needed by this change).** `Harness::fixture_pack()` now returns a per-harness
*copy* of the shipped pack and its media instead of the checked-in path. A run writes its
synthesized clips beside the pack, so the old behaviour had `cargo test` writing into
`config/film-harness/courier-workshop/sound/` and racing other tests on one filename — observed
during development (a stray `courier_line.tts-*.wav` appeared in the checkout). The document copied
is byte-for-byte the shipped one, so `reference_pack.sha256` is unchanged; only its directory moves,
and `checked_in_fixture_sound_matches_the_generator_byte_for_byte` still reads the shipped path.

### GPU speech command

Run **after** this PR is up; serial GPU. The macOS worker builds the candle audio lane
unconditionally (`audio-metal`), so the one `SCENEWORKS_GPU_ID=mlx` worker the smoke starts
advertises `video_generate` **and** `audio_generate` — no second process is needed.

```sh
# Any checkout of this branch. Note the dev Mac's data volume was at 100% capacity (~22 GiB free)
# when this PR went up — `rm -rf target/` in the finished agent worktrees before the release build.
git checkout story/sc-23404-epic-23401-film-harness-phase-2
eval "$(scripts/fetch-prebuilt-mlx.sh --build-type Release)"
export PMETAL_MLX_PREBUILT_DIR PMETAL_METALLIB_PATH
SHOTS=SH010,SH020,SH050,SH060 \
SCENEWORKS_SMOKE_DIR=~/SceneWorks/film-harness-evidence/sc-23404/speech-export \
  scripts/film-harness-smoke.sh
```

Then probe the export for the lines:

```sh
EV=~/SceneWorks/film-harness-evidence/sc-23404/speech-export
scripts/film-harness-speech-probe.py --run "$EV/run/run.json" --out "$EV/speech-probe.json"
echo "probe exit: $?"
```

(`--export` is optional: the probe resolves the MP4 from the record's `projectPath` +
`export.renderPath`, because `renderPath` is project-*relative* and passing it alone is the obvious
mistake.)

**Why those four shots.** `SH010,SH020,SH050,SH060` is the smallest selection that places all three
lines (SH020 courier, SH050 recipient, SH060 reveal). `SHOTS=SH010,...,SH060` renders the whole
film and speaks the same three at 1.5x the GPU cost.

**Expected duration.** MiniMax-H3 q4 measured 760 s and 700 s for SH010/SH020 in sc-22715, so four
shots is roughly **50–75 min** of render (six would be **1.5–2.5 h**), plus three Kokoro synthesis
jobs of a few seconds each and ~15 s of export. The fixture's `maxRunSeconds` is 7200, which
comfortably covers four and is tight for six.

**Expected outputs**, under `$SCENEWORKS_SMOKE_DIR`:
- `run/run.json` — `outcome: "completed"`; `synthesizedSound` with **3 entries**, each
  `status: "completed"`, `attempt: 1`, `model: "kokoro_82m"`, its voice, a `jobId`, an `assetId`,
  and `file: "sound/<role>.tts-<sha>.wav"`; `sound` carrying those three plus the two beds;
  `export.status: "completed"` with `droppedAudioLayers: []`.
- `config/film-harness/courier-workshop/sound/*.tts-*.wav` — three canonical PCM-16 WAVs of real
  Kokoro speech. Gitignored; leave them uncommitted.
- The exported MP4 at `export.renderPath` — 2 streams, container duration = timeline duration.
- `speech-probe.json` — one entry per placed dialogue slot with `present: true` and a `ratio` well
  above 1.5, plus a `synthesizedLines` provenance block. **The probe exits 0.**
- `api.log`, `worker.log` — three `audio_generate` claims and the Kokoro load in the worker log.

A tone probe cannot answer this one: a synthesized line has no known frequency. The probe compares
the decoded mix's RMS **inside** each placed dialogue window against a control window of the same
length beside it — the beds run continuously and are therefore in both numbers, and only the line is
in one. That is the exact shape of the phase-1 failure where the export came back with no lines.

The probe was smoked offline both ways before this PR went up, on a synthetic 10 s mix with a
continuous 100 Hz bed: with a line in the 6.3–7.8 s slot it reports `ratio 12.43, present: true` and
exits 0; with the same bed and nothing in the slot it reports `ratio 1.0, present: false` and exits
1. So a green probe on the real export is a measurement, not a tautology.

**Free sanity checks before spending GPU**: `scripts/film-harness-speech-probe.py --help` runs
offline, and `film-harness validate --plan config/film-harness/courier-workshop/plan.jsonc
--references config/film-harness/courier-workshop/references.jsonc --api $API_URL` accepts the
updated fixture against the live catalog.

### Tests

10 new in `apps/rust-api/src/tests/film_harness.rs` (8 integration over the real routes with the
fake worker extended to claim `audio_generate` and write a deterministic canonical PCM-16 WAV, 2
unit), 7 new unit tests in `crates/sceneworks-core/src/film_plan.rs`, plus the existing
fixture-byte and three-bus tests updated to the new fixture and given a guard that the shipped pack
and the tone generator stay in lockstep:

| test | what it pins |
| --- | --- |
| `a_dialogue_line_with_text_is_synthesized_placed_and_recorded` | AC1: one job through the real route, the dispatched body (prompt / model / voice / key / resolved `modelManifestEntry`), the WAV in the pack dir under the deterministic name, the imported asset's `synthesized: true`, the bus item at the shot's offset for the length actually spoken, and the record's provenance |
| `a_resume_adopts_a_spoken_line_instead_of_speaking_it_again` | AC1: a resume adopts; the API holds exactly one `audio_generate` job |
| `a_sound_entry_with_no_source_or_a_spoken_bed_is_refused_before_dispatch` | AC2: both refusals name the entry by role; no project, no job, `outcome: rejected` |
| `a_failed_synthesis_stops_the_run_resumably_with_the_rest_intact` | AC2: `dialogue_synthesis_failed`, resumable, zero renders, no export; the retry is attempt 2 under a new key and the film finishes |
| `a_synthesis_that_overruns_its_budget_stops_the_run_resumably` | AC2: the per-job limit, `timed_out`, nothing rendered |
| `speech_with_no_live_audio_worker_is_refused_before_the_job_exists` | the preflight: no job is created at all |
| `a_replacement_re_assembles_with_the_sound_the_run_already_has` | the replacement keeps all three tracks and speaks nothing again |
| `the_synthesis_key_separates_content_and_attempts_and_is_stable_otherwise` | the key's axes, including that trimming is idempotent |
| `a_synthesized_clips_name_is_deterministic_in_the_role_and_the_line` | the filename |
| `checked_in_fixture_sound_matches_the_generator_byte_for_byte` | extended: every shipped `dialogue` role is synthesized and every bed is a generator tone |

Everything runs on **every lane, ffmpeg or not**: a synthesized clip is a canonical PCM-16 WAV,
which the import route stores without a transcode (`media_convert::is_canonical_pcm16_wav`), exactly
as the fixture's beds are. The film-harness group was run with ffmpeg and again with
`PATH=/usr/bin:/bin:/usr/sbin:/sbin SCENEWORKS_FFMPEG=/nonexistent/ffmpeg`, which is the hosted
macOS lane's shape.

✅ **`npm run rust:check:neither` has now been run** (it was skipped on the first round for disk
reasons; the box had 166 GB free this time). Exit **0**.

No inference pin bump. No memory-matrix or calibration work.

---

## Review fixes (commit `5219811`)

Eight findings from the adversarial review, plus two scope gaps. All resolved.

### The synthesized clip now comes over the transport (major)

`synthesize_dialogue` fetched the worker's WAV with `std::fs::read(project_path.join(media_path))` —
straight out of the API host's project directory. Every other media hop the harness makes is HTTP,
and `--api` legitimately admits a private-network address, a `.local` name or a bare hostname whose
host *"may be a different machine"* (docs/film-harness.md:311, :826). On any such host **every
synthesis died with `HarnessError::Io`**.

It now goes through `GET /api/v1/projects/{id}/files/{mediaPath}` (the route at `lib.rs:1625`), with
the local read kept **only as a fast path** when `source.is_file()`. `ApiTransport::get_bytes` /
`BytesResponse` / `BytesTransportFuture` are added in the **exact shape sibling PR #2842 gives
them**, so the two merge trivially.

### `replace-take --export` no longer renders a stale timeline (major)

The re-assembly was guarded on `session.stop.is_none()`; the export was not. A replacement whose
sound could not be re-hydrated therefore rendered a fresh MP4 **from the timeline that still carried
the take the human had just replaced**, and `run_export` wrote `ExportRecord { stale: false }` over
the `stale: true` the replacement had set — recording old material as current. The export is now
gated the same way; `export.stale` stays `true` and the stop says why.

### The rest

| finding | resolution |
| --- | --- |
| `film_harness.rs:2239` + `docs:153` — "re-casting … is spoken again" unreachable and latently wrong | docs now state a re-cast is a **NEW run** (the pack-sha guard refuses it on resume); `recorded` keyed on `(role, file)`; a fresh synthesis attempt drops the stale append-only `record.sound` entry, which is the half the key cannot do when the pack **pins `file`** |
| `tests/film_harness.rs:6336` — `audio_job_count` counted the fake worker's claim log | reimplemented as `GET /api/v1/jobs?projectId=…&limit=100` filtered on `type == "audio_generate"` — it now also catches a job nobody claimed |
| `film_plan.rs:312` — `maxShotSeconds` doc | names dialogue synthesis and states no `maxSpeechSeconds` exists by design (mirrored in the docs) |
| `film-harness-smoke.sh:29` + `docs:987-995` — the GPU smoke wrote into the source tree and raced | the script copies the pack (document + `references/` + `sound/`) into `$SMOKE_DIR`, mirroring `Harness::fixture_pack`; an explicit `REFERENCES=` is left alone |
| `film_harness_fixes.rs:1491` — stale "(6.37s..8.37s)" | corrected to 6.37..7.87 (the line is 1.5 s) |

### Note on the two guards

The `(role, file)` key and the stale-clip drop are **guards, not live paths**: `open_session`'s
pack-sha check means a changed pack is a new run, so nothing today reaches `ensure_sound` with a
`record.sound` entry the entries disagree with. They are labelled as such in the code. Evidence they
do work: under the mutation that removes the pack-sha refusal, the resume re-speaks the line, drops
the stale entry and places the 1.4 s re-cast clip instead of the 1.2 s original.

### New tests

| test | what it pins |
| --- | --- |
| `a_synthesized_clip_is_fetched_over_the_file_route_when_the_project_dir_is_not_local` | the clip arrives over the file route when the project dir is unreadable here, byte-identical to what the worker wrote — **and** nothing is downloaded when the dir *is* local (the fast path) |
| `a_replacement_whose_sound_cannot_be_rehydrated_does_not_re_export` | no second export job, `export.stale` stays `true`, saved timeline untouched |
| `re_casting_a_line_is_refused_by_resume_and_spoken_by_a_fresh_run` | the pack-sha refusal, then a fresh run speaking the new line (item length = the **new** line's) |
| `a_dialogue_entry_with_both_text_and_file_synthesizes_into_the_pinned_path` | **scope gap**: synthesis writes into the pinned path (not the derived name), is imported and placed, and a resume over a *cleaned* pack adopts from `record.sound` rather than speaking again |

A new `CountingTransport` test double logs every `get_bytes` and can relocate the project documents'
`path` to an unreadable directory — which is what a remote API host looks like from the controller.

### Mutation evidence

Every new or changed assertion was proved RED by a targeted mutation, run one at a time with a
`touch` between runs and the source restored after each.

| --- | --- | --- |
| clip fetched over the file route | restore `std::fs::read` of the project dir | **RED** — `run completes: Io("No such file or directory")`, the exact failure the review predicted |
| local dir read directly (fast path) | `.filter(\|s\| s.is_file() && false)` | **RED** — "a local project directory is read directly: [/api/v1/projects/…/files/…wav]" |
| no second export job | revert the `&& session.stop.is_none()` guard | **RED** — export job count 2, expected 1 |
| `export.stale` stays true | same mutation, count assertion relaxed | **RED** — "the export must stay stale…" with `stale: false` |
| resume refuses a re-cast pack | pack-sha check narrowed to the plan only | **RED** — `unwrap_err()` on an `Ok` (the resume happily re-spoke the line) |
| synthesis writes into the pinned `file` | always use the derived name | **RED** — `sound/courier_line.tts-fdf0d8ab20f1.wav` vs `sound/courier_line.wav` |
| exactly one synthesis job (both resume tests, via the **reimplemented** `audio_job_count`) | disable the imported-clip adoption | **RED** — 2 jobs in the API's table, expected 1 |

### Gates (this round)

| gate | exit |
| --- | --- |
| `cargo fmt --all -- --check` | **0** |
| `npm run rust:check` (fmt + clippy `-D warnings` + `cargo test` + `cargo build`) | **0** |
| `npm run rust:check:neither` | **0** |
| film-harness group **with** ffmpeg | **0** — 139 passed, 0 failed (240 s) |
| film-harness group **without** ffmpeg (`PATH=/usr/bin:/bin:/usr/sbin:/sbin SCENEWORKS_FFMPEG=/nonexistent/ffmpeg`) | **0** — 139 passed, 0 failed (323 s) |

⚠️ **Pre-existing flake observed, not caused by this PR.** Two earlier `npm run rust:check` runs hit
`stop_signal_tests::a_signal_after_the_run_is_over_still_terminates`
(`apps/rust-api/src/bin/film-harness.rs:851`), a process-global signal race that only fails under
the workspace-wide `cargo test`'s load (it passes 3/3 in isolation). **It was reproduced on the
unmodified base commit `3b44c4df2`** — same test, same assertion — so it predates this branch. That
file is untouched by this diff. Flagging rather than fixing; say the word if you want it made
deterministic.


michaeltrefry:
## `dependency policy + advisories` is red for a reason that is not this PR

`RUSTSEC-2026-0285` — rustls accepts TLS 1.3 handshake messages across encryption-level boundaries.

```
rustls v0.23.40
├── hyper-rustls v0.27.9
│   └── reqwest v0.12.28 → (everything)
Solution: Upgrade to >=0.23.45 (`cargo update -p rustls`)
```

Not from this branch:

- This PR changes **no dependency files at all** — `git diff c4a8556cc..HEAD -- Cargo.toml Cargo.lock deny.toml advisory-ignores.toml '**/Cargo.toml'` is empty.
- `rustls` is a pre-existing transitive dependency of `reqwest`, which nothing here touches.
- #2839 and #2840 **passed** this same lane on run `34863726881`, about three hours before this run. The advisory was published in between.
- The advisory-policy step itself passed (`advisory policy: OK (0 governed ignores)`); it is `cargo deny`'s vulnerability check that fails.

So this will be red on **every open PR and both mains** until someone runs `cargo update -p rustls` and lands the `Cargo.lock` bump. That is a workspace-wide change with its own verification, not something to bury inside a film-harness story, so I am flagging it rather than doing it here — @michaeljon, say the word and it gets its own PR.

